PR #635, opened 2012-03-17
Save original error handlers and restore them afterwards

tomhughes:
Before installing a temporary error handler make sure any existing handler is saved and then restore it afterwards as other code may be using libxml and have installed error handlers.

In particular libxml-ruby installs a structured error handler which can then get wiped out by Nokogiri if you are trying to use both of them in the same process.


flavorjones:
I'm sorry we never got this to work. I just added some measure of libxml-ruby conflict coverage to the test suite, so we can revisit if you like.

PR #735, opened 2012-07-26
Handle the fact that in some cases, not entirely understood by me but reflected in the test case, the namespaces can come into XPathContext#register_namespaces with frozen UTF-16 keys

Empact:
I ran into this via the roo test suite. I suspect the fix could be more targeted, to avoid the case reaching this point in the code, but this is my personal best effort, and the fix works.

The XML was found in the Pfand_from_windows_phone.xlsx file in the roo test suite. https://github.com/Empact/roo/blob/master/test/files/Pfand_from_windows_phone.xlsx

The error without the fix is:
```
test_namespace_encoding(Nokogiri::HTML::TestDocumentEncoding):
Encoding::CompatibilityError: incompatible encoding regexp match (US-ASCII regexp with UTF-16 string)
    nokogiri/lib/nokogiri/xml/xpath_context.rb:11:in `gsub'
    nokogiri/lib/nokogiri/xml/xpath_context.rb:11:in `block in 
    nokogiri/lib/nokogiri/xml/xpath_context.rb:8:in `each'
    nokogiri/lib/nokogiri/xml/xpath_context.rb:8:in `register_namespaces'
    nokogiri/lib/nokogiri/xml/node.rb:152:in `block in xpath'
    nokogiri/lib/nokogiri/xml/node.rb:150:in `map'
    nokogiri/lib/nokogiri/xml/node.rb:150:in `xpath'
    nokogiri/test/html/test_document_encoding.rb:26:in `test_namespace_encoding'
```


flavorjones:
This seems to only be an issue with Ruby 1.9.3 -- I can't reproduce it on 2.0 or 2.1. Since 1.9.3 is about to be EOLed, and it looks like no further research went into this, I'm going to close.

If someone can reproduce this with Ruby 2+, please open a new issue with a test case? Thank you!

PR #757, opened 2012-09-01
parse_io should respect @encoding when no encoding is passed to it

soulcutter:
I assumed that #756 would not be controversial (which is why this pull request is based on that branch).

This change may be slightly more controversial since it changes existing functioning behavior from always defaulting to ASCII (and never using @encoding) to using @encoding by default (and @encoding defaults to UTF-8).

The existing behavior doesn't make sense to me and is not what I would expect, hence this separate pull request. One small issue remains here, existing behavior is that calling parse_io with an encoding has the side-effect of changing the encoding for all uses of the Parser instance, and I did not address that.


flavorjones:
As I mentioned in my comment from April 2013, the test in this PR passes on master right now (both MRI and JRuby), and so I don't think there's anything actionable here anymore. Closing.

Happy to reopen if I've misunderstood, just make a comment.

Issue #799, opened 2012-11-28
1.5.6 final release


benlangfeld:
master contains a great many bug fixes, especially with regards to JRuby compatibility, which are desperately needed. 1.5.5 was released more than 5 months ago. I think it's time for a bugfix release (1.5.6) and don't see any new bugs on master, or anything else that should present a problem.

Can we get a new release?

PR #813, opened 2012-12-19
Run valgrind on Travis CI

benlangfeld:
Relevant discussion in #799. Seems Travis is broken on my fork, so submitting this to trigger an official build.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,3 +25,9 @@ matrix:
       jdk: openjdk7
     - rvm: rbx-19mode
       jdk: openjdk7
+
+before_install:
+  - sudo apt-get update
+  - sudo apt-get install valgrind
+
+script: rake ci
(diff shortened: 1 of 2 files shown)


flavorjones:
@benlangfeld and @leejarvis -- I'd love to get valgrind running on Travis, except that the only assertion I'm currently making on the valgrind run is with my eyeballs.

So an interesting prerequisite is to wire up an assertion of some sort into https://github.com/jbarnette/hoe-debugging, so that the build fails if unexpected output is emitted. Difficulty level: medium.

Also, it would be great if we could with- and without-valgrind as separate builds. I know we can do this, because I'm looking at [jruby's travis config](https://github.com/jruby/jruby/blob/master/.travis.yml) and [jruby's build](https://travis-ci.org/jruby/jruby) and they're clearly doing something amazing with the `matrix` bit that I can't quite grok because I don't fully understand the [fully operational battle station](https://www.youtube.com/watch?v=v5lDKjA_7I0) that is TravisCI in 2015. Difficulty level: medium for nitwits like me, probably low for both of you.

flavorjones:
I still think this would be interesting to do. Happy to chat about my opinions if someone has time on their hands.

flavorjones:
Specifically, I think it would also be great to add the `test:libxml-ruby` and `test:valgrind:libxml-ruby` suites to Travis.

flavorjones:
Closing this, as I've got bigger ideas for CI at the moment.

PR #848, opened 2013-02-13
Make changelog consistent

gravis:
http://i.imgur.com/CTcqp.jpg


knu:
Oops, above was the wrong ref.

Issue #881, opened 2013-04-05
node_set.rb:24: [BUG] Segmentation fault


fenelon:
This mostly happens in concurrent environments. I'm using `nokogiri` with `sidekiq`, and `roadie` (which also uses nokogiri).
### nokogiri -v

```
# Nokogiri (1.5.6)
    ---
    warnings: []
    nokogiri: 1.5.6
    ruby:
      version: 1.9.3
      platform: x86_64-linux
      description: ruby 1.9.3p385 (2013-02-06) [x86_64-linux]
      engine: ruby
    libxml:
      binding: extension
      compiled: 2.9.0
      loaded: 2.9.0
```
### Backtrace:

```
/home/deployer/apps/leveltravel/shared/bundle/ruby/1.9.1/gems/nokogiri-1.5.9/lib/nokogiri/xml/node_set.rb:24: [BUG] Segmentation fault
ruby 1.9.3p385 (2013-02-06) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0055 p:---- s:0236 b:0236 l:000235 d:000235 CFUNC  :[]
c:0054 p:0022 s:0232 b:0232 l:000231 d:000231 METHOD /home/deployer/apps/leveltravel/shared/bundle/ruby/1.9.1/gems/nokogiri-1.5.9/lib/nokogiri/xml/node_set.rb:24
c:0053 p:0019 s:0227 b:0227 l:000226 d:000226 METHOD /home/deployer/apps/leveltravel/shared/bundle/ruby/1.9.1/gems/nokogiri-1.5.9/lib/nokogiri/xml/node.rb:248
c:0052 p:0160 s:0223 b:0223 l:000222 d:000222 METHOD /home/deployer/apps/leveltravel/current/app/operator_apis/modules/samo_operator.rb:193
c:0051 p:0654 s:0217 b:0217 l:000216 d:000216 METHOD /home/deployer/apps/leveltravel/current/app/operator_apis/modules/samo_operator.rb:161
c:0050 p:0012 s:0203 b:0203 l:002208 d:000202 BLOCK  /home/deployer/apps/leveltravel/current/app/operator_apis/modules/samo_operator.rb:93
c:0049 p:0015 s:0199 b:0199 l:000189 d:000198 BLOCK  /home/deployer/apps/leveltravel/shared/bundle/ruby/1.9.1/gems/nokogiri-1.5.9/lib/nokogiri/xml/node_set.rb:239
[...]

PR #895, opened 2013-04-25
Clean all occurrences of _private on GC when libxml-ruby is loaded.

ender672:
libxml-ruby uses the global libxml2 callback xmlDeregisterNodeDefault.

This callback looks in the _private field for every deleted libxml2
node. If the _private field is set, it is treated as a Ruby VALUE ptr.
The callback NULLs the Ruby object's data pointer and mark & free fn
pointers.

When Nokogiri wraps a libxml2 document it puts a nokogiriTuple ptr
in the _private field. We can't let the libxml-ruby callback be
invoked on a libxml2 document before NULLing the document _private
field.

When Nokogiri wraps other libxml2 nodes it puts VALUE ptrs in the
_private field. This makes the libxml-ruby callback generally safe for
these node types.

There is a caveat - it is unsafe to access VALUE ptrs during GC sweep.

This commit tests whether the xmlDeregisterNodeDefault callback is set
during document cleanup. If set, we traverse the document and clean up
all _private fields.

The previous solution was to unset xmlDeregisterNodeDefault callback.
However, this doesn't work in multithreaded environments

Fixes issues with loading Nokogiri + libxml-ruby in multithreaded environments.

See bug #881

### DIFF
--- a/ext/nokogiri/xml_document.c
+++ b/ext/nokogiri/xml_document.c
@@ -17,13 +17,29 @@ static int dealloc_node_i(xmlNodePtr key, xmlNodePtr node, xmlDocPtr doc)
   return ST_CONTINUE;
 }
 
+static void remove_private(xmlNodePtr node)
+{
+  xmlNodePtr child;
+
+  for (child = node->children; child; child = child->next)
+    remove_private(child);
+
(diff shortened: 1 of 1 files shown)


flavorjones:
see related #1426

Issue #885, opened 2013-04-09
collect_namespaces: Different result since 1.5.6


wolfgangw:
```Nokogiri::XML::Document#collect_namespaces``` will return different results with version 1.5.5 and versions since 1.5.6 respectively.

By the way: the sample below, while abstract and reduced, reflects real-world instance documents, not merely an academic exercise.

Consider this document with multiple default namespace scopes:

```xml
<?xml version="1.0"?>
<foo xmlns="foo_space">
  <bar>
    <barchild xmlns="bar_space">
      barchild says: I'm not always here.
      When I am I don't want to clutter
      root node's namespace definitions.
      Also I don't want to be cluttered
      myself -- no prefix for me please.
    </barchild>
  </bar>
</foo>
```

1.5.5:
```irb
irb> [Nokogiri::VERSION, Nokogiri::LIBXML_VERSION]
=> ["1.5.5", "2.7.8"]
irb> Nokogiri::XML(open 'multi-default-ns.xml').collect_namespaces
=> {"xmlns"=>"foo_space"}
```

1.5.6 and ever since:
```irb
irb> [Nokogiri::VERSION, Nokogiri::LIBXML_VERSION]
=> ["1.5.6", "2.7.8"]
irb> Nokogiri::XML(open 'multi-default-ns.xml').collect_namespaces
=> {"xmlns"=>"bar_space"}
```

This on ruby 2.0.0-p0. Reproducible with ruby 1.9.3-p194 / libxml 2.8.0.

```sh
$ diff nokogiri-1.5.5/lib/nokogiri/xml/document.rb nokogiri-1.5.6/lib/nokogiri/xml/document.rb
```

```diff
152,154c152,155
<       # Note this is a very expensive operation in current implementation, as it
<       # traverses the entire graph, and also has to bring each node across the
<       # libxml bridge into a ruby object.
---
>       # Note that this method does an xpath lookup for nodes with
>       # namespaces, and as a result the order may be dependent on the
>       # implementation of the underlying XML library.
>       #
156,158c157,160
<         ns = {}
<         traverse { |j| ns.merge!(j.namespaces) }
<         ns
---
>         xpath("//namespace::*").inject({}) do |hash, ns|
>           hash[["xmlns",ns.prefix].compact.join(":")] = ns.href if ns.prefix != "xml"
>           hash
>         end
```

[...]

PR #1001, opened 2013-11-13
Add Nokogiri::XML::Document#collect_namespaces_[href_keys|prefix_keys]

wolfgangw:
See https://github.com/sparklemotion/nokogiri/issues/885 for details

### DIFF
--- a/lib/nokogiri/xml/document.rb
+++ b/lib/nokogiri/xml/document.rb
@@ -165,43 +165,51 @@ module Nokogiri
       end
 
       ##
-      # In order to collect all namespaces in a document with multiple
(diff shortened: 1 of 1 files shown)


flavorjones:
Closing this. I'm sorry for this response, especially given the amount of time that has passed, but the use case of `collect_namespaces` is unclear to me, and I'd prefer not to increase the variations of methods unless and until I understand it.

Generally speaking, I'd love for a test case to illustrate how a method is going to be used when possible. I'm happy to entertain this PR again if someone can explain to me the use of this suite of methods around collecting namespaces.

PR #1017, opened 2013-12-10
Add `replace_entities` to the XML PushParser

### LINKED ISSUE #76: Problem with & character and SaxPushParser (Maybe SaxParser as well..)
Hey,

It seems that Nokogiri has problem with & inside attributes. if the & is not escaped, it raises an error, which seems logical :  
      Couldn't parse chunk (RuntimeError)

However, if the & is escaped in &amp;, then, the sax parser actually sees &#38; ... which is not what we want!

Here is a sample code :

```
 require "rubygems"
 require "nokogiri"

 class SAXHandler < Nokogiri::XML::SAX::Document
     def start_element(name, attrs = [])
          puts attrs.inspect
     end
 end

 xml = "<site url='http://sfbay.craigslist.org/search/roo/sfc/12?query=house&minAsk=min&maxAsk=max&neighborhood=12&format=rss'/>"

 sax_handler = SAXHandler.new()
 @parser = Nokogiri::XML::SAX::PushParser.new(sax_handler)
 @parser << xml
```

And here is the output :

```
 ["url", "http://sfbay.craigslist.org/search/roo/sfc/12?query=house&#38;minAsk=min&#38;maxAsk=max&#38;neighborhood=12&#38;format=rss"]
```

I think this is pretty critical, and it might as well show up with other characters (besides &).

Let me know if i can help!

spraints:
795cc5a resolved #76 for `Nokogiri::XML::SAX::Parser`. This branch adds the same option to `Nokogiri::XML::SAX::PushParser`.

For example:

``` ruby
require 'nokogiri'
include Nokogiri::XML

def main
  before
  after
end

def before
  puts '-'*40, 'before:'
  parser = SAX::PushParser.new(Dest.new)
  parser << doc
  parser.finish
end

def after
  puts '-'*40, 'after:'
  parser = SAX::PushParser.new(Dest.new)
  parser.replace_entities = true
  parser << doc
  parser.finish
end

def doc
  '<e a="x &amp; y">p &amp; q</e>'
end

class Dest < SAX::Document
  def start_element(name, attrs)
    p attrs
  end
  def characters(text)
    p text
  end
end

main
```

Output:

```
----------------------------------------
before:
[["a", "x &#38; y"]]
"p "
"&"
" q"
----------------------------------------
after:
[["a", "x & y"]]
"p "
"&"
" q"
```


flavorjones:
Tentatively targetting this change for 1.7.0.

rosenfeld:
I don't understand. With `replace_entities` set to `false`, which I believe is the default, no substitutions should be made, right? So, I'd expect `before` to print `&amp;` rather than `&#38;`. Isn't this a bug? I'm particularly interested on being able to get the raw value without any substitutions, which I think should be libxml2 default behavior, right?

spraints:
> With replace_entities set to false, which I believe is the default, no substitutions should be made, right? So, I'd expect before to print &amp; rather than &#38;. Isn't this a bug?

It might be, but that's not the goal of this PR. This branch lets me tell libxml2 to replace `&amp;` (or `&#38;`) with `&`.

flavorjones:
ZOMG, I've merged this in. Thank you for your patience, and apologies this took so long to get any attention.

spraints:
💖  thanks!

PR #1043, opened 2014-02-05
Use the :encoding arg on Document.parse even when contents are blank

batter:
Currently, `Nokogiri::XML::Document.parse` and `Nokogiri::HTML::Document.parse` discards the `encoding` argument if it receives `nil` or `blank` contents for the `string_or_io` argument.  This pull tweaks the code to set the `encoding` value (if it is passed in) even when the `string_or_io` argument is blank.

```ruby
# before this pull request
doc = Nokogiri.HTML(nil, nil, 'UTF-8')
doc.encoding # => nil
# after this pull request
doc = Nokogiri.HTML(nil, nil, 'UTF-8')
doc.encoding # => 'UTF-8'
```

This is really just a convenience thing, so that if you want to compose a blank document with an particular encoding, you can do something like this:

```ruby
doc = Nokogiri.HTML(nil, nil, 'UTF-8')
Nokogiri::HTML::DocumentFragment.new(doc, string)
```

instead of having to do something like this:

```ruby
doc = Nokogiri::HTML::Document.new
doc.encoding = 'UTF-8'
Nokogiri::HTML::DocumentFragment.new(doc, string)
```

This is particularly helpful for Ruby18 users where the `String` class does not have the concept of encoding on it's own.  True, you can technically condense the above example down into this:

```ruby
doc = Nokogiri::HTML::Document.new.tap { |d| d.encoding = 'UTF-8' }
Nokogiri::HTML::DocumentFragment.new(doc, string)
```

But it's cleaner to just have the `parse` method enforce the encoding regardless of contents it receives.

### DIFF
--- a/test/html/test_document_encoding.rb
+++ b/test/html/test_document_encoding.rb
@@ -3,8 +3,8 @@ require "helper"
 
 module Nokogiri
   module HTML
-    if RUBY_VERSION =~ /^1\.9/
-      class TestDocumentEncoding < Nokogiri::TestCase
+    class TestDocumentEncoding < Nokogiri::TestCase
+      if RUBY_VERSION =~ /^1\.9/  
         def test_encoding
           doc = Nokogiri::HTML File.open(SHIFT_JIS_HTML, 'rb')
 
@@ -81,6 +81,11 @@ module Nokogiri
           assert_equal ['http://tenderlovemaking.com/'],
             doc.css('a').map { |a| a['href'] }
         end
+      end
+
+      def test_empty_doc_encoding
+        encoding = 'US-ASCII'
+        assert_equal encoding, Nokogiri::HTML.parse(nil, nil, encoding).encoding
       end
     end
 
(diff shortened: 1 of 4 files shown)


flavorjones:
I'm very sorry to not have responded to this long ago.

This seems like a reasonable change. Will make one comment inline.

test/html/test_document_encoding.rb
```diff
@@ -3,8 +3,8 @@
 
 module Nokogiri
   module HTML
-    if RUBY_VERSION =~ /^1\.9/
-      class TestDocumentEncoding < Nokogiri::TestCase
+    class TestDocumentEncoding < Nokogiri::TestCase
+      if RUBY_VERSION =~ /^1\.9/
```
flavorjones: I imagine, given the time frame of the submission, that this was intended to exclude 1.8 and not to exclude 2.0+. Correct?

flavorjones:
Merged manually. Thank you for this contribution!

Issue #1045, opened 2014-02-07
Add bang versions of finders that raise when nothing found?


nathanl:
If a selector doesn't match anything, it can lead to unhelpful errors.

``` ruby
page.at_css('.foo .bar').text # => NoMethodError: undefined method `text' for nil:NilClass
```

What would you think of bang methods that raise when the selector is not found? Something like:

``` ruby
module Nokogiri
  module XML
    def at_css! *rules
      at_css(*rules).tap { |result| 
        raise Nokogiri::SelectorNotFound if result.nil? 
      }
    end
  end
end
```

flavorjones:
Closing, see the conversation at #1050 for the history of this feature suggestion.

PR #1050, opened 2014-02-13
Bang finder methods - raise if element is nil.

nathanl:
Add 'at!', 'at_css!', and 'at_xpath!' to Node and NodeSet.
Each bang method simply calls its non-bang counterpart and
either returns its value or, if that value is nil, raises
Nokogiri::XML::NotFound.

Fixes https://github.com/sparklemotion/nokogiri/issues/1045

### DIFF
--- a/lib/nokogiri/xml/bang_finders.rb
+++ b/lib/nokogiri/xml/bang_finders.rb
@@ -1,20 +1,29 @@
 module Nokogiri
   module XML
     module BangFinders
-
       def at!(*args)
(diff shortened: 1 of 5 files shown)


leejarvis:
@nathanl I'm gonna close this as I don't think there's enough interest to bring it into Nokogiri. Thanks for the suggestion though and great work on nokogiri_bang_finders!

nathanl:
Sounds good. Thanks for considering. :+1: :smile:

Issue #843, opened 2013-01-31
Assertion failed: (doc->_private), function Nokogiri_wrap_xml_namespace, file xml_namespace.c, line 41. 


juskoljo:
Hi,

I'm having issues with reader.attribute_nodes and reader.attributes methods.

a) reader.attribute_nodes doesn't seem to work as expected? If I remove xsi:schemaLocation from the XML it works ok. Assertion failed: (doc->_private), function Nokogiri_wrap_xml_namespace, file xml_namespace.c, line 41. Abort trap: 6

b) reader.attributes returns {"xmlns:xsi"=>"http://www.w3.org/2001/XMLSchema-instance", "schemaLocation"=>"url filename.xsd"}. How can I get ns/prefix for schemaLocation?

With best regards,
Jussi

```ruby
require 'nokogiri'
xml = '<xml xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="url filename.xsd"></xml>'
reader = Nokogiri::XML::Reader(xml)
while reader.read
  p reader.attribute_nodes
end

1.8.7 :001 > Nokogiri::VERSION_INFO
 => {"ruby"=>{"platform"=>"i686-darwin11.4.2", "engine"=>"mri", "description"=>"ruby 1.8.7 (2012-10-12 patchlevel 371) [i686-darwin11.4.2]", "version"=>"1.8.7"}, "nokogiri"=>"1.5.6", "libxml"=>{"loaded"=>"2.8.0", "binding"=>"extension", "compiled"=>"2.8.0"}, "warnings"=>[]} 
```

PR #1058, opened 2014-02-27
Fix for #843 - the xml reader bombs out when trying access attribute namespaces

2potatocakes:
See #843 for an example of the issue.

Was hoping for some help with this if anyone can lend a hand. Have added a test case which is now passing under C ruby but failing under jruby.

These commits fix an issue in the nokogiri reader when trying to access namespace properties on attribute nodes. Currently when trying to access them via the attribute_nodes method it halts the ruby interpreter :(

If you copy and paste the test case into master, without applying the patch you should be able to see the test suite crash mid way through.

### DIFF
--- a/ext/nokogiri/xml_namespace.c
+++ b/ext/nokogiri/xml_namespace.c
@@ -38,19 +38,19 @@ VALUE Nokogiri_wrap_xml_namespace(xmlDocPtr doc, xmlNsPtr node)
 {
   VALUE ns, document, node_cache;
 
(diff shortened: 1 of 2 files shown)


codekitchen:
We really need this fix, so I tracked down the issue with the jruby implementation and added a fix. New PR is https://github.com/sparklemotion/nokogiri/pull/1327 , since I don't think I can update this one.

flavorjones:
Closing this PR, superseded by #1327.

PR #1120, opened 2014-06-26
Update README examples with to match new Google HTML.

danfinnie:
The perils of HTML scraping...

I wasn't sure what the combined example was supposed to be because the CSS selector and the XPATH selectors are equivalent.  Is that how it's supposed to be?

@flavorjones


leejarvis:
I think they're supposed to be the same. I think the markup for Google changes too often for this to be useful. Maybe we should change the example to provide something more concrete

flavorjones:
Closing, see commit above.

Issue #1119, opened 2014-06-25
Nokogiri install fails because of missing lzma symbols in self built libxml2


dwt:
Here's the errors I'm getting: 

```
Undefined symbols for architecture x86_64:
  "_lzma_auto_decoder", referenced from:
      _xz_head in libxml2.a(xzlib.o)
  "_lzma_code", referenced from:
      _xz_decomp in libxml2.a(xzlib.o)
  "_lzma_end", referenced from:
      ___libxml2_xzclose in libxml2.a(xzlib.o)
  "_lzma_properties_decode", referenced from:
      _is_format_lzma in libxml2.a(xzlib.o)
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [xmllint] Error 1
make[1]: *** [all-recursive] Error 1
make: *** [all] Error 2
````

My c is a bit rusty, but I think this means that when nokogiri builds it's own libxml2 it is missing a dependency.

So what to do? If that is the case, another release is probably necessary.

**Comment by flavorjones** [comment]
I believe #1218 getting merged into master addresses this. Will be in 1.6.7 when it's cut.

If you still have this problem on master, please comment and I'll re-open.

**Comment by jorgeboucas** [comment]
this worked for me: 

```
sudo gem install nokogiri -- --use-system-libraries --with-xslt-dir=/usr/local/opt/libxslt --with-iconv-dir=/usr/local/opt/libiconv --with-xml2-dir=/usr/local/opt/libxml2 --with-xml2-config=/usr/local/opt/libxml2/bin/xml2-config --with-xml2-include=/usr/local/opt/libxml2/include/libxml2 --with-xslt-config=/usr/local/opt/libxslt/bin/xslt-config
```

PR #1125, opened 2014-07-08
link libxml2 with liblzma if it is found

nurse:
If the system have liblzma (and lzma.h), libxml2 enables xz/lzma helper.
It compiles xzlib.c which depends liblzma.
Therefore extconf.rb must run find_library("lzma") and link libxml2
with liblzma.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -505,6 +505,7 @@ else
   end
 end
 
+find_library("lzma", 'lzma_end') # for systems with liblzma
 {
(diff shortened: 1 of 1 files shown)


flavorjones:
@nurse I think you submitted this to address issues similar to #1119. If that is true, I think #1218 addresses the root cause in a different way, and that's in 1.6.7. What do you think?

Apologies for not responding on this PR sooner.

nurse:
@flavorjones Hmm, I don't remember the detail of this, but it's sounds reasonable. thanks!

PR #1134, opened 2014-07-22
Fix #1109. Clear the XPath cache when an attribute is removed

### LINKED ISSUE #1109: removing attribute doesn't affect results of xpath query when using JRuby
Removing an attribute from an element appears to not clear a cache somewhere when using JRuby -  behaves correctly when using MRI.  Currently causing capybara to fail when used with JRuby and nokogiri 1.6.2.1

    doc = Nokogiri::HTML("<html><div selected></div></html>")
    elements = doc.xpath('//div[@selected]') 
    puts elements.size()   # 1
    elements[0].remove_attribute('selected')
    puts doc.xpath('//div[@selected]').size()  # 0 with MRI - 1 with JRuby

jvshahid:
This is a fix for #1109. I'd like someone to eyeball since it introduces some changes to the ruby side of nokogiri to conditionally define `remove_attribute` in MRI only. Under JRuby `remove_attribute` is defined in the Java code and the code was being replaced with the ruby version.

### DIFF
--- a/ext/java/nokogiri/XmlNode.java
+++ b/ext/java/nokogiri/XmlNode.java
@@ -1315,6 +1315,7 @@ public class XmlNode extends RubyObject {
             String key = name.convertToString().asJavaString();
             Element element = (Element) node;
             element.removeAttribute(key);
+            clearXpathContext(getNode());
         }
         return this;
     }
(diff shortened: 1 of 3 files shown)


flavorjones:
I took a slightly different approach to clearing the cache in a8b55d03763e12e67ec89366b202694249686779 which I just put on master.

PR #1149, opened 2014-08-27
failing test case for html without <head> on jruby

esse:
When parsing html without <head> tag on jruby, <head> tag is created and present in returned html.

### DIFF
--- a/test/html/test_document.rb
+++ b/test/html/test_document.rb
@@ -7,6 +7,19 @@ module Nokogiri
         super
         @html = Nokogiri::HTML.parse(File.read(HTML_FILE))
       end
+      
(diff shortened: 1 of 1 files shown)


flavorjones:
Hmm. For me, this particular wart falls under the category of "underlying XML/HTML parsing engine deciding to correct markup differently than you might expect."

For context, there are two different parsing engines that Nokogiri uses: for MRI, we use libxml2; for JRuby we use xerces/nekohtml. Each has its own implementation of markup correction, which often differs.

In this particular case, they generate different, but equally valid, HTML.

Unfortunately, there's no obvious or easy way for Nokogiri to modify the behavior of the underlying engines' markup-correction behavior.

Issue #1112, opened 2014-05-27
Error unlinking empty document in jruby


wobh:
My coworker and I ran into this funny case where the jruby version of Nokogiri kicks up an error that MRI does not:

```
~% jruby -e "require 'nokogiri'; Nokogiri::HTML::DocumentFragment.parse('').unlink"
RuntimeError: TYPE: 11 PARENT NULL
  unlink at nokogiri/XmlNode.java:1313
  (root) at -e:1
~% ruby -e "require 'nokogiri'; Nokogiri::HTML::DocumentFragment.parse('').unlink"
~%
```

The case for this is easily avoided with blank checking, but we were wondering if there was any explanation for it. Thanks!

PR #1152, opened 2014-08-30
if node doesnt have parent unlink in jruby should return self.

esse:
If node doesn't have parent node - simply return self, as would Nokogiri for C would do, because that means that node is already detached (test case included).

This commit fixes issue #1112.


flavorjones:
Thanks for submitting this! I've cleaned up the implementation with a subsequent commit, and added more explicit test coverage in `test/xml/test_unparented_node.rb`.

PR #1178, opened 2014-10-20
Implicit block call instead explicit block call. Faster code.

glaucocustodio:
More info at https://github.com/JuanitoFatas/fast-ruby#proc--block


larskanis:
Thanks, merged!

PR #1193, opened 2014-11-17
Escape full path when patching

danielfone:
This ensures that patching is successful even if the ROOT path contains spaces.
This is the same approach that mini_portile takes. https://github.com/luislavena/mini_portile/blob/v0.6.1/lib/mini_portile.rb#L256


leejarvis:
LGTM. Build failed with unrelated errors. Merging. Thanks @danielfone

PR #1198, opened 2014-11-24
CLI mode: use Pry rather than IRB if it's available

robmiller:
Pry offers a couple of features, namely the pretty-printing of collections and syntax highlighting of returned values, that make working with Nokogiri documents in the REPL more user-friendly.

This pull request doesn't add Pry as a dependency of Nokogiri or change the default behaviour (I figured that was too far), but instead simply uses Pry over IRB if Pry is available and otherwise falls back to the previous IRB behaviour.

I use the CLI interface to Nokogiri quite a lot, so this improves things for me considerably; I assume there are others in the same boat.

### DIFF
--- a/bin/nokogiri
+++ b/bin/nokogiri
@@ -2,6 +2,7 @@
 require 'optparse'
 require 'open-uri'
 require 'irb'
+require 'pry' rescue LoadError
 require 'uri'
 require 'rubygems'
 require 'nokogiri'
@@ -72,7 +73,11 @@ else
     eval @script, binding, '<main>'
   else
     puts "Your document is stored in @doc..."
-    IRB.start
+    if defined? Pry
+      Pry.start
+    else
+      IRB.start
+    end
   end
 end
 


leejarvis:
Not sure about forcing the use of pry just because it's installed. I could install something that depends on Pry but prefer IRB. This means I'm forced to use Pry. I think making it pluggable would be nice, though. Any suggestions for that?

robmiller:
Thanks for the response!

Is there a compelling reason why one might want to use IRB rather than Pry? The latter's features are a superset of the former's, rather than it being a preference or something.

If you feel there is then I guess an option might be to switch based on a command-line option, and then users that want this behaviour can alias it in their shells. (In this case I think it would be good to have the "if Pry is installed then Pry; else IRB" logic be the default and have an option to force IRB, personally, but I'm not sure if you feel that would be disorientating or something.)

knu:
As far as I know, Rails uses Pry only if "pry-rails" is also listed in the Gemfile, and better_errors uses Pry only if `BetterErrors.use_pry! ` is called in one of the initializers.  So, I see it as a preference.  Users (especially existing users of the nokogiri command) that have a customized .irbrc and a history file full of useful code snippets would not want nokogiri to use Pry just because the pry gem was installed perhaps as a development dependency pulled in by some unknown gem.

I'm not exactly sure where and how it should be configured, but I guess a command line option could be added to load some initialization code from a file which might be defaulted to "~/.nokogirirc".

knu:
How does the above PR look like?

leejarvis:
Looks good to me

leejarvis:
I'm closing this as it seems covered by the commit from knu. Thanks for the PR Rob!

Issue #1114, opened 2014-06-02
LoadError when deploying to JRuby/Torquebox on Windows


fpauser:
Ruby: JRuby 1.7.11
Rails: 4.0.5
Nokogiri: 1.6.2.1
OS: Windows 2008 Server (64bit)
Torquebox: 1.3.1
Java: 1.7.0_40-b43

Starting the rails-app gives me:

    (LoadError) load error: nokogiri/nokogiri -- java.lang.NoClassDefFoundError: com/sun/org/apache/xpath/internal/VariableStack

Older versions of Nokogiri (tried it with 1.5.11) work.

fpauser:
Any news on that?

flavorjones:
@fpauser Unfortunately, to my knowledge, nobody on nokogiri-core has reproduced this issue. Though we don't need to reproduce it to follow @bbrowning's advice to remove the dependency on the internal classes, it's hard to know we're doing the right thing if we can't reproduce it.

Can anyone provide instructions on how to build an environment on which I can reproduce? Or, for  extra hugs, provide a docker image? If we can get unblocked on repro we'll fix it.

flavorjones:
Should I be able to reproduce this if I start from https://hub.docker.com/_/websphere-liberty/ ?

atambo:
I think you can reproduce it by using the IBM JDK: http://www.ibm.com/developerworks/java/jdk/ instead of the Oracle JDK but you'd have to be using linux.

karlhe:
Or Windows. IBM JDK doesn't support mac environment though.

altmind:
Hit the same issue.

Fresh server, no leftover rubies, gems of custom java libs.
Debian 6.0(oldstable)
Java(TM) SE Runtime Environment (build 1.7.0_80-b15)
Java HotSpot(TM) 64-Bit Server VM (build 24.80-b11, mixed mode) - official package from http://ppa.launchpad.net/webupd8team/java/ubuntu
torquebox-dist-3.1.2
Fresh rails app 4.2.4(basically only 1 scaffolded model)

Running torquebox with app deployed gives loooong stacktrace with "org.jruby.exceptions.RaiseException: (LoadError) load error: nokogiri/nokogiri -- java.lang.NoClassDefFoundError: com/sun/org/apache/xpath/internal/jaxp/JAXPExtensionsProvider"

Full torquebox error: https://gist.github.com/altmind/e4452843a16c58ad6d06

[...]

PR #1212, opened 2014-12-22
Add xalan jars to fix issues with #1114

atambo:
Jars downloaded from http://xalan.apache.org/xalan-j/downloads.html

@yokolet, @jvshahid, @flavorjones, @bbrowning

Does this look reasonable? I am one of the poor people that need to use the IBM JDK so I think this is the only solution. Running the tests locally gives me 3 failures. It seems like at least 1 of the failures is just a change in an exception message. Should I expect all of the tests to pass on jruby?


mkristian:
+1 
yes, very much. this also can cause funny and hard to debug classloader issues when you turn of classloader.delegate to the parent classloader which you can to on servlet containers and also with jruby-9k (actually jruby-9.0.0.0.pre1 has this a default without a possibility to turn it off,).

since you guys to embed the jars inside the gem any ways it is always possible to bundle a jar with only the few classes from xalan you really use for the XPath stuff.

but we also should make sure xerces and xalan and xml-apis are fitting together. see also my xerces update on: https://github.com/sparklemotion/nokogiri/pull/1251

atambo:
@mkristian, so your commits in update #1251 will handle updating xerces. Do you have any ideas why  embedding xalan causes these test failures?:

https://travis-ci.org/sparklemotion/nokogiri/jobs/44859569

mkristian:
@atambo my comment above with the classloader referred to https://github.com/sparklemotion/nokogiri/blob/master/ext/java/nokogiri/XmlXpathContext.java#L61
and I thought that are the xalan classes nokogiri uses. now I realised that this PR does not address those com.sun.org.apache.* classes.

I would assume it is due to xml-api that something goes wrong. on the other hand updating nekohtml did fix an encoding bugs and produced 19 failures in the test-suite. so I am not surprised that adding xalan.jar has some regressions.

having said this - the same regression will take place whenenver a project whats to use xalan and just picks the latest xalan and requires it.

I might find some time to look into this xalan update.

flavorjones:
Hi all,

I'm not a Java guy, so before I merge anything I need to fully understand what's going on here.

Can someone explain clearly, as if I was a Java newbie:

* the underlying issue this is addressing
* what the pros and cons are
* any alternative solutions that were considered

Thanks for helping me understand and thereby better support Nokogiri and your use cases.

atambo:
The original issue is explained in this comment: https://github.com/sparklemotion/nokogiri/issues/1114#issuecomment-67849315

This commit adds the jars that contain these internal classes to the gem so that JDK's that don't bundle xalan will still be able to use nokogiri. I don't know of any other solution to this issue.

mkristian:
but adding the xalan jar is only the first step, as mentioned here: https://github.com/sparklemotion/nokogiri/issues/1114#issuecomment-67849315

you also need to use the classes from the xalan jar instead of the com.sun.* ones !!!!

alternatively:
instead of the xalan jar which is about 3MB big you can just pick those handful of classes in pack them into some xalan-nokogiri.jar

but in anycase it is worth to see if all those xerces, xalan and xml-apis fit together.

atambo:
So I switched all the internal packages over to the ones included in the xalan jars.

jvshahid:
I was able to repro the failures locally. I'll try to get them fixed so we can merge this pr. Thanks for being patient.

jvshahid:
I started looking at the failure in `test_transform_with_output_style` and it looks like the result of `    assert_no_match(/<td>/, xslt.apply_to(@doc, ['title', 'foo']))` was trivially true because the return value of `apply_to` was the empty string. tl;dr; the current behavior as well as the previous (i.e. before merging the pr) behavior are incorrect.

jvshahid:
Fixed the XSLT failure on a branch, will look at the other failures as soon as I can.

PR #1249, opened 2015-02-20
Do not use strdup() in Node::get() anymore.

maltoe:
Hi,

this replaces `strdup()` in `xml_node.c::get()` as it was causing a segmentation fault for certain input for us on a rather ancient CentOS 6 system.

Unfortunately can't provide a test case for it, as it does not crash with a more recent GCC/glibc. However, MWE can be found [here](https://github.com/FlavourSys/nokogiri-strdup-segfault-mwe).

Malte

### DIFF
--- a/ext/nokogiri/xml_node.c
+++ b/ext/nokogiri/xml_node.c
@@ -826,7 +826,11 @@ static VALUE get(VALUE self, VALUE rattribute)
   if (NIL_P(rattribute)) return Qnil;
 
   Data_Get_Struct(self, xmlNode, node);
-  attribute = strdup(StringValuePtr(rattribute));
+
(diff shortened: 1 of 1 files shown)


larskanis:
Possibly this is a result of using StringValuePtr() instead of StringValueCStr(). This is addressed in https://github.com/sparklemotion/nokogiri/pull/1381 .

maltoe:
@larskanis We weren't using Ruby 2.2 at that time. We have also since reinstalled that CentOS 6 system and the error is gone. I realized we had installed a really recent kernel on that machine (but old userland), maybe that was causing the issue. Not sure though. I think it's fine to just close this.

larskanis:
@maltoe OK, thanks!

PR #1254, opened 2015-02-25
fix XML::Reader working with non-existent attributes

ccutrer:
nokogiri now bundles a newer libxml, so doesn't have to
worry about old bad behavior, and the workaround was broken

without this fix, the test case would return 
`reader.attribute('other') # => bob`
apparently picking up the default namespace for any missing attributes


codekitchen:
:+1: this fixes the issue i'm having trying to use `reader.attribute(name)` to get the values of attributes that may or may not be present.

flavorjones:
Thanks for this PR, tentatively targetting 1.6.8 for it.

flavorjones:
Merged manually. Thank you again for this PR and for your amazing patience!

This will be in 1.6.8 final.

PR #1268, opened 2015-03-23
Download libxml2 and libxslt via HTTP

eagletmt:


flavorjones:
Thank you for this contribution!

I would like to understand why you're suggesting this change. Can you explain why you think this is necessary or better?

eagletmt:
There's an environment where the network is somewhat restricted because of nat, proxy, etc.
Since FTP is less comon than HTTP nowadays, only HTTP(S) works correctly in such an environment.
It would be better to me if I can download the tarball via HTTP.

flavorjones:
Thanks again!

eagletmt:
Thank you!

eagletmt:
Are there any plan to release a new version?
I'm suffering from this issue.

PR #1275, opened 2015-04-14
Change to make sure that Nokogiri works with gem-compiler to generate binary gem

jk-shah:
This change might require more review..

Basically we need to create binary gems. We use gem-compiler for that. In order to make gem-compiler work with nokogiri,  we need to make sure all original files exists out there so that it can be repackaged without any errors.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -60,12 +60,6 @@ def do_clean
       FileUtils.rmdir(dir.parent, parents: true, verbose: true)
     }
 
-    if enable_config('static')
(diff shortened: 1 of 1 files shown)


flavorjones:
Sorry, I cannot accept this patch as-is. You're proposing to delete behavior without striving to understand why we're cleaning.

I will, however, work on making sure Nokogiri doesn't invalidate its gemspec after running a "clean". The gem-compiler use case was not considered when we were jumping through these hoops.

flavorjones:
See https://github.com/sparklemotion/nokogiri/commit/20d7591983637365d469cd20a7eea663ad78f7ae

PR #1291, opened 2015-05-22
Add iframe attritube 'autoplay'

joelparkerhenderson:
Add iframe attritube 'autoplay'. (Also reindent the successive lines).

The goal is to have Nokogiri HTML output for an iframe use a binary attribute for autoplay, such as:

    <iframe src="..." autoplay>

### DIFF
--- a/lib/nokogiri/html/element_description_defaults.rb
+++ b/lib/nokogiri/html/element_description_defaults.rb
@@ -216,9 +216,9 @@ module Nokogiri
       HR_DEPR = ['align', 'noshade', 'size', 'width']
       VERSION_ATTR = ['version']
       HTML_CONTENT = ['head', 'body', 'frameset']
-      IFRAME_ATTRS = [COREATTRS, 'longdesc', 'name', 'src', 'frameborder',
-                      'marginwidth', 'marginheight', 'scrolling', 'align',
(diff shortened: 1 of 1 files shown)


flavorjones:
Hi @joelparkerhenderson,

Thanks for sending this. I'm not sure I understand what the problem is that's being solved; and no tests have been provided to help me understand.

If this is somehow related to HTML5, please understand that libxml2 doesn't support HTML5 at this time.

I'm going to close this; but if you'd like to continue the conversation, comment and I'd be happy to reopen.

joelparkerhenderson:
Thanks. The problem (as of May 21) was that Nokogiri was interfering with libxml2. 

The current Nokogiri (as of Nov 22) doesn't have the problem, so I presume someone else fixed it elsewhere. Closing this issue is AOK.

> please understand that libxml2 doesn't support HTML5 at this time.

That's an interesting topic, if you're curious. What I see is that libxml2 can parse HTML5, though with a HTML4 markup declarations, and will generate elements and attributes in the tree for syntax that it doesn't know. Daniel Veillard has written that too. (What I saw back in May was libxml2 parsing the autoplay attribute, but Nokogiri not disallowing it it because the attribute wasn't in the list of allowable attributes.)

PR #1296, opened 2015-06-05
Using symbol to proc

bruno-:
This pull request contains simple mechanic changes from using blocks to symbol
to proc.

### DIFF
--- a/lib/nokogiri/xml/node.rb
+++ b/lib/nokogiri/xml/node.rb
@@ -331,13 +331,13 @@ module Nokogiri
       ###
       # Get the attribute values for this Node.
       def values
-        attribute_nodes.map { |node| node.value }
+        attribute_nodes.map(&:value)
       end
 
       ###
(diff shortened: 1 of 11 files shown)


larskanis:
I don't have a strong opinion to this, but it's surely a bit more readable.
@sparklemotion/nokogiri-core Any objections to merge this?

knu:
Actually, since a proc made from a symbol does not (need to) capture the environment, the symbol to proc style `map(&:foo)` will run faster than the closure style.

knu:
By the way, do we have coverage numbers to make sure the lines touched are all covered by the test suite?

larskanis:
> By the way, do we have coverage numbers to make sure the lines touched are all covered by the test suite?

Good point - will check this.

larskanis:
@knu  We have a very good coverage. In fact all the changes of this PR are covered.

larskanis:
@bruno- Merged, thank you!

PR #1297, opened 2015-06-05
XML::Node#ancestors optimization

bruno-:
While reading the source I stumbled upon this and figured it can't hurt to
optimize a bit.


larskanis:
Merged, thank you!

PR #1298, opened 2015-06-05
Fix cli tool example links

bruno-:
Running an example command `nokogiri http://www.ruby-lang.org/` results with the following error

    'open_loop': redirection forbidden: http://www.ruby-lang.org/ -> https://www.ruby-lang.org/ (RuntimeError)

As for the other example `curl -s http://nokogiri.org` returns

    <html>
    <head><title>301 Moved Permanently</title></head>
    <body bgcolor="white">
    <center><h1>301 Moved Permanently</h1></center>
    <hr><center>nginx</center>
    </body>
    </html>

which (I think) has `h1` tag only by accident.


larskanis:
Thanks, merged!

PR #1303, opened 2015-06-17
unescape css identifiers to allow use in css queries

twalpole:
Allow for querying css that uses "special" characters in id and class names such as
<div id="a.b"></div>
querying with 
document.css('#a\\.b')
will work correctly with this PR - failed previously

### DIFF
--- a/lib/nokogiri/css/parser.rb
+++ b/lib/nokogiri/css/parser.rb
@@ -8,9 +8,15 @@ require 'racc/parser.rb'
 
 
 require 'nokogiri/css/parser_extras'
+
 module Nokogiri
   module CSS
     class Parser < Racc::Parser
+
+
+def unescape_css_identifier(identifier)
+  identifier.gsub(/\\(?:([^0-9a-fA-F])|([0-9a-fA-F]{1,6})\s?)/){ |m| $1 || [$2.hex].pack('U') }
+end
 ##### State transition tables begin ###
 
 racc_action_table = [
@@ -406,7 +412,7 @@ end
 # reduce 17 omitted
 
 def _reduce_18(val, _values, result)
- result = Node.new(:CLASS_CONDITION, [val[1]]) 
+ result = Node.new(:CLASS_CONDITION, [unescape_css_identifier(val[1])]) 
     result
 end
 
@@ -650,7 +656,7 @@ end
 # reduce 57 omitted
 
 def _reduce_58(val, _values, result)
- result = Node.new(:ID, val) 
+ result = Node.new(:ID, [unescape_css_identifier(val.first)]) 
     result
 end
 
--- a/lib/nokogiri/css/parser.y
+++ b/lib/nokogiri/css/parser.y
@@ -55,7 +55,7 @@ rule
     | simple_selector
     ;
   class
-    : '.' IDENT { result = Node.new(:CLASS_CONDITION, [val[1]]) }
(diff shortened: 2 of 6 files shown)


twalpole:
@flavorjones Any feedback on this?   Had this issue reported to Capybara when a user was trying to select elements with '.' in their ids using a css id selector.

lib/nokogiri/css/parser.rb
```diff
@@ -8,9 +8,15 @@
...
 
 require 'nokogiri/css/parser_extras'
+
 module Nokogiri
   module CSS
     class Parser < Racc::Parser
+
+
+def unescape_css_identifier(identifier)
+  identifier.gsub(/\\(?:([^0-9a-fA-F])|([0-9a-fA-F]{1,6})\s?)/){ |m| $1 || [$2.hex].pack('U') }
```
flavorjones: Can you provide some context around this regex? Did you pull it from an RFC?

twalpole: it matches a  \ followed by any character other than hex characters, so  \: , etc  - or any 1-6 digit hex number potentially followed by whitespace - these are the two methods of escaping characters in css ids, classes, etc

twalpole: and from http://www.w3.org/TR/css3-selectors/#selector-syntax 
"Characters in Selectors can be escaped with a backslash according to the same escaping rules as CSS. [CSS21]"

flavorjones:
@twalpole - Ping. Saw your email on nokogiri-talk. How can I help?

twalpole:
@flavorjones What changes are needed for this PR to be accepted?   - the current test failures are only The ones caused by the latest commit on master

flavorjones:
@twalpole It sounds like you're asking why this wasn't nominated for inclusion in 1.6.7. A few reasons:

* 1.6.7 is an "installation improvement" release, and Windows devs have been waiting quite a while to get this (it's a blocker for Rails on Windows). To the extent that I can omit things that aren't about improving installation, I reduce risk around this release.
* the change you've proposed is a bit invasive, in that it touches every CSS query, and unfortunately I'm not all that confident in our tests around CSS parsing (they're largely ad-hoc and added after-the-fact).

These things combined led me to move it slightly lower on the priority list.

That said, I'd like to do a fast-follow of 1.6.8 to get a bunch of bugfixes in, so I'm marking this for 1.6.8 and will make a point of getting it out as quickly as I reasonably can.

Thanks for your patience.

twalpole:
Thanks for the explanation/status update - I really just wanted to know whether it was acceptable , needed changes, or I needed to consider trying to somehow patch it from capybara.  I'll wait for 1.6.8 to start firming up

flavorjones:
@twalpole -Thanks for following up.

I would like one more test added, to more accurately reflect the underlying use case at https://github.com/jnicklas/capybara/issues/1541, which is the ability to query for an encoded css attribute. Would you mind adding such a test?

The unit tests written so far simply assert on the xpath generated from an escaped query, and not that a properly-unescaped query will actually return the expected nodeset.

twalpole:
@flavorjones  test added, would you like me to squash them down to one commit?

flavorjones:
@twalpole Thanks so much for doing that! LGTM, will be in 1.6.8 when it drops.

flavorjones:
Merged onto master! Thanks again, will be in 1.6.8.

PR #1304, opened 2015-06-18
Add line column to message

spk:
Useful for `map(&:to_s)` and simple output


flavorjones:
Oooh, I really like this feature. I may clean it up so that it reads better, but thanks for putting the time and though into this!

spk:
Hi @flavorjones, thanks, don't hesitate to comment I will take the time to fix the issues you will report.

flavorjones:
Looks like this works on rbx-3 without needing to make the change in cbdd787.

flavorjones:
Merged the initial commit of this PR manually.

Thank you for the PR!

PR #1305, opened 2015-06-25
.travis.yml: not use versioned jruby

marutosi:


flavorjones:
Merged manually. Thank you!

PR #1306, opened 2015-06-25
.travis.yml: use jruby-9.0.0.0.rc2 instead of jruby 9.0.0.0.pre1

marutosi:


marutosi:
Covered by 3e98a5a5633a656.

PR #1318, opened 2015-07-18
fix appending text nodes before existing text nodes (jruby)

jkraemer:
While trying to fix at least some of Loofah's current test failures with JRuby I stumbled upon this.

I honestly do not understand what those 4 lines I removed from XmlNode are supposed to do - to me they look quite wrong since as a result the node to be inserted is simply discarded. Removing them fixed my test and caused no others to fail.

jruby 1.7.21 (1.9.3p551) 2015-07-07 a741a82 on OpenJDK 64-Bit Server VM 1.7.0_79-b14

### DIFF
--- a/ext/java/nokogiri/XmlNode.java
+++ b/ext/java/nokogiri/XmlNode.java
@@ -1635,10 +1635,6 @@ public class XmlNode extends RubyObject {
 
         Node nextSib = thisNode.getNextSibling();
 
-        if (nextSib != null &&
-            nextSib.getNodeType() == Node.TEXT_NODE &&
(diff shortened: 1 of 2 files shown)


marutosi:

marutosi:
I have confirmed this https://github.com/sparklemotion/nokogiri/pull/1318 reduces loofah jruby errors.
https://travis-ci.org/flavorjones/loofah/builds/80234142
https://travis-ci.org/marutosi/loofah/builds/90518033

marutosi:
What is status?

flavorjones:
Merged! Will be in v1.7.1. Sorry for the delay in getting to this, and thanks for your patience!

PR #1319, opened 2015-07-18
add test case checking decoration of empty NodeSet, which fails on JRuby

jkraemer:
It seems that empty NodeSets do not have any decorators applied to them on JRuby, while they indeed get decorated on Ruby. The problem seems to be that `initialize` in XmlNodeSet.java expects the first element of the node set as an argument, where it retrieves the document from which is then used for decoration. empty set -> no first element -> no decoration takes place. I have a test case illustrating the problem, but unfortunately no idea for a solution.

Loofah (and therefore also rails-html-sanitizer) has several test cases failing on JRuby because of this.

### DIFF
--- a/test/xml/test_document_fragment.rb
+++ b/test/xml/test_document_fragment.rb
@@ -219,6 +219,16 @@ module Nokogiri
         assert fragment.children.respond_to?(:awesome!), fragment.children.class
       end
 
+      def test_decorator_is_applied_to_empty_set
+        x = Module.new do
+          def awesome!
+          end
+        end
+        util_decorate(@xml, x)
+        fragment = Nokogiri::XML::DocumentFragment.new(@xml, "")
+        assert fragment.children.respond_to?(:awesome!), fragment.children.class
+      end
+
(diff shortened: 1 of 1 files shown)


marutosi:

marutosi:
As I wrote at https://github.com/flavorjones/loofah/pull/94#issuecomment-165411456, I cannot understand loofah and nokogiri support policy.

flavorjones:
Looking at this now.

flavorjones:
Just an update -- not sure how to fix this without rewriting all of NodeSet. And so the delay in fixing.

flavorjones:
@kares This is the second important JRuby issue, which is also blocking Loofah and Rails on JRuby.

I've pushed branch `1319_jkraemer_jruby-empty-nodeset-not-decorated` with the above failing test on it.

kares:
@flavorjones thanks, wondered about this as I stumbled at the doc retrieval code described in the issue 
-> believe it might need some internal refactoring, a 'quick' hack/fix might not be possible here

PR #1325, opened 2015-07-26
Install headers outside lib dir

swills:
This allows finding them via Gem::Specification.find_by_name('nokogiri').extension_dir which based on the docs seems to be the correct way.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -584,6 +584,43 @@ if ENV['CPUPROFILE']
   end
 end
 
+$INSTALLFILES = [ 
+  ['html_document.h', '$(archdir)'],
+  ['html_element_description.h', '$(archdir)'],
+  ['html_entity_lookup.h', '$(archdir)'],
+  ['html_sax_parser_context.h', '$(archdir)'],
+  ['html_sax_push_parser.h', '$(archdir)'],
+  ['nokogiri.h', '$(archdir)'],
+  ['xml_attr.h', '$(archdir)'],
+  ['xml_attribute_decl.h', '$(archdir)'],
+  ['xml_cdata.h', '$(archdir)'],
+  ['xml_comment.h', '$(archdir)'],
+  ['xml_document.h', '$(archdir)'],
+  ['xml_document_fragment.h', '$(archdir)'],
+  ['xml_dtd.h', '$(archdir)'],
+  ['xml_element_content.h', '$(archdir)'],
+  ['xml_element_decl.h', '$(archdir)'],
+  ['xml_encoding_handler.h', '$(archdir)'],
+  ['xml_entity_decl.h', '$(archdir)'],
+  ['xml_entity_reference.h', '$(archdir)'],
+  ['xml_io.h', '$(archdir)'],
+  ['xml_libxml2_hacks.h', '$(archdir)'],
+  ['xml_namespace.h', '$(archdir)'],
+  ['xml_node.h', '$(archdir)'],
+  ['xml_node_set.h', '$(archdir)'],
+  ['xml_processing_instruction.h', '$(archdir)'],
+  ['xml_reader.h', '$(archdir)'],
+  ['xml_relax_ng.h', '$(archdir)'],
+  ['xml_sax_parser.h', '$(archdir)'],
+  ['xml_sax_parser_context.h', '$(archdir)'],
+  ['xml_sax_push_parser.h', '$(archdir)'],
+  ['xml_schema.h', '$(archdir)'],
+  ['xml_syntax_error.h', '$(archdir)'],
+  ['xml_text.h', '$(archdir)'],
+  ['xml_xpath_context.h', '$(archdir)'],
+  ['xslt_stylesheet.h', '$(archdir)'],
+]
+
 create_makefile('nokogiri/nokogiri')
 
 if enable_config('clean', true)


swills:
I think the failure here was intermittent, is it possible to try it again?

flavorjones:
Hi @swills, thanks for the PR.

I'm unfamiliar with what `Gem::Specification.find_by_name('nokogiri').extension_dir` is supposed to do or why it's valuable. Can you please provide an explanation or link to some docs that will help me understand what this PR is trying to do?

swills:
Hi,

Yeah, it's a bit hard to find good docs on this. I can't find the thing I read when I submitted this, but here's the generic docs on it:

http://ruby-doc.org/stdlib-2.1.2/libdoc/rubygems/rdoc/Gem/Specification.html#method-c-find_by_name

The other doc which I can no longer find made it seem like .extension_dir (Gem::Specification.find_by_name('nokogiri').extension_dir) was the "right" way for gems to find .h files installed by other gems that they depend on.

This came up as I was looking at packaging gems in FreeBSD ports. One issue that came up is that intermediate build output such as .o files and other things were included in the packages, which seems sub-optimal (.so files are fine, but .o files normally aren't included in OS binary packages).

(I understand some would argue packaging gems is sub-optimal, but having packages of gems allows us to do things like "pkg install redmine" or "pkg install gitlab" and easily install such web apps, so let's leave that aside for now if we can.)

Anyway, as I was looking into changing the packages to delete these intermediate build outputs from packaged gems, I discovered that nokogumbo, which depends on finding some .h files from nokogiri looks in the same dir that the .o files were in. Since I'd cleaned out that directory, it couldn't find them. But after some research, I found that patching nokogiri as in this patch and then patching nokogumbo to find things by Gem::Specification.find_by_name('nokogiri').extension_dir instead of hard coding the path to the .h files. You can see the nokogumbo pull request here:

https://github.com/rubys/nokogumbo/pull/27

Just noticed the link there to this:

http://www.rubydoc.info/github/rubygems/rubygems/Gem/BasicSpecification#extension_dir-instance_method

Perhaps that's the doc I had found previously which made me think this was the right way for gems to find these things.

Hope that helps and thanks for taking a look. Let me know if you have more questions.

swills:
Did my answer help at all? Is there more info needed? Also, would it be possible to retry the failed build, I think it was an intermittent issue with the build rather than with the patch.

swills:
One other thing to note about this patch is that it only adds the files in the place where Gem::Specification.find_by_name('nokogiri').extension_dir can find them, doesn't remove them from the other location they get installed, so it shouldn't break anything for other consumers. Though it does use a tiny bit more disk space, hope that's not a concern.

swills:
Is there anything else I can do to help with this issue?

flavorjones:
@swills Thanks for your patience. I just need to find the time to context switch into opensource mode.

swills:
Ok, thanks. FWIW, the patch itself is fairly simple.

swills:
Should we remove the "needs more info" tag?

flavorjones:
@swills Just now getting time to look, I appreciate your patience.

I don't have any objection to adding this, but it's worth noting that there are no tests for this code. Because none of the core developers are using nokogiri the manner you're describing, it's going to be very easy for someone to break this by accident.

Can we add tests for this? If not, can it be autogenerated at build time?

flavorjones:
@swills Would love to chat about how we can write tests for this behavior.

flavorjones:
OK, I dug into this a bit today. I'll explain what I saw on my local dev machine, which I hope will serve as a guide for future rubyists.

I added this to an experimental `extconf.rb`:

```
$INSTALLFILES = [
  ["foo.txt", '$(archdir)'],
  ["bar.txt", 'bardir'],
]
```

(and touched two empty files in `ext/nokogiri`, and added those two files to my `Manifest.txt`).

When I installed the resulting gem, which installed to `/home/flavorjones/.rvm/gems/ruby-2.4.0/gems/nokogiri-1.7.0.1`, then the file `ext/nokogiri/foo.txt` was installed into two other places:

* `~/.rvm/gems/ruby-2.4.0/gems/nokogiri-1.7.0.1/lib/nokogiri/foo.txt`
* `~/.rvm/gems/ruby-2.4.0/extensions/x86_64-linux/2.4.0/nokogiri-1.7.0.1/nokogiri/foo.txt`

while `ext/nokogiri/bar.txt` was installed into only:

* `~/.rvm/gems/ruby-2.4.0/gems/nokogiri-1.7.0.1/ext/nokogiri/bardir/bar.txt`


On the same machine, the GemSpecification call you mention emits:

```
2.4.0 :001 > Gem::Specification.find_by_name('nokogiri').extension_dir
 => "/home/flavorjones/.rvm/gems/ruby-2.4.0/extensions/x86_64-linux/2.4.0/nokogiri-1.7.0.1" 
```

---

OK, so I think I've described the behavior correctly. @swills, is this the desired outcome?

flavorjones:
@swills bump?

flavorjones:
I've pinged the Nokogumbo maintainers to see if they're interested in having this merged.

swills:
@flavorjones Yep, I believe that's what we're looking for. FWIW, we've had this patch in FreeBSD ports for nearly 2 years without issues:

https://svnweb.freebsd.org/ports/head/textproc/rubygem-nokogiri/files/patch-ext_nokogiri_extconf.rb?view=log

flavorjones:
@swills @stevecheckoway I guess I'm looking for guidance from y'all on how I might test this to avoid breaking downstream libraries in the future. If you can help me with that (a script would be fine), super. If you can't, I'm not sure it's a good idea to merge this without some test coverage.

stevecheckoway:
@flavorjones I understand why you want tests; I'm not sure what the right way to test is. How do you test that the files you want are included in the gem in other cases? One thing that comes to mind is a test that uses the `Gem::Specification` API to look for the files you intend to distribute as part of the gem and/or extension. (Right now, the nokogiri gem contains a whole bunch stuff like object files and markdown files that don't seem necessary, but see below about useful files that get removed.)

In nokogumbo's case, we want to include `nokogiri.h` and `libxml2`'s `libxml/tree.h`. This evening, I came up with what I thought was a reasonable solution that didn't involve this patch:

``` ruby
ng_gem = Gem::Specification.find_by_name('nokogiri').gem_dir

# Search for libxml/tree.h in nokogiri's version of libxml2. If we can't find
# it there, check the system library.
have_libxml_tree_h = false
dependencies = YAML.load_file(File.join(ng_gem, 'dependencies.yml'))
libxml2_version = dependencies['libxml2']['version']
# Follow what nokogiri does and use host_alias if set.
host = RbConfig::CONFIG['host_alias'].empty? ?
  RbConfig::CONFIG['host'] : RbConfig::CONFIG['host_alias']
if find_header('libxml/tree.h',
               File.join(ng_gem, 'ports', host, 'libxml2', libxml2_version, 'include', 'libxml2'))
  have_libxml_tree_h = true
elsif have_library('xml2', 'xmlNewDoc')
  pkg_config('libxml-2.0')
  have_libxml_tree_h = have_header('libxml/tree.h')
end

# Search for nokogiri.h.
have_nokogiri_h = find_header('nokogiri.h', File.join(ng_gem, 'ext', 'nokogiri'))
```

For nokogiri 1.7.x (for whatever value of x I happened to have in my `Gemfile.lock`), this worked pretty well. It found `/path/to/gem/ports/x86_64-apple-darwin17.4.0/x.y.z/include/libxml2/libxml/tree.h` (for whatever version x.y.z of libxml2 was being linked) and then `/path/to/gem/ext/nokogiri/nokogiri.h`.

Unfortunately, after updating to nokogiri 1.8.4, this stopped working due to

``` ruby
    if enable_config('static')
      # ports installation can be safely removed if statically linked.
      FileUtils.rm_rf(root + 'ports', verbose: true)
    else
      FileUtils.rm_rf(root + 'ports' + 'archives', verbose: true)
    end
```

in `ext/nokogiri/extconf.rb`

One option would be to install the headers—both nokogiri's and those of the non-system libraries like libxml2 and libxslt—in the extension directory, similar to this PR. I don't know if it's possible to run tests on an installed gem which seems like what you'd need.

flavorjones:
> How do you test that the files you want are included in the gem in other cases

Well, this would be the first time we'd be testing that, so we're on a journey together to figure it out. :-D

It would be fine (to me) to run a test script against a packaged gem (potentially even installing it and inspecting the directory hierarchy post-installation). It doesn't need to run inside the `rspec` testing suite.

I don't think that inspecting the `Gem::Specification` will be very robust, as we do a lot of crazy stuff in `extconf.rb`, for example cross-compiling for windows, or modifying the `Specification` depending on whether you're building against JRuby or Ruby ... it's _possible_, I'm just not sure it would be _robust_ in the face of future changes to `extconf.rb`.

stevecheckoway:
I've been trying to get all of the headers to install in `$(archdir)/include` which puts them inside nokogiri's installed extension directory. The nokogiri headers themselves are easy to do. Easier than this PR, in fact. All that's necessary is

``` ruby
$INSTALLFILES << ['*.h', '$(archdir)/include']
```
What I haven't managed to do is get `mkmf` to install anything that's not in the `ext/nokogiri` directory.

The documentation for `mkmf` is lacking and the code isn't clear to me. It seems like it is only willing to install files that are in the current directory or in `$(srcdir)` (I don't know when those two are ever different in this case.) The code is here. https://github.com/ruby/ruby/blob/trunk/lib/mkmf.rb#L894

This code seems to work.
``` ruby
# Install the header files in the extension directory.
$INSTALLFILES << ['*.h', '$(archdir)/include']

unless using_system_libraries?
  # $INSTALLFILES only works for files in this directory for some reason so
  # copy the headers here first.
  require 'fileutils'
  FileUtils.rm_rf('include', secure: true)
  FileUtils.mkdir('include')
  [libxml2_recipe, libxslt_recipe].each do |recipe|
    FileUtils.cp_r(Dir[File.join(recipe.path, 'include/*')], 'include')
  end
  $INSTALLFILES << ['include/**/*.h', '$(archdir)']
end
```

If something like this is acceptable, we can figure out how to make a test for it and I can make a new PR.

stevecheckoway:
Here's a simple test script. It also demonstrates how one can use the `Gem::Specification` API to locate the headers.

``` ruby
# First install the nokogiri gem in /tmp/gem_path, then run this.
# E.g., from inside the nokogiri repo,
# $ bundle exec rake gem
# $ cd pkg
# $ gem install -i /tmp/gem_path nokogiri*.gem
# $ ruby nokogiri_header_test.rb

Gem::Specification.dirs = ['/tmp/gem_path']
ng_dir = Gem::Specification.find_by_name('nokogiri').extension_dir
%w[nokogiri.h libxml2/libxml/tree.h libxslt/xslt.h libexslt/exslt.h].each do |header|
  unless File.exists?(File.join(ng_dir, 'nokogiri/include', header))
    raise "Missing header #{header}"
  end
end
```

This should be easy to turn into a normal test, but installing the gem is pretty slow so I imagine this isn't something you'd want to do on every invocation of `rake test` It would also be good to test the system library configuration and check for the presence of `nokogiri.h` and the absence of the others.

flavorjones:
I'm closing this PR in favor of iterating on @stevecheckoway's #1788, which is a little bit cleaner.

I really appreciate everyone's time and attention, the conversation here is great and drove a very positive outcome in #1788 so it was a valuable effort.

PR #1326, opened 2015-08-03
Patching config.guess allows nokogiri installation on machines based on ppc64le architecture.

hernan-erasmo:
When this patch is applied, it updates config.guess file inside the temp dir where libxslt gets extracted before installation.


flavorjones:
Ah - I somehow missed this PR. Will take a look!

larskanis:
This looks good to me. The `config.guess` of the last libxslt release is ancient and no newer version is  available in the git repo. So I think patching like this is OK.

Tests run successfully [on Linux](https://travis-ci.org/larskanis/nokogiri/builds/79716033) and [on Windows](https://ci.appveyor.com/project/larskanis/nokogiri/build/1.0.15), when merged to master.

hernan-erasmo:
That's good to hear! You guys let me know if there's anything else that I can do.

Do you think it's possible that this patch makes it into 1.6.7 final?

larskanis:
@flavorjones  This PR patches `config.guess` to the version of 2015-07-03 but libxml2 uses an older version of 2013-06-10  (the [latest version](http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD) is 2015-09-14 currently). What is you opinion - should we patch libxslt to 2013-06-10 or patch libxml2 to 2015-07-03 as well or merge this and use two different versions of `config.guess` in libxml2 and libxslt?

hernan-erasmo:
(@flavorjones)

Following @larskanis idea, I've created PR #1352 that updates both libxslt and libxml2 config.guess files to the [latest version](http://git.savannah.gnu.org/gitweb/?p=config.git;a=blob_plain;f=config.guess;hb=HEAD)

IMHO this would be the most consistent approach, but in any case you can now choose between PR #1326 and PR #1352 and use the one you think it's more appropiate.

Hope you don't mind that I've opened 2 PR's for 2 different ways to solve the same problem :)

flavorjones:
@hernan-erasmo Thanks for this. I think I'd like to get this (or your other PR) into 1.6.7, since it seems to be turning into an "installation improvement" release. ;)

flavorjones:
Confirmed the contents of config.guess are what we expect, post-patch.

This isn't optimal, as it's persisting the *difference* between libxml2's current file and the latest from the `config` project, but it's good enough for now. When it breaks (and it will, whenever libxml2/libxslt update their config), we'll likely change this implementation to simply copy in the latest `config.guess` rather than try to patch what's there.

@hernan-erasmo: Have you sent a patch upstream to libxml2/libxslt to update their config.guess? If not, I'd like to ask that you do.

Merged into master now. Will be in 1.6.7.rc4.

PR #1333, opened 2015-08-13
fix over-eager namespace pruning when reparenting nodes

### LINKED ISSUE #1332: element in incorrect namespace after adding as a child
```
require 'nokogiri'

doc = Nokogiri::XML('<xml xmlns:a="abc"></xml>')
doc2 = Nokogiri::XML('<more xmlns="abc" />')
doc.root << doc2.root

puts doc
```

I would expect the output to be either

a) it used the outer namespace definition, and moved the element into it
```
<?xml version="1.0"?>
<xml xmlns:a="abc">
  <a:more/>
</xml>
```

or

b) it preserves the namespace that doesn't match on prefix, returning
```
<?xml version="1.0"?>
<xml xmlns:a="abc">
  <more xmlns="abc"/>
</xml>
```

but instead the output is
```
<?xml version="1.0"?>
<xml xmlns:a="abc">
  <more/>
</xml>
```

and more's namespace is no longer "abc".  This seems to be a problem with the `relink_namespace` function, that it removes duplicate namespaces, even if the prefixes don't match (and even if they did, you'd need to make sure the prefix wasn't redeclared between the higher definition and the lower definition)

ccutrer:
fixes #1332

### DIFF
--- a/ext/nokogiri/xml_node.c
+++ b/ext/nokogiri/xml_node.c
@@ -74,9 +74,11 @@ static void relink_namespace(xmlNodePtr reparented)
           reparented->parent,
           curr->href
       );
-      /* If we find the namespace is already declared, remove it from this
-       * definition list. */
-      if(ns && ns != curr) {
+      /* If we find the namespace is already declared with the same prefix,
+       * remove it from this definition list. */
+      if(ns && ns != curr &&
+         (ns->prefix == NULL && curr->prefix == NULL ||
+         xmlStrEqual(ns->prefix, curr->prefix))) {
         if (prev) {
           prev->next = curr->next;
         } else {
--- a/test/xml/test_node_reparenting.rb
+++ b/test/xml/test_node_reparenting.rb
@@ -237,6 +237,46 @@ module Nokogiri
                 assert @doc.at('//foo:second', "foo" => "http://flavorjon.es/")
               end
             end
+
+            describe "and a child node with a namespace matching the parent's non-default namespace, but with a different prefix" do
+              it "inserts a node that keeps its namespace" do
+                assert node = @doc.at('//xmlns:first')
+                child = Nokogiri::XML::Node.new('second', @doc)
+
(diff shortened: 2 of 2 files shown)


sgringwe:
I would love to see this pull request merged as it affects us too, any update on this?

sgringwe:
@larskanis is there any update on this? we are still locked into @ccutrer 's version and would love to of course be able to get updates from newer versions

larskanis:
This PR sounds reasonable to me. It doesn't fully align JRuby and MRI implementations, but brings it closer together.

@flavorjones Could you please have a look at this?

bradleybeddoes:
@larskanis @flavorjones any chance of this making it into 1.6.8? I'm also getting tripped up on this one.

flavorjones:
Yes, plan is to include in 1.6.8 final. Spent tonight going through old issues, will spend my next hack night finishing 1.6.8. Thanks for your patience.

flavorjones:
Apologies to just getting to this now -- the second test passes on master, can you help me understand why it's there?

Also, there are two conditionals being added to the `ns` check (both are NULL, or both are equal) and I'd like to make sure they're both being tested.

bradleybeddoes:
What is the process here for assisting on a PR proposed by someone else?. 

I am happy to try and assist resolve these questions but I don't want to impact/take any credit for the good work already done by @ccutrer.

bradleybeddoes:
Further to the above comment if it is helpful in any way I've proposed https://github.com/sparklemotion/nokogiri/pull/1444

bradleybeddoes:
@flavorjones did you have any further thoughts on this PR or #1444 ?

Apologies for being a pest about this, it is just a really important fix for a new project I need to ship asap.

flavorjones:
I merged one of these tests and the fix into master manually and given you credit in the commit log.

I apologize I couldn't find a better way to keep your name on the commits.

Thank you so much!

PR #1339, opened 2015-08-24
remove allow_failure jruby-9.0.0.0 from .travis.yml

marutosi:


knu:
Now what was the reason behind the excludes for osx?

knu:
Hmm, master's got somehow broken, but I have no idea as to why, since the commit should not affect other versions of jruby which failed.

flavorjones:
I'm looking at broken jruby builds on master, don't worry about it for now.

We excluded jrubies on OSX because I don't expect jruby to act differently on different platforms the same way I expect C compilers and extconf to act differently.

We excluded 2.2 on OSX because it wasn't available yet at the time. I'll try adding it back today and see whether Travis supports it.

flavorjones:
Just following up: 2.2 on OSX is working on travis now: https://travis-ci.org/sparklemotion/nokogiri

PR #1344, opened 2015-09-04
Update to mini_portile-0.7.0.rc5

larskanis:
This corresponds to https://github.com/flavorjones/mini_portile/pull/60 .


flavorjones:
As discussed with @larskanis we're instead going to make nokogiri depend on `mini_portile2`, so I'm closing this without merging.

PR #1355, opened 2015-09-22
fix error in validating files with jruby

twalpole:
This fixes the error under jrubies when validating invalid files.  Under jruby a filename being passed as a string was not being treated as a filename but rather as a document string.


flavorjones:
Ooh, thank you for this. I had just opened #1361 for this.

flavorjones:
Merged!

Issue #1155, opened 2014-09-04
xpath //namespace::* returns garbage -> can lead to segfault


etm:
warnings: []
    nokogiri: 1.6.3.1
    ruby:
      version: 2.0.0
      platform: x86_64-linux-gnu
      description: ruby 2.0.0p384 (2014-01-12) [x86_64-linux-gnu]
      engine: ruby
    libxml:
      binding: extension
      source: packaged
      libxml2_path: "/var/lib/gems/2.0.0/gems/nokogiri-1.6.3.1/ports/x86_64-pc-linux-gnu/libxml2/2.8.0"
      libxslt_path: "/var/lib/gems/2.0.0/gems/nokogiri-1.6.3.1/ports/x86_64-pc-linux-gnu/libxslt/1.1.28"
      compiled: 2.8.0
      loaded: 2.8.0

---------------------

Execute the following code for the file http://gruppe.wst.univie.ac.at/~mangler/SimTest.xml:

    require 'nokogiri'
    a = Nokogiri::XML::parse(File.open('SimTest.xml')){|config| config.noblanks.noent.nsclean.strict }
    b = a.xpath('//namespace::*').to_a
    puts b.length 

It tells that it finds 6464 namespaces. Most of them are binary garbage. i.e.

    #(Namespace:0x13bad48 {
       prefix = "pK\xC8\u0002",
       href = "\x90\xE6\x83\u0002"
       }),

Accessing them leads to segfaults. The first couple of namespaces are correct, but then (i suppose) it progresses into uncharted memory territory :-)

Funnily it works correctly for some XML files with namespaces. Funnily it works for some XML files with namespaces sometimes (heisenbug). The XML http://gruppe.wst.univie.ac.at/~mangler/SimTest.xml is the first one that i found that consistently lets me reproduce the error.

flavorjones:
Looking at this today with @akshaysawant ... think we have a fix.

flavorjones:
There's some history here, which I'll relate.

Originally @tenderlove committed dbda52b to try to plug a memory leak, but that was dereferencing possibly-invalid pointers during `deallocate` and was reverted.

Then, @ender672 committed 2675a75 which, though it didn't leak memory, introduced the bug you're seeing, which is that we're accessing the namespace nodes returned from an XPath query after the NodeSet has been GCed and freed the underlying namespace structs.

[...]

PR #1360, opened 2015-10-01
Correctly handle xmlNs memory in xpath query results

flavorjones:
This patch addresses #1155

This patch introduces a bifurcation in how Namespace objects are managed; Namespace objects wrapped around xmlNs structs from an xpath query now have their own Ruby object lifecycle and are GCed independently from their original document.

See commits and comments in xml_node_set.c and xml_namespace.c for more details.

Shout out to @akshaysawant who helped diagnose the underlying issue.

### DIFF
--- a/ext/nokogiri/xml_namespace.c
+++ b/ext/nokogiri/xml_namespace.c
@@ -2,6 +2,34 @@
 
 VALUE cNokogiriXmlNamespace ;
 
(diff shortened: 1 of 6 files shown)


flavorjones:
Also note that this is clean under valgrind, and doesn't leak memory related to namespace xpath queries.

PR #1362, opened 2015-10-01
Correctly handle xmlNs memory in xpath query results

flavorjones:
This patch addresses #1155

This patch introduces a bifurcation in how Namespace objects are managed; Namespace objects wrapped around xmlNs structs from an xpath query now have their own Ruby object lifecycle and are GCed independently from their original document.

See commits and comments in xml_node_set.c and xml_namespace.c for more details.

Shout out to @akshaysawant who helped diagnose the underlying issue.

Also note that this is clean under valgrind, and doesn't leak memory related to namespace xpath queries.

(This replaces #1360 and is rebased on a green commit.)

### DIFF
--- a/ext/nokogiri/xml_namespace.c
+++ b/ext/nokogiri/xml_namespace.c
@@ -2,6 +2,34 @@
 
 VALUE cNokogiriXmlNamespace ;
 
+static void dealloc_namespace(xmlNsPtr ns)
+{
+  /*
+   *
+   * this deallocator is only used for namespace nodes that are part of an xpath
+   * node set.
+   *
+   * see Nokogiri_wrap_xml_namespace() for more details.
+   *
+   */
+  NOKOGIRI_DEBUG_START(ns) ;
+  if (ns->href) {
+    xmlFree((xmlChar *)ns->href);
+  }
+  if (ns->prefix) {
+    xmlFree((xmlChar *)ns->prefix);
+  }
+  xmlFree(ns);
+  NOKOGIRI_DEBUG_END(ns) ;
+}
+
+
+int Nokogiri_namespace_eh(xmlNodePtr node)
+{
+  return (node->type == XML_NAMESPACE_DECL);
+}
+
+
 /*
  * call-seq:
  *  prefix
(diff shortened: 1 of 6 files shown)


ext/nokogiri/xml_node_set.c
```diff
@@ -128,29 +110,19 @@ static VALUE push(VALUE self, VALUE rb_node)
...
-    return Qnil ;
+  xmlNodeSetPtr node_set;
+  xmlNodePtr node;
+
+  Check_Node_Set_Node_Type(rb_node);
+
+  Data_Get_Struct(self, xmlNodeSet, node_set);
+  Data_Get_Struct(rb_node, xmlNode, node);
+
+  if (xmlXPathNodeSetContains(node_set, node)) {
```
larskanis: Looks like this condition is unnecessary now, since xpath_node_set_del() does the same check internally.

ext/nokogiri/xml_node_set.c
```diff
@@ -408,39 +370,103 @@ static VALUE unlink_nodeset(VALUE self)
...
+    if (cur->nodeTab[i] == val) break;
+
+  if (i >= cur->nodeNr) {	/* not found */
+    return;
+  }
+  cur->nodeNr--;
+  for (;i < cur->nodeNr;i++)
+    cur->nodeTab[i] = cur->nodeTab[i + 1];
+  cur->nodeTab[cur->nodeNr] = NULL;
+}
```
larskanis: This leaves a small memory leak, because the namespace nodes are not removed from the `@namespace_cache` . Not sure if it's worth to address it.

ext/nokogiri/xml_node_set.c
```diff
@@ -356,26 +320,25 @@ static VALUE slice(int argc, VALUE *argv, VALUE self)
...
-    else
-      elts[i] = Nokogiri_wrap_xml_node(Qnil, set->nodeTab[i]);
+  elts = (VALUE *)calloc((size_t)(node_set->nodeNr), sizeof(VALUE));
+  for(i = 0; i < node_set->nodeNr; i++) {
+    elts[i] = Nokogiri_wrap_xml_node_set_node(node_set->nodeTab[i], self);
+    rb_gc_register_address(&elts[i]);
   }
 
-  list = rb_ary_new4((long)set->nodeNr, elts);
+  list = rb_ary_new4((long)node_set->nodeNr, elts);
```
larskanis: I guess, that using `rb_ary_push()` once per node is faster than doing `rb_gc_register_address()` + `rb_gc_unregister_address()` for each node together with `rb_ary_new4()` .

ext/nokogiri/xml_namespace.c
```diff
@@ -34,21 +62,37 @@ static VALUE href(VALUE self)
   return NOKOGIRI_STR_NEW2(ns->href);
 }
 
+static int part_of_an_xpath_node_set_eh(xmlNsPtr node)
+{
+  return (node->next && ! Nokogiri_namespace_eh(node->next));
```
larskanis: Is this check really correct? I did not look into the libxml sources, but it looks strange to me.

larskanis:
Great work @flavorjones ! Looks like a correct solution to the GC issue. For sure much better than the previous way per `st_table` .

ext/nokogiri/xml_namespace.c
flavorjones: Agree that it's strange, but that's what libxml2 is doing. Which is why I wrapped it in a function with a semantic name.

flavorjones: See https://github.com/GNOME/libxml2/blob/master/xpath.c#L4190-L4191 for original source. Also check out the comments I made in xml_node_set.c in this series of patches.

For a better explanation, I'd recommend reading the patches in order, I tried to tell a story with them.

ext/nokogiri/xml_node_set.c
flavorjones: Well, the return value changes, and so we need to know whether the object is in the node set or not.

We could have `xpath_node_set_del` return a boolean (found or not-found), but honestly I'll take the performance hit in exchange for more readable code at this point, given how complicated namespace handling is.

It's rare that namespaces are returned in xpath queries anyway, and so I'd prefer to leave it as-is.

ext/nokogiri/xml_node_set.c
flavorjones: I'm not sure I understand what you mean. When would we remove namespace nodes from the cache?

ext/nokogiri/xml_node_set.c
flavorjones: Happy to talk about optimizing the `#to_array` method, but this code pre-exists the current patch, and so I'd prefer not to address that right now.

If you feel strongly we should look into optimizing this, can you open an Issue so we can discuss and benchmark?

flavorjones:
Merged manually (with conflicts).

ext/nokogiri/xml_node_set.c
larskanis: I opened https://github.com/sparklemotion/nokogiri/pull/1397 to address this.

PR #1365, opened 2015-10-13
Quick documentation update at java NokogiriHandler

jgpacker:
Now this method always calls start_element_namespace (which may call start_element if not overridden).


flavorjones:
Thanks for the PR!

Issue #1345, opened 2015-09-08
Nokogiri Build Failure: Bad Design


wjk:
Whenever I try to compile Nokogiri for OS X, I get a failure message because `/usr/include/iconv.h` is not found on my computer. This is because Nokogiri is hardcoding _that exact path_ into its `extconf.rb` script. Even though `iconv.h` is very much present (inside my Xcode SDK), and even though Nokogiri compiles _just fine_ when I manually patch `extconf.rb` to remove the check, the fact is that Nokogiri won’t play nice.

While I can manually patch Nokogiri to remove this shortsighted check, the fact is that Nokogiri is required in so many places that manually patching it every time is infeasible or impossible. Sadly, this check has been in place for so long that, even if you agree with me and decide to fix it, it won’t help me much because I have gems that depend on versions that have already been pushed that include the check, and that cannot be changed for stability reasons. Does anyone have any ideas as to how this could be resolved?

(This issue has bigger repercussions than simply the project I am working on. OS X El Capitan includes a feature that makes `/usr/include` permanently immutable, even by the root user. If this gem is not patched to accommodate this, it will become impossible to install on Macs that run OS X El Capitan.)

flavorjones:
Hi, @wjk. Thanks for bringing up this issue.

I believe what you're referring to was introduced in commit 8ed4486 to address #1111. This change made it into Nokogiri 1.6.4 and later. The message you receive should be letting you know that you need to run `xcode-select --install`, have you done that? If not, why not? If so, where is your iconv.h located? Can you provide some more information about your system? What version of XCode are you running, what version of OSX are you running, are you running homebrew (and what version), etc.?

It sounds like you have a better idea to fix #1111 than what we came up with -- can you be more specific about how you think we should be trying to detect iconv?

I'd like to make the installation experience better on OSX, but unfortunately at this point the ecosystem has become quite fractured and it's hard to handle every edge case. We're actively looking for help from the community to make Nokogiri better, so any information or thoughts you have on how to do this would be valuable to us. We're also trying to make this page the canonical reference for how-to-install:

> http://www.nokogiri.org/tutorials/installing_nokogiri.html

so once we figure out what's going on here, we'll make sure to update that page as well.

---

[...]

Issue #1231, opened 2015-01-22
OSX install problem (1.6.5, 1.6.6.2)


jjb:
```
➔ uname -a
Darwin johns-MBP 14.0.0 Darwin Kernel Version 14.0.0: Fri Sep 19 00:26:44 PDT 2014; root:xnu-2782.1.97~2/RELEASE_X86_64 x86_64
➔ rbenv version
2.1.5 (set by /Users/john/.rbenv/version)
➔ gem install nokogiri -v '1.6.6.1'
Fetching: nokogiri-1.6.6.1.gem (100%)
Building native extensions.  This could take a while...
ERROR:  Error installing nokogiri:
	ERROR: Failed to build gem native extension.

    /Users/john/.rbenv/versions/2.1.5/bin/ruby extconf.rb
checking if the C compiler accepts ... yes
checking if the C compiler accepts -Wno-error=unused-command-line-argument-hard-error-in-future... yes
Building nokogiri using packaged libraries.
checking for gzdopen() in -lz... yes
checking for iconv... yes
************************************************************************
IMPORTANT NOTICE:

Building Nokogiri with a packaged version of libxml2-2.9.2
with the following patches applied:
	- 0001-Revert-Missing-initialization-for-the-catalog-module.patch
	- 0002-Fix-missing-entities-after-CVE-2014-3660-fix.patch

Team Nokogiri will keep on doing their best to provide security
updates in a timely manner, but if this is a concern for you and want
to use the system library instead; abort this installation process and
reinstall nokogiri as follows:

    gem install nokogiri -- --use-system-libraries
        [--with-xml2-config=/path/to/xml2-config]
        [--with-xslt-config=/path/to/xslt-config]

If you are using Bundler, tell it to use the option:

    bundle config build.nokogiri --use-system-libraries
    bundle install

Note, however, that nokogiri is not fully compatible with arbitrary
versions of libxml2 provided by OS/package vendors.
************************************************************************
Extracting libxml2-2.9.2.tar.gz into tmp/x86_64-apple-darwin14.0.0/ports/libxml2/2.9.2... OK
Running patch with /Users/john/.rbenv/versions/2.1.5/lib/ruby/gems/2.1.0/gems/nokogiri-1.6.6.1/ports/patches/libxml2/0001-Revert-Missing-initialization-for-the-catalog-module.patch...
Running 'patch' for libxml2 2.9.2... OK
[...]

Issue #972, opened 2013-09-19
Install on Mac OSx with MacPorts fails to find libxml2


JESii:
Short output from gem install nokogiri is:

``` bash
Extracting libxml2-2.8.0.tar.gz into
tmp/x86_64-apple-darwin12/ports/libxml2/2.8.0... OK
Running 'configure' for libxml2 2.8.0... OK
Running 'compile' for libxml2 2.8.0... OK
Running 'install' for libxml2 2.8.0... OK
Activating libxml2 2.8.0 (from
/Users/jseidel/.rvm/gems/ruby-1.9.3-p327@rails3213/gems/nokogiri-1.6.0/ports
/x86_64-apple-darwin12/libxml2/2.8.0)...
Extracting libxslt-1.1.26.tar.gz into
tmp/x86_64-apple-darwin12/ports/libxslt/1.1.26... OK
Running 'configure' for libxslt 1.1.26... OK
Running 'compile' for libxslt 1.1.26... OK
Running 'install' for libxslt 1.1.26... OK
Activating libxslt 1.1.26 (from
/Users/jseidel/.rvm/gems/ruby-1.9.3-p327@rails3213/gems/nokogiri-1.6.0/ports
/x86_64-apple-darwin12/libxslt/1.1.26)...
checking for libxml/parser.h... yes
checking for libxslt/xslt.h... yes
checking for libexslt/exslt.h... yes
checking for iconv_open() in iconv.h... yes
checking for xmlParseDoc() in -lxml2... no

-----

libxml2 is missing.  please visit
http://nokogiri.org/tutorials/installing_nokogiri.html for help with
installing dependencies.
```

However, MacPorts says that's already installed:

``` bash
port installed | grep lib
libxml @1.8.17_0 (active)
libxml2 @2.9.1_0 (active)
```

What in the heck is going on here? It seems really weird to me that the
installer is compiling libxml2 and then can't find it!

I've followed the installation instructions at Nokigiri Installation and
that doesn't help. I also tried installing 1.5.9 and that failed the same way.

I've upgraded all ports, upgraded MacPorts...

petertorelli:
[...]

PR #1369, opened 2015-11-02
listen to --with-iconv-dir=<path> on OS X

kiwiroy:
## About

Among the issues touched by this are #972, #1119, #1231 and #1345 - not claiming they are fixed...

## Environment
Here's where I started debugging/installing and prepared this contribution from.

OS X:
```
% sw_vers
ProductName:       Mac OS X
ProductVersion: 10.10.5
BuildVersion:   14F27
```

Xcode:
```
% xcodebuild -version
Xcode 7.1
Build version 7B91b
% xcode-select -v
xcode-select version 2339.
```

Macports:
Spot that I started on the non-universal variants.
```
% port version
Version: 2.3.4
% port installed zlib lzma libiconv
The following ports are currently installed:
  libiconv @1.14_0
  libiconv @1.14_0+universal (active)
  lzma @4.65_0+universal (active)
  zlib @1.2.8_0
  zlib @1.2.8_0+universal (active)
```

## Install guide
@flavorjones you [asked](https://github.com/sparklemotion/nokogiri/issues/1119#issuecomment-68426814) for additions to OS X install guide.

`export ARCHFLAGS="-arch x86_64"` as diagnosed by @oliver-whiteman can help with:

```
Undefined symbols for architecture i386:
  "_crc32", referenced from:
      _xmlCreateZMemBuff in libxml2.a(xmlIO.o)
      _xmlZMemBuffAppend in libxml2.a(xmlIO.o)
<snip>
```
A better solution is to install the `universal` variants
```bash
sudo port install zlib +universal
sudo port install libiconv +universal
sudo port install lzma +universal
```
Sadly, this isn't the whole story as @oliver-whiteman found.
```
checking for main() in -llzma... yes
checking for xmlParseDoc() in libxml/parser.h... no
checking for xmlParseDoc() in -lxml2... no
checking for xmlParseDoc() in -llibxml2... no
-----
libxml2 is missing.  Please locate mkmf.log to investigate how it is failing.
```
As @rsanchezsaez mentioned this will affect **all** users of Macports with `zlib` and `libiconv` ports installed, where they are not the **universal** variants. The option of moving `/opt/local` sideways did not fill me with joy and `mkmf.log` shows this inability to find `iconv_open` and hence this PR.

### merge
So if this is merged the guide should reflect the requirement of:

```bash
sudo port install zlib +universal
sudo port install libiconv +universal
sudo port install lzma +universal
bundle config build.nokogiri --with-xml2-include=/opt/local/include/libxml2 --with-xml2-lib=/opt/local/lib --with-xslt-dir=/opt/local --with-iconv-dir=/opt/local
```

**or**

As has been pointed out, installation of `libxml2` and `libxslt` macports/brew and the use the `--use-system-libraries` or `NOKOGIRI_USE_SYSTEM_LIBRARIES=1`

Making note that the **universal** variants are installed.
```bash
sudo port install libxml2 +universal
sudo port install libxslt +universal
bundle config build.nokogiri --use-system-libraries --with-xml2-config=/opt/local/bin/xml2-config --with-xslt-config=/opt/local/bin/xslt-config
## <edit> as noted by @knu this is probably enough
bundle config build.nokogiri --use-system-libraries --with-opt-dir=/opt/local
## <edit> or even (given pkg-config installed and will find libxml-2.0.pc and libxslt.pc)
bundle config build.nokogiri --use-system-libraries
```

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -144,6 +144,8 @@ def iconv_libs
       if (config = preserving_globals{ dir_config(name) }).any?
         idirs, ldirs = config
         return ["-L#{ldirs}"]
+      else
+        return nil
       end
     }
   end
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -590,8 +590,13 @@ end
   'xmlRelaxNGSetValidStructuredErrors', 'xmlSchemaSetValidStructuredErrors',
   'xmlSchemaSetParserStructuredErrors'
 ].each { |func|
-  have_func(func, nil, iconv_libs) or
-    abort("#{func}() is missing")
+  if RUBY_VERSION <= '1.9.3'
+    have_func(func, nil) or
+      abort("#{func}() is missing")
+  else
+    have_func(func, nil, iconv_libs) or
+      abort("#{func}() is missing")
+  end
 }
 
 if ENV['CPUPROFILE']
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -134,6 +134,21 @@ int main(void)
   }
 end
 
+def iconv_libs
+  # OS X users with Macports zlib, lzma and libiconv
+  # these users _must_ have the +universal variants installed
+  if RUBY_PLATFORM =~ /darwin/
+    # If --with-iconv-dir or --with-opt-dir is given, it should be
+    # added when testing libxml (have_func / have_library) after activate
+    %w[iconv opt].each { |name|
+      if (config = preserving_globals{ dir_config(name) }).any?
+        idirs, ldirs = config
+        return ["-L#{ldirs}"]
+      end
+    }
+  end
+end
+
 def iconv_configure_flags
   # If --with-iconv-dir or --with-opt-dir is given, it should be
   # the first priority
@@ -452,7 +467,7 @@ else
       abort <<'EOM'.chomp
 -----
 The file "iconv.h" is missing in your build environment,
-which means you haven't installed Xcode Command Line Tools properly.
+which means you have not installed Xcode Command Line Tools properly.
 
 To install Command Line Tools, try running `xcode-select --install` on
 terminal and follow the instructions.  If it fails, open Xcode.app,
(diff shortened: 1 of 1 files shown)


knu:
Does this change anything?  Nokogiri itself does not require or link with iconv, but libxml2 does, so if you are using libxml2 built outside of Nokogiri there should be no need to mess with iconv.

ext/nokogiri/extconf.rb
```diff
@@ -563,18 +580,19 @@ def install
   "exslt" => ['exsltFuncRegister',      'libexslt/exslt.h'],
 }.each { |lib, (func, header)|
   have_func(func, header) ||
-  have_library(lib, func, header) ||
+  have_library(lib, func, header, iconv_libs) ||
```
kiwiroy: @knu - sure, but this line is the explicit issue, without a `-L/opt/local/lib` passed here compilation of `conftest.c` fails

kiwiroy:
...sigh... `have_func` on ruby <= 1.9.3 only accepts 2 args.

ext/nokogiri/extconf.rb
knu: If that's all you need, passing `--with-opt-lib=/opt/local/lib` will do, I think.

knu: Also, isn't there `pkg-config` in the PATH?  If you omit --with-xml2-* flags, extconf.rb should call pkg-config to read appropriate flags from the libxml-2.0 package configuration.

```ruby
  dir_config('xml2').any?  or pkg_config('libxml-2.0')
  dir_config('xslt').any?  or pkg_config('libxslt')
  dir_config('exslt').any? or pkg_config('libexslt')
```

kiwiroy: > If that's all you need, passing --with-opt-lib=/opt/local/lib will do, I think.

Is this when using `--use-system-libraries`?

`have_library` works fine then and yes I assume it is enough. When building `libxml-2.0` and `libxslt` it knows nothing and errors as per [this](https://github.com/sparklemotion/nokogiri/issues/1235#issuecomment-71496544).

kiwiroy:
The consequence of 36f1c82 is that OS X installs will still be in the same state for older rubies - though I have no knowledge of which was distributed in each version of OS X.

flavorjones:
I think that #1218, which was merged into master just before this PR came in, may address this problem in a different way.

Can you try building from source (master branch) and let me know if you still see this problem?

flavorjones:
1.6.7.rc4 just shipped, which it might be interesting to try to see if it affected any of the behavior you're describing.

kiwiroy:
Ok, this took me a while to remember, but I think that this can be closed.

I successfully built using:

* `bundle config build.nokogiri --use-system-libraries`
* `bundle config build.nokogiri --with-opt-dir=/opt/local` (macports prefix)

Macports has the universal variants of the following:

```bash
% port installed libxml2 libxslt zlib libiconv lzma
The following ports are currently installed:
  libiconv @1.14_0
  libiconv @1.14_0+universal (active)
  libxml2 @2.9.2_2
  libxml2 @2.9.2_2+universal (active)
  libxslt @1.1.28_1
  libxslt @1.1.28_1+universal (active)
  lzma @4.65_0+universal (active)
  zlib @1.2.8_0
  zlib @1.2.8_0+universal (active)
```

PR #1376, opened 2015-11-18
Add patch for unclosed comment uninitialized access issue

rafaelfranca:
See https://bugzilla.gnome.org/show_bug.cgi?id=746048 for more information about this security issue.


larskanis:
@rafaelfranca Thank you for the patch! We will merge it and release nokogiri-1.6.6.4.

PR #1380, opened 2015-11-20
Fixes test failures introduced in #1212

jvshahid:
There are few changes to the tests/assertions. I'd feel better if someone from @sparklemotion/nokogiri-core can review the changes.


test/test_xslt_transforms.rb
```diff
@@ -268,7 +270,7 @@ def test_non_html_xslt_transform
         <xsl:output encoding="UTF-8" indent="yes" method="xml" />
 
         <xsl:template match="/">
-          <xsl:value-of select="/a" />
+          <a><xsl:value-of select="/a" /></a>
```
jvshahid: @sparklemotion/nokogiri-core I'm not sure why libxml is not complaining about this transformation. the output is definitely not an xml document. instead of bending the java implementation to support libxml forgiveness, I changed the test to output proper xml. Any objections ?

test/xslt/test_exception_handling.rb
```diff
@@ -28,7 +28,7 @@ def test_java_exception_handling
           xsl.transform xml
           fail('It should not get here')
         rescue RuntimeError => e
-          assert_match(/HIERARCHY_REQUEST_ERR/, e.to_s, 'The exception message does not contain the expected information')
+          assert_match(/Can't have more than one root/, e.to_s, 'The exception message does not contain the expected information')
         end
```
jvshahid: It looks like the exception message has changed or may be the different implementations are throwing different exceptions. Either case, this change looks good to me and is more readable.

PR #1381, opened 2015-11-23
Ensure strings are zero terminated, where C-str pointers are expected.

larskanis:
Ruby-2.2 changes the String handling in such a way, that strings can
no longer be expected to be zero terminated. When zero terminated
strings are expected in the C code, StringValueCStr() must be used.
See: https://github.com/ruby/ruby/blob/v2_2_0/NEWS#L319

Using StringValuePtr() without a length check, can also become a
security issue, because it can leak the memory after the string.

This was already an issue in the pg.gem: https://github.com/ged/ruby-pg/pull/5


flavorjones:
Thank you for doing this!

PR #1393, opened 2015-12-09
Fix error when running rspec-rails tests

deivid-rodriguez:
I was getting the following error during `nokogiri` installation when running `rspec-rails` tests

```
extconf.rb:396:in `<main>': undefined method `gem' for main:Object
```

I think this should fix it.


deivid-rodriguez:
Closing since it's been fixed in the `rspec-rails` side.

PR #1397, opened 2015-12-18
Simplify NodeSet#to_a with a minor speed-up.

larskanis:
This stood up while reviewing https://github.com/sparklemotion/nokogiri/pull/1362#discussion_r47873435 . The speed-up is around 10%:

```ruby
xml = Nokogiri::XML.parse("<r>#{"<a/>"*100}</r>")
set = xml.xpath('//a')
st = Time.now; 1000000.times{ set.to_a }; Time.now - st
```

... changed from 8.2 to 7.3 seconds on my (slow) netbook.


flavorjones:
Great, thanks for following up and making this change!

PR #1385, opened 2015-11-27
Fix encoding tests for XML::Node, HTML::Node and XML::Document

larskanis:
See the commit messages for the full description.

This fixes the current CI test failures on Windows.

PR #1398, opened 2015-12-18
Fix test failure when Encoding.default_external is not UTF-8.

### LINKED ISSUE #1387: on master, some encoding tests fail on Windows
see comment on commit https://github.com/sparklemotion/nokogiri/commit/895c84936faa33ae29ac80af6aec49f08f2034b5 for background

larskanis:
This fixes a test failure on Windows, that was caused
by loading a HTML document per File.read(). This function
returns a string in some cpXXX encoding on Windows, which
was used as the document encoding later on, so that the test
failed.

This fixes #1387 and obsoletes #1385 .

### DIFF
--- a/test/xml/test_node_encoding.rb
+++ b/test/xml/test_node_encoding.rb
@@ -5,7 +5,7 @@ module Nokogiri
     class TestNodeEncoding < Nokogiri::TestCase
       def setup
         super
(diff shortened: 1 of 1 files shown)


flavorjones:
Wow. Thanks!

PR #1399, opened 2015-12-18
Fix encoding of xml node namespaces and improve encoding tests.

larskanis:
Fix encoding of keys returned by XML::Node#namespaces. The key parts were encoded like the document, but should be UTF-8 per definition. The value parts were already encoded as UTF-8.
    
Encoding of the various text returning functions was tested against the document encoding, but it should be UTF-8 per definition. The XML/HTML documents are now changed to a non-UTF-8 versions, to make the tests more meaningful.

This obsoletes #1385  - although the same content, the attached patches are better separated and so should be better to review, now.

### DIFF
--- a/lib/nokogiri/xml/node.rb
+++ b/lib/nokogiri/xml/node.rb
@@ -1,3 +1,4 @@
+# encoding: UTF-8
 require 'stringio'
 require 'nokogiri/xml/node/save_options'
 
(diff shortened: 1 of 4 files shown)


larskanis:
@flavorjones Could you please review this change?

larskanis:
I just rebased this PR to master.

flavorjones:
This looks good to me. Merging.

Issue #1370, opened 2015-11-06
Nokogiri fails to build with latest ruby:2.2.3 docker image


adrienkohlbecker:
Hello,

The following command: `docker pull ruby:2.2.3 && docker run -t -i ruby:2.2.3 gem install nokogiri -v '1.6.6.2'` results in an error.
That was not the case a few days ago, so I suspect [this commit](https://github.com/docker-library/ruby/commit/74ee8aec9c17ea2134db8a8ef199cf092c829576) fro two days ago on the base image may be at fault, but I've not been able to confirm it.

Here is the build log:

```
➜  ~  docker pull ruby:2.2.3 && docker run -t -i --rm ruby:2.2.3 gem install nokogiri -v '1.6.6.2'
Pulling repository docker.io/library/ruby
f08537117c4d: Download complete
575489a51992: Download complete
6845b83c79fb: Download complete
f9fffdafe16d: Download complete
d63502d4fbd8: Download complete
a3d4ddaffb90: Download complete
df231b622270: Download complete
3802d6035798: Download complete
c314b7cf3e15: Download complete
e9311cad9737: Download complete
8825dcf66d36: Download complete
34e5c4779386: Download complete
3b178d645c21: Download complete
177675c40a08: Download complete
e553b9534450: Download complete
2defd8698c56: Download complete
fb515cf11884: Download complete
Status: Image is up to date for ruby:2.2.3
Fetching: mini_portile-0.6.2.gem (100%)
Successfully installed mini_portile-0.6.2
Fetching: nokogiri-1.6.6.2.gem (100%)
Building native extensions.  This could take a while...
ERROR:  Error installing nokogiri:
	ERROR: Failed to build gem native extension.

    current directory: /usr/local/bundle/gems/nokogiri-1.6.6.2/ext/nokogiri
/usr/local/bin/ruby -r ./siteconf20151106-1-o83i7k.rb extconf.rb
checking if the C compiler accepts ... yes
Building nokogiri using packaged libraries.
checking for gzdopen() in -lz... yes
checking for iconv... yes
************************************************************************
IMPORTANT NOTICE:

Building Nokogiri with a packaged version of libxml2-2.9.2
with the following patches applied:
	- 0001-Revert-Missing-initialization-for-the-catalog-module.patch
	- 0002-Fix-missing-entities-after-CVE-2014-3660-fix.patch

Team Nokogiri will keep on doing their best to provide security
updates in a timely manner, but if this is a concern for you and want
[...]

PR #1405, opened 2016-01-05
Don't remove empty directories after build.

larskanis:
Deletion of these empty directories is flawed on docker with overlayfs, so we better avoid it. The build artifacts are still removed after gem install.

This fixes nokogiri issue #1370 . For further details to the root issue see: https://github.com/docker/docker/issues/19082


flavorjones:
Thanks for tracking this down.

shinnn:
@flavorjones Would you consider publishing v1.6.7.3 that includes this change? This change is really helpful for the developers who are using Docker-based CI services (e.g. wercker).

DeviaVir:
@flavorjones echo-ing what @shinnn asked, we're running docker in production and keep getting bitten by this bug, our workaround to downgrade rubygems for just nokogiri is no longer viable due to other deps.

PR #1411, opened 2016-01-18
Dont assume rubygems is loaded before using gem command

JonRowe:
This is a clarification of why #1393 was needed. It's not that `rspec-rails` fails without this, but any installation where rubygems is not available (we deliberately do this on rspec's suites for speed reasons).  It would be better to ensure rubygems is available before using it's components.


larskanis:
@JonRowe  Thanks, you're right - I'll add this to the documentation of mini_portile as well.

JonRowe:
@larskanis it's not the mini_portile usage thats the issue it's the `gem` method

JonRowe:
but thanks! :)

PR #1412, opened 2016-01-19
Work with ruby 2.3's --enable-frozen-string-literal

jeremyevans:
These changes are the minimal ones necessary to allow Sequel's specs
to pass. There may well be other changes that are required.


larskanis:
Thank you @jeremyevans ! I opened #1413 , which addresses the remaining issues. Unfortunately we can not test it on travis-ci, currently.

PR #1413, opened 2016-01-20
Work with ruby 2.3's --enable-frozen-string-literal - version 2

larskanis:
Although bundler is fixed on master in respect to the frozen string literal, no version was released with this fixes, so far. So the travis-ci tests fail before the tests can run, currently.

However when `rake test` is started without bundler, all tests pass.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,35 +3,37 @@ dist: trusty
 group: beta
 sudo: required
 
-rvm:
-- 2.0
-- 2.1
-- 2.2.5
-- 2.3.1
-- 2.4.0
-- jruby-1.7
-- jruby-9.1.5.0
-- rbx-3
-
-os:
-- linux
-- osx
-
 matrix:
-  allow_failures:
-  # https://github.com/travis-ci/travis-ci/issues/5361
+  include:
+  - os: linux
+    rvm: 2.0
+  - os: osx
+    rvm: 2.0
+  - os: linux
+    rvm: 2.1
+  - os: osx
+    rvm: 2.1
+  - os: linux
+    rvm: 2.2.5
   - os: osx
     rvm: 2.2.5
+  - os: linux
+    rvm: 2.3.1
   - os: osx
     rvm: 2.3.1
   - os: linux
-    rvm: rbx-3
-  exclude:
+    rvm: 2.4.0
+    env:
+      - RUBYOPT="--enable-frozen-string-literal --debug=frozen-string-literal"
   - os: osx
+    rvm: 2.4.0
+    env:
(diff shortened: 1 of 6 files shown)


flavorjones:
@larskanis Thanks for doing this work.

Are you sure that we want to test against 2.3.0 only with the frozen string option set?

I was thinking that perhaps we can refactor the travis matrix in a slightly different way: leave the existing matrix as it is, and then explicitly include two additional builds, which are linux/osx with the frozen string option set.

What do you think?

flavorjones:
Also, it looks like the rubygems we're using doesn't support frozen strings yet? Let's hold off on merging until the full toolchain supports it.

larskanis:
I usually prefer to name the tasks explicitly, that I distribute to someone else... However in this case I don't have a strong opinion. If you don't like it, I'll revert this change.

> Are you sure that we want to test against 2.3.0 only with the frozen string option set?

I thought it wouldn't have a valuable benefit, to have two tests on 2.3.0. The frozen-string-literal is an additional restriction, that we should meet (someday). So why should we additionally test for not applying this restriction?

Apart from that, this PR is work in progress. I already fixed [one issue in ruby](https://github.com/ruby/ruby/commit/fdcc924fce72b74e1e16881ac3b6e71a9b377b84) and noticed some remaining issues in bundler. This seems to be enough for nokogiri, but in order to run the bundler tests, a lot of fixes are required on rspec.

larskanis:
@flavorjones IMHO this is ready for merge now. Frozen strings are enabled for Ruby-2.4 only, because rubygems on Ruby-2.3 isn't compatible.

The test matrix for travis is re-arranged to explicit name the single test runs. At least for me this is much more readable than cross join the `rvm` and `os` entries and subtract the excludes from that matrix. In order to enable frozen strings per `env` variable this matrix would have to be extended by one dimension and all rubyies not supporting frozen strings would have to be added to the exclude list. So I think it's better to simply list all single test runs. Apart from that test runs are equally to the current on the master branch.

flavorjones:
@larskanis I'm in the process of moving our CI to Concourse, where we'll have more control over the crazy flavors of tests that we want to run; so I'm not worried about the Travis bit.

I can walk you through what the Concourse pipelines look like; but when I merge this I'll likely set up a separate pipeline to test frozen strings.

(You can get a peek at the pipelines at https://ci.nokogiri.org/ and the `/concourse/nokogiri.yml.erb` file.)

flavorjones:
Awesome! Merged, and I added the frozen string RUBYOPT options to the Concourse pipelines in 95684fd.

Thank you!

PR #1416, opened 2016-01-22
fix over-eager namespace pruning when reparenting nodes

### LINKED ISSUE #1332: element in incorrect namespace after adding as a child

afn:
fixes #1332

### DIFF
--- a/ext/nokogiri/xml_node.c
+++ b/ext/nokogiri/xml_node.c
@@ -74,9 +74,11 @@ static void relink_namespace(xmlNodePtr reparented)
           reparented->parent,
           curr->href
       );
-      /* If we find the namespace is already declared, remove it from this
-       * definition list. */
-      if(ns && ns != curr) {
+      /* If we find the namespace is already declared with the same prefix,
+       * remove it from this definition list. */
+      if(ns && ns != curr &&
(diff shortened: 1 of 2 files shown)


larskanis:
This is a dup of #1333 .

afn:
Oh, thanks - didn't realize there was already a PR for this!

PR #1417, opened 2016-01-24
more robust 'pkg-config' in extconf.rb

flavorjones:
in the case where `pkg-config` isn't installed, which appears to be the
case for lots of freebsd people, let's fall back to the pkg-config gem,
which knows how to parse .pc files.


flavorjones:
@larskanis would love your thoughts on this. works for me on freebsd and ubuntu.

flavorjones:
Merged.

PR #1431, opened 2016-02-22
download tar files in parallel

tenderlove:
We can download the tar files in parallel so that when it comes time to
compile libxslt, we don't have to wait for a download.  This should help
out people who have slow internet connections.

Down side is that the output looks weird because the progress bar is
being overwritten.

/cc @flavorjones

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -504,7 +504,9 @@ EOM
     } or abort 'zlib is missing; necessary for building libxml2'
   end
 
-  libxml2_recipe = process_recipe("libxml2", dependencies["libxml2"], static_p, cross_build_p) do |recipe|
+
+  xml2thread = Thread.new do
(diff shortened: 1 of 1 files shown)


larskanis:
Given that we already [bundle the tar files into the gem](https://github.com/sparklemotion/nokogiri/blob/master/Rakefile#L211) , this affects builds from git only. Moreover it starts both downloads in parallel (from the same server), so that the libxml2 build is delayed by the same amount of time, that libxslt is started earlier. The libxslt download should be started together with the lixml2 build, to get a speedup. But I'm not sure how valuable this feature is, when the gem users don't have a benefit from it.

flavorjones:
I agree with Lars and am a little confused by this. I've actively discouraged the community from using anything except gems (either final or RCs), and in that case no downloads are being executed.

Help me understand? What am I missing?

PR #1437, opened 2016-03-01
Fix #1436 - compatibity with gem-compiler.

### LINKED ISSUE #1436: gem-compiler fails to run on nokogiri
```
$ gem fetch nokogiri
Fetching: nokogiri-1.6.7.2.gem (100%)
Downloaded nokogiri-1.6.7.2

$ gem compile nokogiri-1.6.7.2.gem
Unpacking gem: 'nokogiri-1.6.7.2' in temporary directory...
Building native extensions.  This could take a while...
WARNING:  See http://guides.rubygems.org/specification-reference/ for help
ERROR:  While executing gem ... (Gem::InvalidSpecificationException)
    ["ports/archives/libxml2-2.9.2.tar.gz", "ports/archives/libxslt-1.1.28.tar.gz"] are not files
```

See related:
- https://github.com/sparklemotion/nokogiri/pull/1275
- https://github.com/luislavena/gem-compiler/issues/12

The key commit here is 12759cf which, though well-intended, deletes files declared in the gemspec manifest, and so breaks an implicit contract that gem-compiler depends upon.

@knu and @larskanis I don't think we want to delete the source tarballs. Thoughts?

larskanis:
The extconf.rb clean task removes most of the files, that are used only while build. However this broke gem-compiler, because it removed the archives as well, although they are part of the gemspec.

With this patch the archives are kept on disk. The downside is, that it increases the disk space used by nokogiri from 21MB to 30MB after install.

An alternative is gem-compiler's option `--prune`, as described here: https://github.com/luislavena/gem-compiler/issues/12

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -61,9 +61,10 @@ def do_clean
 
     if enable_config('static')
       # ports installation can be safely removed if statically linked.
-      FileUtils.rm_rf(root + 'ports', verbose: true)
-    else
-      FileUtils.rm_rf(root + 'ports' + 'archives', verbose: true)
+      Pathname.glob(root.join('ports', '*')) do |dir|
+        next if dir.fnmatch?('*/archives')
+        FileUtils.rm_rf(dir, verbose: true)
+      end
     end
   end
 


larskanis:
I think the `--prune` option to gem-compiler is a acceptable solution, so that we can close this PR.

PR #1439, opened 2016-03-02
Frozen String Literal Compatibility

twalpole:
Turns on frozen string literals in 2.3.0+ and fixes code for it.

### DIFF
--- a/lib/nokogiri.rb
+++ b/lib/nokogiri.rb
@@ -115,15 +115,19 @@ module Nokogiri
     def Slop(*args, &block)
       Nokogiri(*args, &block).slop!
     end
+
+    def install_default_aliases
+      # Make sure to support some popular encoding aliases not known by
+      # all iconv implementations.
+      {
+        'Windows-31J' => 'CP932',	# Windows-31J is the IANA registered name of CP932.
+      }.each { |alias_name, name|
+        EncodingHandler.alias(name, alias_name) if EncodingHandler[alias_name].nil?
+      }
+    end
   end
 
-  # Make sure to support some popular encoding aliases not known by
-  # all iconv implementations.
-  {
-    'Windows-31J' => 'CP932',	# Windows-31J is the IANA registered name of CP932.
-  }.each { |alias_name, name|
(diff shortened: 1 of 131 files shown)


flavorjones:
@larskanis I know you've already put thought and work into frozen strings ... can you comment?

twalpole:
Sorry -- I didn't see @larskanis work towards this.  One thing to note is that I used `String.new("")`  rather than `String.new` for a new empty string.  One would think they would be identical however it turns out the first produces a string with the encoding of the document  - the second has an encoding of ASCII-8BIT for some reason.

twalpole:
@flavorjones The second commit here fixes the random failures that have been occurring on PRs, caused by test order and when the encoding alias added by default was cleared.  If you don't want the frozen literal changes you may want to cherrypick the second commit separately

larskanis:
I don't think we want to pollute all the nokogiri files with this magic comment.

@twalpole The difference between `String.new` and `String.new("")` is because the first one doesn't have a source file context, so it can not get any encoding information. The latter one don't build a string with the encoding of the XML document, but with the encoding of the surrounding source file. However this difference doesn't change the nokogiri-API at all.

IMO we can close this in favor of #1413 .

twalpole:
The "pollution" of the magic comment allows nokogiri to gain the benefits of default frozen literals before every other gem a project (using nokogiri) depends on being updated to allow someone to use the ruby switch.  

Either way - the second part of the commit attempts to fix the random invalid encoding test failures that have been occuring, so might be worth cherrypicking

@larskanis Yes, by document I meant the file it was in, not the XML document - updated comment to clarify, and I would say String.new does have a context (the file it's called in) although I guess the default value passed to it is declared in the source file.  It may not change the Capybara API but it does mean if you then immediately concat to the new string there are encoding translations that need to be made and you end up with a string that is not in the encoding one would generally assume (if just creating a literal)

larskanis:
@twalpole Sorry, I missed the second commit, but cherry-picked it to master, now.

flavorjones:
Closing, frozen string support is now on master.

PR #1444, opened 2016-03-08
Ensure namespace pruning is not over zealous

bradleybeddoes:
Namespace pruning would previously consider the same namespace URI as
equivalent even though the prefix in use may have differed.

Thus child nodes added with a different prefix for an already known
namespace would cause the resulting document to become schema invalid.


bradleybeddoes:
@flavorjones any further thoughts on merging this or #1333 for the next 1.6.8 RC? 

Any idea on timelines if so?

flavorjones:
I've merged some of your tests from this PR into master, and I've given you credit in the commit log.

Apologies I couldn't figure out a better way to keep your name on the commits.

Thank you so much!

PR #1451, opened 2016-03-26
Made documentation better for ParseOptions

ashivasubramanian:
Hello,

We were working with Nokogiri, when we found that the ParseOptions class was hard to understand. In the [documentation](http://www.rubydoc.info/github/sparklemotion/nokogiri/Nokogiri/XML/ParseOptions), available methods were not documented, so we thought we can improve it. Hope you accept it.

Regards,
Shivasubramanian A


flavorjones:
Thank you for your contribution!

PR #1478, opened 2016-06-05
Double quotes in HTML attributes should be escaped as &quot; not %22

### LINKED ISSUE #1382: Double quotes URL-escaped in attribute when using JRuby
Hi!

When using JRuby to parse, modify, and output an HTML document with double quotes in an attribute, the double quote is URL-escaped as `%22`.  I believe this is wrong behaviour.  The double quote should be left alone as it is when using MRI.

```
$ rvm use jruby-1.7.22
Using ~/.rvm/gems/jruby-1.7.22
$ irb
jruby-1.7.22 :001 > require 'nokogiri'
 => true 
jruby-1.7.22 :002 > puts Nokogiri::HTML(%q{<tag attr='"' />}).to_html
<html><head></head><body><tag attr="%22"></tag></body></html>
 => nil 
```

```
$ rvm use jruby-9.0.4.0
Using ~/.rvm/gems/jruby-9.0.4.0
$ irb
jruby-9.0.4.0 :001 > require 'nokogiri'
 => true 
jruby-9.0.4.0 :002 > puts Nokogiri::HTML(%q{<tag attr='"' />}).to_html
<html><head></head><body><tag attr="%22"></tag></body></html>
 => nil
```

```
$ rvm use 2.2.2
Using ~/.rvm/gems/ruby-2.2.2
$ irb
2.2.2 :001 > require 'nokogiri'
 => true 
2.2.2 :002 > puts Nokogiri::HTML(%q{<tag attr='"' />}).to_html
<!DOCTYPE html PUBLIC "-//W3C//DTD HTML 4.0 Transitional//EN" "http://www.w3.org/TR/REC-html40/loose.dtd">
<html><body><tag attr='"'></tag></body></html>
 => nil
```

donv:
Hi!

This fixes #1382 , or aims to.  I have searched a lot, but I cannot find any source claiming that double quotes should be escaped as `%22` in HTML attributes.  `&quot;` seems to be the common escape sequence, although there are others.

I have not found where the C-based Nokogiri does the escaping, yet.  I'd like to fix that too, but I could need some pointers.

As described in issue #1382 , the C-based Rubies work as expected in IRB, but they fail the modified unit test (test/html/test_node.rb#test_get_attribute).  They still escape the double quote to `%22`.

Anyway, I'm looking forward to the discussion on this PR.  I will learn something either way. 😄

### DIFF
--- a/ext/java/nokogiri/XmlAttr.java
+++ b/ext/java/nokogiri/XmlAttr.java
@@ -129,9 +129,7 @@ public class XmlAttr extends XmlNode{
             case '\n': buffer.append("&#10;"); break;
             case '\r': buffer.append("&#13;"); break;
             case '\t': buffer.append("&#9;"); break;
-            case '"': if (htmlDoc) buffer.append("%22"); 
-                else buffer.append("&quot;");
-                break;
+            case '"': buffer.append("&quot;"); break;
             case '<': buffer.append("&lt;"); break;
             case '>': buffer.append("&gt;"); break;
             case '&': buffer.append("&amp;"); break;
--- a/ext/java/nokogiri/internals/SaveContextVisitor.java
+++ b/ext/java/nokogiri/internals/SaveContextVisitor.java
@@ -312,9 +312,7 @@ public class SaveContextVisitor {
             case '\n': buffer.append("&#10;"); break;
             case '\r': buffer.append("&#13;"); break;
             case '\t': buffer.append("&#9;"); break;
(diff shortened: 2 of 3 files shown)


flavorjones:
Hi,

Thanks for submitting this. Please see my comments on #1382 about this issue, and canonical documents in general.

donv:
Thanks.  Using #canonicalize seems to solve my problems.

Issue #1496, opened 2016-06-24
Please drop the pkg-config dependency from the precompiled gems


voxik:
Not sure if that is possible, but there is no reason the precompiled versions of gem should depend on pkg-config. The java version does not depend on it, so hopefully this dependency can be avoided also for the other platforms.

flavorjones:
There's another issue opened objecting to the license, and so we may remove
it entirely. Stay tuned.

md5:
@voxik The other issue is #1488

flavorjones:
Will be fixed in 1.6.8.1

PR #1498, opened 2016-06-25
Make the pkg-config gem optional.

### LINKED ISSUE #1488: Nokogiri LGPL-2.1?
In closing issue #1417, you have added the pkg-config gem which is LGPL-2.1.

Certainly this should be an optional gem for those who cannot have the LGPL license in their projects.

larskanis:
Using the pkg-config gem as a runtime dependency of nokogiri results in a license conflict. pkg-config is LGPL but nokogiri is MIT.

Making the pkg-config gem optional solves this issue.

Fixes #1488 and #1496 .

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -5,7 +5,6 @@
 source "https://rubygems.org/"
 
 gem "mini_portile2", "~>2.1.0"
-gem "pkg-config", "~>1.1.7"
 
 gem "rdoc", "~>4.0", :group => [:development, :test]
 gem "hoe-bundler", "~>1.2.0", :group => [:development, :test]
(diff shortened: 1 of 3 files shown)


larskanis:
Unfortunately https://github.com/sparklemotion/nokogiri/commit/caee481d095e974df63386db63d14599afe09869 broke the OSX setup on Travis-CI. Revert it?

tenderlove:
+1 on making this optional.  Like to avoid LGPL if possible.

md5:
This looks like a reasonable way to deal with making `pkg-config` optional. Can this be merged so that Nokogiri no longer has an LGPL dependency?

flavorjones:
Will be fixed in v1.6.8.1.

PR #1517, opened 2016-07-27
Prefer libxml2's strdup and strchr equivalents

jeremy:
Background: Working around a puzzling segfault when using libc strdup on a C string from Ruby. Returns a pointer that's outside the program heap entirely.

Turns out libxml2 provides its own strdup/strchr equivalents that use the mem management functions we pass it (including ruby's malloc, ruby_strdup, etc), so may as well prefer these anyway.

### DIFF
--- a/ext/nokogiri/xml_node.c
+++ b/ext/nokogiri/xml_node.c
@@ -833,37 +833,40 @@ static VALUE set(VALUE self, VALUE property, VALUE value)
 static VALUE get(VALUE self, VALUE rattribute)
 {
   xmlNodePtr node;
-  xmlChar* value = 0;
-  VALUE rvalue ;
-  char* attribute = 0;
-  char *colon = 0, *attr_name = 0, *prefix = 0;
+  const xmlChar *value = 0;
+  VALUE rvalue;
+  xmlChar *colon;
+  const xmlChar *attribute, *attr_name, *prefix;
   xmlNsPtr ns;
 
   if (NIL_P(rattribute)) return Qnil;
 
   Data_Get_Struct(self, xmlNode, node);
-  attribute = strdup(StringValueCStr(rattribute));
+  attribute = xmlCharStrdup(StringValueCStr(rattribute));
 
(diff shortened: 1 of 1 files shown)


tenderlove:
Background: it seems like some platforms [may not actually behave nicely if you use `strdup` from libc](https://stackoverflow.com/questions/8359966/strdup-returning-address-out-of-bounds).  So we should probably try to avoid using `strdup`.

flavorjones:
Thanks for submitting this!

Do you have a way to reliably reproduce the segfault? I'd like to verify
the expected behavior and potentially add something to the test suite to
prevent regression.

flavorjones:
Hmm, ok, now that I'm looking at the diff maybe that's not necessary.

jeremy:
@flavorjones Confirm; the segfault is unrelated. I'm seeing libc `strdup` intermittently return total nonsense in… some cases and that happens to trigger a segfault in `strchr` since it gets a pointer to memory it can't access.

Using the `xmlStrdupFunc` we provided to `xmlMemSetup`—Ruby's `ruby_strdup`—happens to work around this issue as a side effect.

flavorjones:
@jeremy I'm not sure if you still think this should be merged? Your most recent comment above seems to indicate that the problem you're trying to solve may be unrelated?

jeremy:
Yes, it should. I've been running this in production for months now and it completely solves the issue I hit.

Here's the problem I was troubleshooting: "using libc `strdup` on a C string from Ruby was returning a pointer that's outside the program heap entirely." This then caused `strchr` to segfault.

This patch obviates the root issue with `strdup` by using libxml's mem management API which internally delegates to Ruby mem handling routes. That fixes the strdup issue I hit in certain environments, which means we never hit the strchr segfault.

This is good hygiene as a user of libxml in any case, too!

(That said, being able to segfault at all here is troubling. We aren't doing any error checking.)

flavorjones:
@jeremy Thanks so much for submitting this and for persisting. Merging!

PR #1526, opened 2016-08-15
Change code-climate badge to make it retina friendly

crispgm:
The ```.png``` badge is not retina friendly on Mac. Instead, I changed it to ```.svg```, which Code Climate officially supports as well and has the same content.


flavorjones:
Thank you!
b

PR #1538, opened 2016-09-26
Removing unnecessary <code> tags

ashivasubramanian:
The &lt;code&gt; tags are not necessary to get code formatting - the spaces in front of the line are good enough.


ashivasubramanian:
The build fails, but it doesn't appear related to the changes made, since they are only documentation changes.

flavorjones:
Thank you!

Issue #1515, opened 2016-07-21
Different compiler needed on OpenBSD


frenkel:
On OpenBSD you need to install a newer compiler to get Nokogiri 1.6.8 to compile. This newer compiler is called `egcc` (for external I think). It would be great if Nokogiri's install process could detect it, so a normal `bundle install` would run without problems. Currently we need to manually connect to the server and run `CC=egcc bundle install`. Would you be willing to add such an exception/workaround? If so, what would be the best place to put this? I can then make a pull request for it.

Thanks and keep up the good work!

flavorjones:
I'm not sure I understand why this would be happening.

If you take a look at `ext/nokogiri/extconf.rb` you'll see we compile everything using `RbConfig::MAKEFILE_CONFIG['CC']`, which should be the same compiler used to compile Ruby itself.

Can you explain a bit about why "you need to install a newer compiler to get Nokogiri 1.6.8. to compile"? I'd like to understand a bit better the problem you're trying to solve.

frenkel:
We install ruby from packages, and it was compiled using gcc 4.2 which is the default compile (i.e. `CC=gcc`). Apparently the newer libxml needs to be compiled with something newer, so we need to override the `CC` environment variable and use `egcc` which is gcc 4.9. It would be great when I can extend nokogiri with a check that it will use egcc when on OpenBSD, because otherwise you will always have to override CC manually.

flavorjones:
Hi, i unfortunately still don't understand the problem you're trying to solve.

Can you show me what happens when you compile using the default settings? What breaks? Can you show the logs?

frenkel:
The problem is simple: default gcc in OpenBSD is too old to compile libxml2. It gives this error during install of nokogiri:

```
Extracting libxml2-2.9.4.tar.gz into tmp/x86_64-unknown-openbsd6.0/ports/libxml2/2.9.4... OK
Running 'configure' for libxml2 2.9.4... OK                                     
Running 'compile' for libxml2 2.9.4... ERROR, review '/home/frank/.gem/ruby/2.2/gems/nokogiri-1.6.8/ext/nokogiri/tmp/x86_64-unknown-openbsd6.0/ports/libxml2/2.9.4/compile.log' to see what happened. Last lines are:
========================================================================        
/usr/include/unistd.h:404: warning: redundant redeclaration of 'ftruncate'      
/usr/include/sys/types.h:217: warning: previous declaration of 'ftruncate' was here
[...]

PR #1543, opened 2016-10-05
Use egcc on OpenBSD to fix libxml compilation

frenkel:
First PR for OpenBSD support. #1515 it was suggested that I should also create a PR for the documentation. I'll submit that to the other repo in a few minutes.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -20,6 +20,10 @@ def darwin?
   RbConfig::CONFIG['target_os'] =~ /darwin/
 end
 
+def openbsd?
+  RbConfig::CONFIG['target_os'] =~ /openbsd/
+end
+
 def nix?
   ! (windows? || solaris? || darwin?)
 end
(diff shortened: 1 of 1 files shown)


parhs:
This is a good fix. However I precompiled with CC=egcc in openbsd 5.9  but in 6.0 it can't be loaded, it needs to be recompiled

frenkel:
Yes, that's always when library versions change (which happens quite a lot between OpenBSD releases).

parhs:
So is there a solution to have both precompiled ?

frenkel:
No, when you run OpenBSD 6.0 you don't know what 6.1 will contain, so you can only compile it for your current version. Same problem is on Mac OS X when Apple decides to update some important library dependency.

ext/nokogiri/extconf.rb
```diff
@@ -387,7 +391,16 @@ def using_system_libraries?
   do_clean
 end
 
-RbConfig::MAKEFILE_CONFIG['CC'] = ENV['CC'] if ENV['CC']
```
knu: What about leaving this unchanged and put `if openbsd? && !ENV['CC'] ... ENV['CC'] = find_executable 'egcc' end` above it so you could still override `CC`?

frenkel: Good point! I've changed it.

knu:
Out of curiosity, is clang not a choice on OpenBSD?  I heard it started to bundle clang starting from 6.0.

frenkel:
Well, they imported in CVS after the 6.0 release, but I believe they won't make it a supported choice in 6.1. Clang can be installed on amd64 using packages in 5.9 and 6.0 already though.

knu:
Committed, thanks.

PR #1545, opened 2016-10-10
Avoid Fixnum deprecation warning on Ruby trunk

matthewd:
These conditionals don't seem to particularly care whether the value is a Bignum, so it looks like it's safe to just switch to Integer... the only result is that ParseOptions could see some additional unused bits set.


larskanis:
Makes sense. Thank you!

PR #1558, opened 2016-11-12
Determine the minimum ruby version for fat binary gems from the cross…

larskanis:
… ruby versions.

This is more exactly and avoids duplication with source gem definition.

While being over it, also determine the versions per-platform.


larskanis:
Setting ruby version constraints for binary gems is part of rake-compiler now. I'll update nokogiri accordingly.

PR #1563, opened 2016-12-23
jruby review - contention + cloning objects

kares:
this is some Java code cleanup along the line of some issues, notably:

- Nokogiri's class-cache was synchronized yet read-only after initialization (un-wanted contention)
- XPathFunction related internals used clone() on non-clonable objects thus always throwing
-> caused an unnecessary slow-down (`catch (CloneNotSupportedException e)`)


jvshahid:
lgtm

kares:
see JRuby 1.7 (still around and supported) was dropped, hoped for a Nokogiri release with these for 1.7
there's still likely to be at least one JRuby 1.7.x (patch) release.
any chance we can do a Nokogiri release for JRuby (there's very little change in Nokogiri 1.7.0) ?

flavorjones:
@kares Not sure what you're referring to when you say "jruby 1.7 was dropped." Definitely not the case, we still [support it](https://github.com/sparklemotion/nokogiri/blob/master/test_all#L20) and still [run CI on it](https://travis-ci.org/sparklemotion/nokogiri/builds/191123286).

I'm merging this now, so it will be in our next release, likely to be Nokogiri 1.7.1. Sound good?

kares:
Thanks Mike, have seen your tweet for the 1.7 release and it wasn't clear that you're not dropping JRuby 1.7 while dropping MRI < 2.1 ... now I see the -java gem does not have `required_ruby_version => '>= 2.1.0'`

PR #1574, opened 2017-01-04
Look for bundled libs in lib64 else lib

mbixby17451:
This adds awareness of the lib64 build convention. On openSUSE, MacOS, and perhaps other distros, the bundled libxml2 and libxslt libraries get built in a lib64 subdirectory per distro convention.  This commit changes Nokogiri's hardcoded 'lib' convention to first check for bundled library existence in lib64, and if not found, falls back to lib.

See issue #1562 for details.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -628,9 +628,11 @@ EOM
     $libs = $libs.shellsplit.map do |arg|
       case arg
       when '-lxml2'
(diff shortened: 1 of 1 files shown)


flavorjones:
Hi. Thanks for submitting this PR!

As I commented on #1562, I haven't yet managed to reproduce what you're seeing, and would prefer not to try to evaluate a PR until I more completely understand the problem being solved.

I'm going to close this PR for now; but I'm open to re-evaluating that decision once I better understand the problem we're trying to solve.

Thanks again!

PR #1576, opened 2017-01-05
Add tests for HTML SAX parsing of br tag from string and io

gregors:
I've added a test to show a bug in HTML SAX parsing. When parsed from a simple string it works  fine. When parsed from IO it fails on the first invalid tag.


flavorjones:
Going to manually merge this commit as part of my fix for #1577. Will give you a shout-out in the CHANGELOG! Thanks.

Issue #1586, opened 2017-01-15
[JRuby] XML::Reader does not return errors


flavorjones:
**What problems are you experiencing?**

`XML::Reader` does not return parse errors on JRuby (though it does with libxml2/MRI).

**What's the output from `nokogiri -v`?**

```
# Nokogiri (1.7.0)
    ---
    warnings: []
    nokogiri: 1.7.0
    ruby:
      version: 2.3.1
      platform: java
      description: jruby 9.1.6.0 (2.3.1) 2016-11-09 0150a76 OpenJDK 64-Bit Server VM 25.111-b14 on 1.8.0_111-8u111-b14-2ubuntu0.16.04.2-b14 [linux-x86_64]
      engine: jruby
      jruby: 9.1.6.0
    xerces: Xerces-J 2.11.0
    nekohtml: NekoHTML 1.9.21
```

**Can you provide a self-contained script that reproduces what you're seeing?**

these tests are in `test/xml/test_reader.rb` already, I've ommitted the `skip`:

```
      def test_errors_is_an_array
#        skip "TODO: xerces does not internally call `errors`" if Nokogiri.jruby?
        reader = Nokogiri::XML::Reader(StringIO.new('&bogus;'))
        assert_raises(SyntaxError) {
          reader.read
        }
        assert_equal [SyntaxError], reader.errors.map(&:class)
      end

      def test_pushing_to_non_array_raises_TypeError
#        skip "TODO: xerces does not internally call `errors`" if Nokogiri.jruby?
        reader = Nokogiri::XML::Reader(StringIO.new('&bogus;'))
        def reader.errors
          1
        end
        assert_raises(TypeError) {
          reader.read
        }
      end
```

and the failures are:

```
  1) Failure:
Nokogiri::XML::TestReader#test_pushing_to_non_array_raises_TypeError [test/xml/test_reader.rb:203]:
[TypeError] exception expected, not
Class: <Nokogiri::XML::SyntaxError>
Message: <"Nokogiri::XML::SyntaxError">
---Backtrace---
nokogiri/XmlReader.java:355:in `read'
test/xml/test_reader.rb:204:in `block in test_pushing_to_non_array_raises_TypeError'
---------------


  2) Failure:
Nokogiri::XML::TestReader#test_errors_is_an_array [test/xml/test_reader.rb:194]:
Expected: [Nokogiri::XML::SyntaxError]
  Actual: []
```

this has been a known problem since @knu spotted this in 0af5a8e.

PR #1590, opened 2017-01-26
[jruby] cleanup and performance improvements

kares:
most of the stuff is just static (Java) analysis and wondering around the code spotting inefficiencies

the other parts are from doing some quick (xpath) profiling which revealed that there's an unused method
[during `node.content` doing NKF](https://github.com/sparklemotion/nokogiri/commit/70f30e5af68f575766a986954cc4619be87eacaf) (shall post some numbers later as I might get into some more detailed profiling). there's also a [fix for #1586](https://github.com/sparklemotion/nokogiri/commit/076c3c68c28b877de503f4eb4e9e184139687513) done along the way.


flavorjones:
This is great stuff, thank you so much for submitting these changes!

PR #1604, opened 2017-02-12
this isit

kares:
@flavorjones #1319 and #1320 at your service!

these are really the 'simplest' thing that works ... 

- #1320 needs a Xalan "patch" (as noted on the issue). alternatively a Xalan fork could be maintained.
  if there's more of these hacks needed maybe switching to another xpath engine should be considered.
- #1319 similar issue might show up elsewhere, very ripe for refactoring the internals (maybe next time)

### DIFF
--- a/lib/nokogiri/xml/node_set.rb
+++ b/lib/nokogiri/xml/node_set.rb
@@ -293,7 +293,11 @@ module Nokogiri
       # Returns a new NodeSet containing all the children of all the nodes in
       # the NodeSet
       def children
-        inject(NodeSet.new(document)) { |set, node| set += node.children }
+        node_set = NodeSet.new(document)
+        each do |node|
+          node.children.each { |n| node_set.push(n) }
+        end
+        node_set
       end
 
       ###
--- a/lib/nokogiri/xml/node_set.rb
+++ b/lib/nokogiri/xml/node_set.rb
@@ -75,7 +75,7 @@ module Nokogiri
         rules, handler, ns, _ = extract_params(args)
 
         inject(NodeSet.new(document)) do |set, node|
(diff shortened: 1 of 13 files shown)


flavorjones:
Wow! Thanks so much for submitting this. Looking at everything now ...

lib/nokogiri/xml/node_set.rb
```diff
@@ -293,7 +293,11 @@ def == other
       # Returns a new NodeSet containing all the children of all the nodes in
       # the NodeSet
       def children
-        inject(NodeSet.new(document)) { |set, node| set += node.children }
```
flavorjones: I'm not sure I understand why this change is being made. Is it for performance? If so, can you share your benchmarks?

I'm asking only because it seems out of place in this patchset.

flavorjones:
Other than the one question I made above, this looks great.

flavorjones:
Tagging https://github.com/flavorjones/loofah/issues/88 as that is definitely addressed by this patchset.

lib/nokogiri/xml/node_set.rb
kares: shouldn't be necessary.
wanted to make sure only a single `NodeSet` is created (as I was debugging how the whole piece behaves with decorators), than I left it in as it seemed to make sense but I do not have benchmarks.

it "must" be faster :) but you're right this should have been 2 PRs I just :nail_care:  as I went with the fixes.

on my defense - its uglier but similar to the `def reverse` code bellow :wink: 
I am also fine with moving it out ... if you prefer to have it separately?

flavorjones: Nah, it's fine, I was just curious.

PR #1606, opened 2017-02-13
gemspec for -java version of the gem should specify >= 1.9.3

kares:
... JRuby 1.7 is still supported but not in --1.8 mode

PR #1607, opened 2017-02-14
get rid of JRuby --1.8 legacy code in ext

kares:
... a #1606 follow-up


kares:
Hey Mike, I am not sure the 1.8 wait is needed. Since **1.7.1** (thanks for the release) added the `>= 1.9.3`.
... or do you just prefer to have it that way? I'm asking cause I have some more JRuby pieces lying around.

kares:
oh, I see 1.7.1 was not done off master but only includes CVE fixes over 1.7.0 ... ignore that part than.

flavorjones:
Thanks so much! Merged.

PR #1608, opened 2017-02-17
Add missing constant LIB_DIRS to nokogiri/extconf.rb

codenamev:
I have a custom path set with the `CPUPROFILE` environment variable.  Because of this, I consistently get the following when trying to install the gem:

```
extconf.rb:661:in `<main>': uninitialized constant LIB_DIRS (NameError)
```

Grepping the entire source tree I confirmed that this constant is never declared.

I'm using ruby 2.3.3.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -4,6 +4,7 @@ ENV['RC_ARCHS'] = '' if RUBY_PLATFORM =~ /darwin/
 require 'mkmf'
 
 ROOT = File.expand_path(File.join(File.dirname(__FILE__), '..', '..'))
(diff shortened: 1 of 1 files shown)


flavorjones:
@codenamev Thanks for submitting this PR!

Can you help me understand how to reproduce what you were seeing? I need to understand the "why?" behind changes that go into extconf.rb.

flavorjones:
@codenamev Ah, never mind, I see it.

codenamev:
Sorry for not providing clearer context.  Glad you found it :smile:.  I was going to add tests, but didn't see any others that test the native extension stuff as a reference, and this is small potatoes.

flavorjones:
I'm instead going to just remove that block for now; we can add it back later, correctly, if it's valuable.

PR #1613, opened 2017-03-10
Add ruby-head as allow_failures in .travis.yml.

junaruga:
I want to add `ruby-head` as allow_failures in .travis.yml.
I think this is useful.
Because we can prepare before next version Ruby 2.5 release.
We can support the next version as faster

I found the code in `rails` and `rspec`.
https://github.com/rails/rails/blob/master/.travis.yml
https://github.com/rspec/rspec-core/blob/master/.travis.yml

Thanks

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,11 +30,22 @@ matrix:
     env:
       - RUBYOPT="--enable-frozen-string-literal --debug=frozen-string-literal"
   - os: linux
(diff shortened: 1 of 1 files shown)


flavorjones:
I don't think this is useful until there's an actual preview release; but I have no objections and so will merge.

flavorjones:
Thank you!

PR #1621, opened 2017-03-22
Add release date to v1.7.1

mobilutz:
The release date of `v1.7.1` was missing in the changelog.


flavorjones:
Thanks for the reminder. Unfortunately, the date in this PR is incorrect (at least for my local time (currently UTC-4 thanks to daylight savings time) which has been the convention in the past).

I've updated the CHANGELOG, though, so thank you!

mobilutz:
Thats always a problem with timestamps ;)

PR #1622, opened 2017-03-23
Added 'frozen_string_literal: true' comment to all ruby files

MOZGIII:
Warning! This breaks stuff (currently).
See #1623


larskanis:
Thank you for the pull request. However there is no need to add a comment to each and every file. See https://github.com/sparklemotion/nokogiri/issues/1623#issuecomment-288693291 .

PR #1637, opened 2017-05-04
In SyntaxError, not display line and column if are zero

### LINKED ISSUE #1611: SyntaxError.to_s should not display line and column if both (either?) are zero
related to 879d346 (which is on master but has not been released yet)

this code, running on master with the above commit:

```ruby
Nokogiri::XML('<ns1:Root></ns1:Root>').xpath('//ns1:Root)
```

emits:

```
Nokogiri::XML::XPath::SyntaxError: 0:0: ERROR: Undefined namespace prefix: //ns1:Root
```

Note that the `0:0` is line and column, and should not be emitted in this case.

ccarruitero:
Fixes #1611

### DIFF
--- a/lib/nokogiri/xml/syntax_error.rb
+++ b/lib/nokogiri/xml/syntax_error.rb
@@ -57,8 +57,12 @@ module Nokogiri
         end
       end
 
+      def nil_or_zero?(attribute)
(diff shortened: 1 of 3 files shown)


flavorjones:
Awesome, thank you for submitting! Looks good to me, merging.

PR #1676, opened 2017-09-28
Replaced character stack with string buffer

andrew-aladev:
I've found an inconsistency between `c` and `java` version of html and xml sax parsers.

Let it be an html:
```
<p>
  text 1
  <span>text 2</span>
  text 3
</p>
```

We can expect that our sax parser will receive the following information:
```
start_element "p"
characters "text 1"
start_element "span"
characters "text 2"
end_element "span"
characters "text 3"
end_element "p"
```

`c` version works perfect but `java` does not. I've found a `characterStack` in `ext/java/nokogiri/internals/NokogiriHandler.java` that is actually a bug. I couldn't understand the reason of using stack for characters.

I was thinking for a long time about this stack and I've found the only reason for it: we will receive a list of character strings anyway when xml or html syntax is broken at the end of the document. But nokogiri has absolutely another methods of handling broken syntax. I think nokogiri should not keep this stack in the code.

So I've removed characters stack and added simple string buffer. The fix itself is very small.

Than I've added a special document to the test helper in order to test the strict order of parsed items. I've implemented 2 tests with regular text and whitespace for html and xml sax parsers.

Text nodes are regular text and comments for html parser. Xml test received a bonus: cdata blocks.

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -298,6 +298,7 @@ test/helper.rb
 test/html/sax/test_parser.rb
 test/html/sax/test_parser_context.rb
 test/html/sax/test_push_parser.rb
+test/html/sax/test_parser_text.rb
 test/html/test_builder.rb
 test/html/test_document.rb
 test/html/test_document_encoding.rb
@@ -325,6 +326,7 @@ test/xml/node/test_subclass.rb
 test/xml/sax/test_parser.rb
 test/xml/sax/test_parser_context.rb
 test/xml/sax/test_push_parser.rb
+test/xml/sax/test_parser_text.rb
 test/xml/test_attr.rb
 test/xml/test_attribute_decl.rb
 test/xml/test_builder.rb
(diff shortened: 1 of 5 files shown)


andrew-aladev:
I will fix ruby-1.9.3 tomorrow.

flavorjones:
Hi @andrew-aladev, thanks for submitting this. I love the additional test coverage!

@knu and I have asked a couple of the core contributors who know the jruby implementation a bit better than us to take a look.

And yes, if you avoid the `%i` syntax all tests will pass on jruby 1.7. I've increased PR test coverage in Concourse in commit be56b1e to include jruby 1.7 in response.

flavorjones:
Also pinging @kares for an opinion.

kares:
looking good, just get CI :green_heart: and its perfect :) 
the stack was introduced due order (if I recall right) but tests verify its preserved, one minor thing to consider would be whether we need `StringBuffer`'s synchronization ... (or could use a `StringBuilder`)

andrew-aladev:
I've replaced `StringBuffer` with `StringBuilder`. JRuby 1.7 and MRI ruby 1.9.3 seems to work good.

flavorjones:
Merged! Thank you for your contribution!

PR #1685, opened 2017-11-01
Fix a couple of issues on OpenBSD

jeremyevans:
If the base compiler or compiler specified by CC is clang, then
don't require egcc.

Add /usr/local/include as an include directory when compiling.

With these changes, the current release of OpenBSD should be
able to install nokogiri using gem install without
--use-system-libraries.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -393,8 +393,11 @@ when arg_config('--clean')
 end
 
 if openbsd? && !using_system_libraries?
-  ENV['CC'] ||= find_executable('egcc') or
-    abort "Please install gcc 4.9+ from ports using `pkg_add -v gcc`"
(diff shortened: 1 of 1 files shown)


flavorjones:
Looks good, PR pipeline went green. Merging!

Are there any changes to the [installatoin tutorial](https://github.com/sparklemotion/nokogiri.org-tutorials) that we should make as a result?

jeremyevans:
For the install tutorial, where it mentions FreeBSD/OpenBSD, maybe just change `FreeBSD` to `FreeBSD/OpenBSD>=6.2` and `OpenBSD` to `OpenBSD<6.2`.

flavorjones:
Docs change done and pushed to http://www.nokogiri.org/tutorials/installing_nokogiri.html

PR #1696, opened 2017-12-11
remove ruby 1.9.3 and 2.0 from appveyor.yml

marutosi:
Ref: 84870381847cb

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -19,6 +19,3 @@ environment:
     - ruby_version: "22-x64"
     - ruby_version: "21"
     - ruby_version: "21-x64"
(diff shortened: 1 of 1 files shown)


flavorjones:
I'm strongly considering removing `appveyor.yml` as our [current test pipeline](https://ci.nokogiri.org/) is quite complex and includes windows builds for all current versions of windows.

Appveyor tends to run slowly, and so in comparison its feedback loops are quite long and I don't find any use in maintaining the file.

I'm curious what your thoughts are?

marutosi:
I don't have thought of Windows build environment.
But I close this PR because #1702 (cc80904) covered it.

Issue #1238, opened 2015-02-01
Segmentation fault inspecting some entities


flavorjones:
Originally reported on the `nokogiri-talk` mailing list:

> https://groups.google.com/d/msg/nokogiri-talk/Kl8E_RjBf8w/0zIw9IuiBGIJ

I've rewritten the original example slightly:

``` ruby
require "nokogiri"
require "pp"

doc1 = Nokogiri::HTML('<p></p>')
doc2 = Nokogiri::HTML('<p></p>')

ent = Nokogiri::XML::EntityReference.new(doc1, "amp")

doc2.css("p").each do |pp|
  pp << ent
end
pp doc2
```

running:

``` yaml
# Nokogiri (1.6.6.2)
    ---
    warnings: []
    nokogiri: 1.6.6.2
    ruby:
      version: 2.1.5
      platform: x86_64-linux
      description: ruby 2.1.5p273 (2014-11-13 revision 48405) [x86_64-linux]
      engine: ruby
    libxml:
      binding: extension
      source: packaged
      libxml2_path: "/home/miked-public/code/oss/nokogiri/ports/x86_64-unknown-linux-gnu/libxml2/2.9.2"
      libxslt_path: "/home/miked-public/code/oss/nokogiri/ports/x86_64-unknown-linux-gnu/libxslt/1.1.28"
      libxml2_patches:
      - 0001-Revert-Missing-initialization-for-the-catalog-module.patch
      - 0002-Fix-missing-entities-after-CVE-2014-3660-fix.patch
      libxslt_patches:
      - 0001-Adding-doc-update-related-to-1.1.28.patch
      - 0002-Fix-a-couple-of-places-where-f-printf-parameters-wer.patch
      - 0003-Initialize-pseudo-random-number-generator-with-curre.patch
      - 0004-EXSLT-function-str-replace-is-broken-as-is.patch
      - 0006-Fix-str-padding-to-work-with-UTF-8-strings.patch
      - 0007-Separate-function-for-predicate-matching-in-patterns.patch
      - 0008-Fix-direct-pattern-matching.patch
      - 0009-Fix-certain-patterns-with-predicates.patch
      - 0010-Fix-handling-of-UTF-8-strings-in-EXSLT-crypto-module.patch
      - 0013-Memory-leak-in-xsltCompileIdKeyPattern-error-path.patch
      - 0014-Fix-for-bug-436589.patch
      - 0015-Fix-mkdir-for-mingw.patch
[...]

PR #1713, opened 2018-01-28
issue #1238 segfault on predefined entities

flavorjones:
Addresses #1238


flavorjones:
Concourse passed.

Travis failed, and appveyor is still running. I'm less and less convinced travis and appveyor are adding any value.

Merging.

PR #1715, opened 2018-01-29
Remove mysterious clearCachedNode call in JRuby that breaks Blather

### LINKED ISSUE #1708: JRuby cache clearing issue since 1.6.2 affects Blather (Adhearsion)
The [Blather](https://github.com/adhearsion/blather) project, an important part of [Adhearsion](https://github.com/adhearsion/adhearsion), has been [severely affected](https://github.com/adhearsion/blather/issues/145) by a cache clearing issue in the JRuby implementation since 1.6.2, and as such has had to lock the Nokogiri version back to 1.6.1. This is obviously a security concern and we are missing out on other fixes made since.

Blather (indirectly) subclasses `Nokogiri::XML::Node` as `Blather::XMPPNode`. It's hard to explain exactly what happens as I'm not the author and it's not the simplest of libraries but somewhere along the line, `node.parent` is called and rather than returning the original `Blather::XMPPNode` instance that was created earlier, a new `Nokogiri::XML::Element` instance is returned instead. It may not matter that the instance is different but the change in class certainly does matter. It should go without saying that this does not happen under MRI.

I bisected the problem back to sparklemotion/nokogiri@e51bb1d, made by @mbklein. I obviously had to raise an eyebrow when I saw that the following comment had since been added above the first clearCachedNote statement by @jvshahid.

```java
// TODO: this feels kind of weird, why are we clearing the XmlNode
// cache here !!!
clearCachedNode(node);
```

Indeed, if I comment out this line, the problem goes away and all the Nokogiri tests still pass. Unhelpfully, if I comment out the other clearCachedNode line below, all the Nokogiri tests still pass, so perhaps there is insufficient testing here.

My limited understanding tells me that because the cache is being cleared, the original Ruby instance is lost and a new one is being created in `NokogiriHelpers#constructNode` with the hardcoded class of `Nokogiri::XML::Element`. Given that we only have the `org.w3c.dom.Node` instance to recreate from, Ruby-specific details such as which subclass was used originally, are lost.

It is not clear to me exactly when the cache is supposed to be cleared. If it can happen quite easily then perhaps the approach that Blather takes is no longer viable. However, given the interesting comment noted above, I suspect it is currently being cleared more than it should.

chewi:
It was previously described at "weird" and even the original author cannot remember why it is there. Its presence causes Blather's `Blather::XMPPNode` subclass instances to be recreated where they become generic `Nokogiri::XML::Element` instances instead.

Closes #1708.

### DIFF
--- a/ext/java/nokogiri/XmlNode.java
+++ b/ext/java/nokogiri/XmlNode.java
@@ -464,9 +464,6 @@ public class XmlNode extends RubyObject {
             return;
         }
 
-        // TODO: this feels kind of weird, why are we clearing the XmlNode
-        // cache here !!!
-        clearCachedNode(node);
         Element e = (Element) node;
 
         // disable error checking to prevent lines like the following


chewi:
Closing in favour of #1716.

PR #1717, opened 2018-01-30
Bunch of changes to make XmlSaxPushParser on JRuby compatible with the MRI implementation

### LINKED ISSUE #1710: NullPointerException in XmlSaxPushParser#native_write triggered by fix for #1261
The last remaining 15 [Blather](https://github.com/adhearsion/blather) test failures are caused @yokolet's fix for #1261 in sparklemotion/nokogiri@3b121ca6e651f513922f9857ba2b4af9d176afb9. I'm not sure exactly what triggers it but sometimes `parserTask.parser.getNokogiriHandler()` returns `null`, leading to a NullPointerException.

Reverting the change allows all the Blather tests to pass. So does skipping the `try` block if the handler is `null`. Strangely changing `SAXException` to just `Exception` in the `catch` clause also makes it work. I'm not sure what the correct fix would be, though presumably reverting the change isn't the way to go.

jvshahid:
This pr should fix #1710 and fix the differences in the push parser implementations on MRI and JRuby. Each change is split in its own commit along with the test that fails on JRuby.

### DIFF
--- a/ext/java/nokogiri/HtmlSaxPushParser.java
+++ b/ext/java/nokogiri/HtmlSaxPushParser.java
@@ -160,7 +160,7 @@ public class HtmlSaxPushParser extends RubyObject {
 
         if (!options.recover && parserTask.getErrorCount() > errorCount0) {
             terminateTask(context);
-            throw new RaiseException(parserTask.getLastError(), true);
+            throw parserTask.getLastError();
         }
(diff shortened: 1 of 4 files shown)


chewi:
Thanks, this looks promising. I'll try to test it tomorrow.

Issue #1711, opened 2018-01-23
suggestion: package only files that are needed at runtime in the gem


stevecrozz:
I noticed that the nokogiri gem distribution published on rubygems.org contains files that are not strictly necessary for nokogiri at runtime. I suggest we follow the lead of the rails family of gems and only package those files that are needed at runtime:
Example: https://github.com/rails/rails/blob/v5.1.4/activesupport/activesupport.gemspec#L18

My interest is in cutting down the size of the docker images that I maintain so that moving images across the internet becomes faster (especially helpful during deploys). I introduced this topic on the mailing list first  and @flavorjones recommended we open this topic for discussion here on github.

I'd love to hear any arguments for/against this idea.

flavorjones:
Thanks for opening this issue for discussion, @stevecrozz.

For context, I'll repeat what I said on the mailing list, which is that it's been occasionally useful to install the tests to determine if Nokogiri is _really_ working on a system, particularly if using an old version of libxml2/libxslt. (In fact, this is in part how I determine if an RC is working properly on, e.g. windows.)

That said, I don't feel strongly about it and would like to know if other people thing gems should be limited to distributing just the runtime bits.

larskanis:
I usually remove [the test -> compile dependency](https://github.com/sparklemotion/nokogiri/blob/master/Rakefile#L306) in the git working directory to verify that the current installed gem works as expected. So IMHO we can remove the files in question.

Maybe it makes sense then to remove the test -> compile dependency above in the Rakefile permanently and call `rake compile test` in cases when we used to run `rake test` before?

stevecrozz:
There's not a whole lot of activity here, but I'm happy to advocate for this idea a bit to help push it forward. There is some value in having all the source files on a system in order to run tests. And there's also value in trimming the size of the gem. I'm not sure how to weigh these values relatively. Maybe we could mitigate the potential damage by providing an alternate method for getting the test files when needed. Is it tenable to ask users to git clone the source code at a specific tag when they need to run tests? That's what I would do if I wanted to run tests.

Even if that doesn't work for some users, it still **could** be more valuable to trim the gem than to package files in the release that will be so rarely used. But I recognize that's an easy thing for me to say, not having to support nokogiri users :)

stevecrozz:
Btw, I am happy to prepare the PR if that helps sweeten the deal.

flavorjones:
[...]

PR #1719, opened 2018-02-06
Make builds minimal

stevecrozz:
This is in reference to #1711 where I suggest we package minimal builds that only include the essentials. This change saves me about 2MiB on disk which sounds like nothing, but on my alpine linux Docker images its actually significant. Since we started discussing already, its probably best to talk about the general idea of removing non-essential files from the build in #1711, but here's a concrete example of what I'm thinking.


tenderlove:
This seems alright to me.  Agree with @flavorjones that it's sometimes helpful to have the tests, but I also don't feel strongly about it (especially since we tag the release version in git anyway).

flavorjones:
OK - I agree in principle with minimizing the gem; however I want to make sure I have automated testing of the generated gem to ensure that we're not omitting necessary files. This is at the top of my TODO list.

junaruga:
> however I want to make sure I have automated testing of the generated gem to ensure that we're not omitting necessary files.

@flavorjones thank you for the working with this. How was the status?
I proposed it with another pull-request.

stevecheckoway:
Making gems minimal in size sounds good to me as long as not too much is removed. E.g., https://github.com/sparklemotion/nokogiri/pull/1788 wants the headers installed in the extension directory or at least not removed.

Until that (or something like it) happens, Nokogumbo is actually using the `dependency.yml` file. See https://github.com/rubys/nokogumbo/blob/master/ext/nokogumbo/extconf.rb#L15. I'd like to remove most of that code (i.e., downloading, configuring, and installing headers for libxml2), but unless Nokogiri vends the embedded headers, having the version, and SHA-256 hash of the library source code is very helpful.

stevecrozz:
I've rebased this branch onto most recent master and @stevecheckoway I have restored dependencies.yml to Manifest.txt as it sounds like gems may depend on it.

Manifest.txt
```diff
@@ -256,6 +238,7 @@ patches/libxml2/0003-Fix-infinite-loop-in-LZMA-decompression.patch
 patches/sort-patches-by-date
 suppressions/README.txt
 suppressions/nokogiri_ruby-2.supp
+<<<<<<< HEAD
```
junaruga: We can remove `git` conflict mark here.

Manifest.txt
```diff
@@ -368,3 +351,5 @@ test/xml/test_xinclude.rb
 test/xml/test_xpath.rb
 test/xslt/test_custom_functions.rb
 test/xslt/test_exception_handling.rb
+=======
+>>>>>>> Make builds minimal
```
junaruga: We can remove `git` conflict mark here.

stevecrozz:
Thanks @junaruga. Updated.

flavorjones:
@stevecrozz Can you please rebase on master? We fixed some things in 464717b related to JRuby 9.2 backwards incompatibilities that are making this PR fail tests on JRuby 9.2.

stevecrozz:
No prob @flavorjones. I force-pushed a newly rebased branch.

flavorjones:
Woo, that went green. OK, I may try to get this into v1.9.0. Thanks!

flavorjones:
I've merged this manually! Will be in v1.9.0. Thanks again for your patience and for submitting this PR!

PR #1700, opened 2017-12-11
v1.7.x: allow jruby failures on travis

marutosi:
Travis tries to install jruby-9.1.15.0200, then test failed.

PR #1720, opened 2018-02-20
CI against jruby-head

yahonda:
Rails 6 will require CRuby 2.4.1 or higher https://github.com/rails/rails/pull/32034. It means JRuby 9.2 will be necessary to support Rails 6.

Then I have opened a pull request https://github.com/rails/rails/pull/32064 to run CI with`jruby-head`and it gets `java.lang.ClassFormatError` below.

https://travis-ci.org/rails/rails/jobs/343639864

```ruby
LoadError: load error: nokogiri/nokogiri -- java.lang.ClassFormatError: Duplicate method name&signature in class file nokogiri/XmlSyntaxError$INVOKER$i$to_s19
``` 

I have opened a pull request to run nokogiri CI with jruby-head to see if this error reproduces without Rails.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -42,9 +42,12 @@ matrix:
   - os: linux
     rvm: jruby-9.1.5.0
   - os: linux
+    rvm: jruby-head
+  - os: linux
     rvm: rbx-3
   allow_failures:
   - rvm: ruby-head
+  - rvm: jruby-head
   fast_finish: true
 
 # notifications:


yahonda:
Found https://github.com/sparklemotion/nokogiri/pull/1700#issuecomment-366487786 . I am fine whichever CI tool as long as I can see if `LoadError: load error: nokogiri/nokogiri -- java.lang.ClassFormatError: Duplicate method name&signature in class file nokogiri/XmlSyntaxError$INVOKER$i$to_s19` error reproduces with developing version of JRuby which will be released as JRuby 9.2.

Thank you.

flavorjones:
Currently we use the Docker images published at https://hub.docker.com/_/jruby/ to test. There doesn't appear to be a 9.2 image, or a HEAD image, available. When there is, we'll test it.

yahonda:
Thanks for the update. I understand how Nogiri is tested. 
The `LoadError` it self has been addressed https://github.com/jruby/jruby/commit/065cf7813511083f785239b2329f15fc9c0b27b6 I do not need to push this pull request anymore now.

I hope JRuby head will be available from the docker image soon.

PR #1723, opened 2018-02-21
[fix] avoid potential (legacy) issues booting on newer JRubies 

kares:
this change makes Nokogiri's JRuby ext backwards compatible as well as future compatible

see #1720 where JRuby (head) 9.2 had to work around 1.8 compat mode


flavorjones:
Thank you!

PR #1735, opened 2018-03-14
Flavorjones fix compilation warnings

larskanis:
This is a proposal for leaving -Wconst-qual in place.


flavorjones:
The reason I wanted to remove `-Wconst-qual` is because `ruby.h` generates a ton of errors with it turned on. Are you not seeing the same thing I am?

flavorjones:
Ah, never mind! You've fixed it! :thumbsup:

PR #1737, opened 2018-03-20
Close unclosed tag

mgenereu:
Could I be wrong about this?


flavorjones:
Thank you for your contribution! Merging.

PR #1745, opened 2018-03-25
Add frozen_string_literal: true to all files

oniofchaos:
I see that the option is enabled in `.travis.yml` for loading ruby but this doesn't help when including the gem in a project.


flavorjones:
The magic comment was added to a few generated files, I'm curious how you did this (maybe with https://github.com/Invoca/magic_frozen_string_literal) and whether you considered how those generated files would stay updated?

oniofchaos:
Yeah, I used `magic_frozen_string_literal`. I see that `lib/nokogiri/css/tokenizer.rb` and `lib/nokogiri/css/parser.rb` have comments indicating they are auto generated. I did not notice this until you pointed it out. Are there any others I missed? Where can I go to edit the auto generator script to include the pragma?

flavorjones:
Hi, @oniofchaos. Sorry for the delay here. I've updated the test suite to have RuboCop add the magic comment to all lib files, to future-proof the concept in this PR. I've credited you in the changelog, but won't be merging this commit since it's been superseded by the RuboCop automation.

Thanks so much for your work here, I wouldn't have gotten around to this without your help.

PR #1748, opened 2018-03-29
push parser shouldn't throw a NPE when it receives no input

### LINKED ISSUE #1710: NullPointerException in XmlSaxPushParser#native_write triggered by fix for #1261

jvshahid:
fixes #1710

### DIFF
--- a/ext/java/nokogiri/internals/NokogiriHandler.java
+++ b/ext/java/nokogiri/internals/NokogiriHandler.java
@@ -87,6 +87,7 @@ public class NokogiriHandler extends DefaultHandler2 implements XmlDeclHandler {
         this.runtime = runtime;
         this.attrClass = (RubyClass) runtime.getClassFromPath("Nokogiri::XML::SAX::Parser::Attribute");
         this.object = object;
+        charactersBuilder = new StringBuilder();
         String objectName = object.getMetaClass().getName();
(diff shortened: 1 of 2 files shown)


chewi:
Tested against Blather, all good.

chewi:
Sorry for the ping but I would like to update beyond 1.6 this side of our PCI-DSS audit.

flavorjones:
I'm going to vote for merging this, but would like to file a bug report to get MRI implementation in line with this JRuby behavior (if possible / feasible). @jvshahid if you agree, please merge this and I can cut a release in the next few days.

jvshahid:
merged and opened https://github.com/sparklemotion/nokogiri/issues/1758 to address the inconsistencies between MRI and JRuby

Issue #1678, opened 2017-10-03
NodeSet.dup does not deep copy


darrenleeweber:
Take note of the `Element` object IDs in this console exploration of whether or not `NodeSet.dup` is actually returning new objects (as a deep copy implies).  The `Document.dup` is doing a deep copy, but `NodeSet.dup` is not.  (This console is using a class wrapping Nokogiri instance variables, but it should be obvious anyway; LMK if you need a "pure" Nokogiri" example.)

```ruby
> record_setB.print
<?xml version='1.0'?>
<records>
  <REC>
    <UID>WOS:A2</UID>
  </REC>
  <REC>
    <UID>WOS:B2</UID>
  </REC>
</records>

> record_setB.doc
=> #(Document:0x2b17e94437d0 {
  name = "document",
  children = [
    #(Element:0x2b17ea12b1c8 {
      name = "records",
      children = [
        #(Element:0x2b17ea12f804 { name = "REC", children = [
            #(Element:0x2b17ea1333dc { name = "UID", children = [ #(Text "WOS:A2")] })] }),
        #(Element:0x2b17ea13b438 { name = "REC", children = [
            #(Element:0x2b17ea13e188 { name = "UID", children = [ #(Text "WOS:B2")] })] })]
      })]
  })

> record_setB.doc.dup
=> #(Document:0x2b17ea3a2708 {
  name = "document",
  children = [
    #(Element:0x2b17ea3a843c {
      name = "records",
      children = [
        #(Element:0x2b17ea3acba4 { name = "REC", children = [ 
            #(Element:0x2b17ea3b0c18 { name = "UID", children = [ #(Text "WOS:A2")] })] }),
        #(Element:0x2b17ea3b8b0c { name = "REC", children = [
            #(Element:0x2b17ea3c18c4 { name = "UID", children = [ #(Text "WOS:B2")] })] })]
      })]
  })
```

So far, so good.  The `Document.dup` is returning all new object IDs thanks to a deep copy.  Now let's see if the `NodeSet.dup` is doing a deep copy:
```ruby
> rec_node_setB = record_setB.doc.search('REC')
=> [#<Nokogiri::XML::Element:0x2b17ea12f804 name="REC" 
  children=[#<Nokogiri::XML::Element:0x2b17ea1333dc name="UID" 
    children=[#<Nokogiri::XML::Text:0x2b17ea137324 "WOS:A2">]>]>,
[...]

PR #1755, opened 2018-05-02
Documentation update about behaviour of duplicate

opoudjis:
Addresses https://github.com/sparklemotion/nokogiri/issues/1678

### DIFF
--- a/ext/nokogiri/xml_node.c
+++ b/ext/nokogiri/xml_node.c
@@ -500,6 +500,9 @@ static VALUE internal_subset(VALUE self)
  *
  * Copy this node.  An optional depth may be passed in, but it defaults
  * to a deep copy.  0 is a shallow copy, 1 is a deep copy.
+ * NOTE: any changes to the duplicated node will also alter the original node, 
+ * since they are contained in the same document. If you wish to change a duplicate
+ * of a node, create a new document based on the node instead.
  */
 static VALUE duplicate_node(int argc, VALUE *argv, VALUE self)
 {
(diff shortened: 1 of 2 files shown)


flavorjones:
Thanks so much for submitting this documentation.

I don't think the docs added for `Node#dup` are accurate, though. Specifically this bit:

```
any changes to the duplicated node will also alter the original node, 
since they are contained in the same document
```

is not correct. See for example this code:

```ruby
#! /usr/bin/env ruby

require "nokogiri"

xml = <<EOXML
<root><thing number="1">one</thing></root>
EOXML

doc = Nokogiri::XML xml
node = doc.at_css("thing")
copy = node.dup

copy.to_xml # => "<thing number=\"1\">one</thing>"
copy.set_attribute("foo", "bar")
copy.to_xml # => "<thing number=\"1\" foo=\"bar\">one</thing>"

node.to_xml # => "<thing number=\"1\">one</thing>"
doc.to_xml # => "<?xml version=\"1.0\"?>\n<root>\n  <thing number=\"1\">one</thing>\n</root>\n"
```

I'm going to merge the `NodeSet#dup` documentation, though. I'm happy to talk about what behavior you saw that led you to writing the `Node#dup` docs

flavorjones:
Merged! See b11b0f9. Thank you for your contriibution!

opoudjis:
The issue arose with `node1.xpath("//link | //locality").each(&:remove)`, but I guess that is a NodeSet issue after all. Thank you.

PR #1757, opened 2018-05-10
Add methods for class name manipulation to Node

knu:
NodeSet has useful methods `#add_class` and `#remove_class`, but we don't have those in Node.

This PR adds the following methods for easy manipulation of the "class" attribute as a list of words:

- Node#classes
- Node#add_class(name)
- Node#append_class(name)
- Node#remove_class(name)
- NodeSet#append_class(name)

I carefully avoided deduplication because in HTML(5) a class attribute value is just a list of words and some may rely on the exact sequence of words in a class attribute value.


flavorjones:
Thank you, @knu. I appreciate the attention to detail in the tests. Merging!

PR #1760, opened 2018-05-15
fix DOM2DTM to work in J9

### LINKED ISSUE #1759: DOM2DTMExt nextNode fails when running in J9
Running this simple script in IBM J9 on AIX:

```ruby
require 'nokogiri'
Nokogiri::XML.fragment('<root/>')
```

gives the following error:

```
    java.lang.IllegalAccessError: Class org/apache/xml/dtm/ref/dom2dtm/DOM2DTMExt illegally accessing "package private" member of class org/apache/xml/dtm/ref/dom2dtm/DOM2DTMdefaultNamespaceDeclarationNode
                   nextNode at org/apache/xml/dtm/ref/dom2dtm/DOM2DTMExt.java:564
                   _firstch at org/apache/xml/dtm/ref/DTMDefaultBase:-1
```

It seems like J9 is having issues actually seeing these classes as being in the same package.  Instead of this sort of package hack, I think it might make more sense to actually inherit DOM2DTM from DOM2DTMExt and only override and replace methods as necessary than to replicate the class and patch methods.  I'm not sure if this is a possibility, but I'll be doing some experimentation and making a pull request if I crack it.

# Version information

```sh
$ ./jruby/bin/nokogiri -v
# Nokogiri (1.8.2)
    ---
    warnings: []
    nokogiri: 1.8.2
    ruby:
      version: 2.3.3
      platform: java
      description: jruby 9.1.17.0-SNAPSHOT (2.3.3) 2018-05-11 d8b1ff9 IBM J9 VM 2.6 on
        pap6470sr10fp10-20170726_05 (SR10 FP10) +jit [aix-ppc64]
      engine: jruby
      jruby: 9.1.17.0-SNAPSHOT
    xerces: XML4J 4.5.28
    nekohtml: NekoHTML 1.9.21
```

Taywee:
J9 wasn't coping with hot-patching standard library packages with a user class.  This was done to work around a package private constructor. This is fixed by moving into a local package and copying the class with the package private constructor into a local copy.

Note that DOM2DTMdefaultNamespaceDeclarationNode in copied almost entirely ver-batim out of the OpenJDK source tree.  The only changes are the package and the import of the DTMException, which now uses the external version instead of the internal one.

closes #1759

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -81,6 +81,8 @@ ext/java/nokogiri/internals/XalanDTMManagerPatch.java
 ext/java/nokogiri/internals/XmlDeclHandler.java
 ext/java/nokogiri/internals/XmlDomParserContext.java
 ext/java/nokogiri/internals/XmlSaxParser.java
+ext/java/nokogiri/internals/dom2dtm/DOM2DTM.java
+ext/java/nokogiri/internals/dom2dtm/DOM2DTMdefaultNamespaceDeclarationNode.java
 ext/java/nokogiri/internals/c14n/AttrCompare.java
 ext/java/nokogiri/internals/c14n/C14nHelper.java
 ext/java/nokogiri/internals/c14n/CanonicalFilter.java
@@ -108,7 +110,6 @@ ext/java/nokogiri/internals/c14n/NameSpaceSymbTable.java
 ext/java/nokogiri/internals/c14n/NodeFilter.java
 ext/java/nokogiri/internals/c14n/UtfHelpper.java
 ext/java/nokogiri/internals/c14n/XMLUtils.java
-ext/java/org/apache/xml/dtm/ref/dom2dtm/DOM2DTMExt.java
 ext/nokogiri/depend
 ext/nokogiri/extconf.rb
 ext/nokogiri/html_document.c
--- a/ext/java/nokogiri/internals/XalanDTMManagerPatch.java
+++ b/ext/java/nokogiri/internals/XalanDTMManagerPatch.java
@@ -20,8 +20,9 @@
 package nokogiri.internals;
 
 import org.apache.xml.dtm.DTM;
+import nokogiri.internals.dom2dtm.DOM2DTM;
+import nokogiri.internals.dom2dtm.DOM2DTMdefaultNamespaceDeclarationNode;
 import org.apache.xml.dtm.DTMWSFilter;
-import org.apache.xml.dtm.ref.dom2dtm.DOM2DTMExt;
 import org.apache.xml.res.XMLErrorResources;
 import org.apache.xml.res.XMLMessages;
 import org.w3c.dom.Node;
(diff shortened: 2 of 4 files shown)


flavorjones:
Thank you for submitting this PR!

@jvshahid and @kares I'd like your approval before merging. I'm also curious if we're OK with using OpenJDK code from a license perspective?

kares:
> I'd like your approval before merging.

would prefer keeping the class in `org.apache` but I understand its causing issues with 9 modules.
git rename was handled properly so an eventual look into history to understand why will be there...

> I'm also curious if we're OK with using OpenJDK code from a license perspective?

not sure, with JRuby we generally avoid copy-ing code from OracleJDK which is essentially OpenJDK although licensing terms does change with different release.

jvshahid:
@Taywee why did you copy `DOM2DTMdefaultNamespaceDeclarationNode.java` from OpenJDK instead of using https://github.com/apache/xalan-j/blob/trunk/src/org/apache/xml/dtm/ref/dom2dtm/DOM2DTMdefaultNamespaceDeclarationNode.java. I think Xalan-J's is the original and has an Apache license.

Taywee:
@jvshahid Ignorance, probably.  I saw DOM2DTMExt being almost exactly identical to the version in the OpenJDK tree and assumed that's where it had come from.  My knowledge of typical licensing issues in the Java world is relatively limited (I don't work in Java much), mostly just used this as a patch to get it running in J9 on AIX.  If these are issues, I'd be happy to rectify them and roll those changes into this PR.

kares:
> I'm not sure how we ended down this path in the first place. I saw your comment on #1604 that Xalan isn't maintained anymore, but it looks to me like XalanJ is still maintained

yeah sorry about that, I personally have a lot on my OSS plate to spend more "free" time on this, atm.
mostly did it as @flavorjones pointed out that fixing those 2 long-term issues would be very beneficial.
so it was really a '*patch* till it works' job. status was seemingly dead back then - no new release in years.

kares:
> In the thread I linked to above, there is a mention of another library Saxon which has more advanced features and is better maintained. I think we should try it out.

smt of a more recent XSLT impl to try would be awesome esp. as things are getting 'hacky' with xalan

flavorjones:
OK, I'm going to rebase this branch and, assuming it goes green, will merge it.

@jvshahid do you want me to create a new Issue to track exploration of using Saxon?

jvshahid:
Yes please

On Dec 3, 2018 17:15, "Mike Dalessio" <notifications@github.com> wrote:

OK, I'm going to rebase this branch and, assuming it goes green, will merge
it.

@jvshahid <https://github.com/jvshahid> do you want me to create a new
Issue to track exploration of using Saxon?

—
You are receiving this because you were mentioned.

Reply to this email directly, view it on GitHub
<https://github.com/sparklemotion/nokogiri/pull/1760#issuecomment-443890989>,
or mute the thread

PR #1765, opened 2018-05-27
'#each' returns enumerator when no block given

park53kr:
## Overview

The previous code produced an unconditional error unless the blocks were passed to each.

```ruby
headers.each.with_index(1) # returns 'no block given' error
```
So I can't call `Enumerator` methods like with_index

## Codes

- return Enumerator if the block is not given.
- add test code


flavorjones:
Merged! Thank you for your contribution!

PR #1770, opened 2018-06-19
Add SemVer compatibility and gem version badges to README

greysteil:
First up, thanks for Nokogiri. I have to reach for an XML parser more often than I'd like, and am always extremely grateful for it. 

Would you be up for adding a badge that shows how SemVer compliant / bug free new releases are? I was looking through the data we gather at Dependabot and realised we could put one together, so I threw together the below.

[SemVer compatibility](https://dependabot.com/compatibility-score.html?dependency-name=nokogiri&package-manager=bundler&version-scheme=semver)

If you click through then there's a description of how it's calculated - basically it takes all of the relevant updates Dependabot has done for projects that use nokogiri and checks what percentage of the time specs pass on the upgrade PR.

The score is slightly lower than 100% because some projects have flaky specs, but I'm working on filtering them out from the data, too :octocat:.

Finally, I've removed the versioneye badge because it now just hangs (they went out of business last year) and added a gem version badge (because I always want to know it before adding a dependency to my Gemfile).


flavorjones:
Thanks for this PR! The failing tests obviously have nothing to do with this patch and I'll look into it tomorrow.

flavorjones:
Hey, so I really, really appreciate the thoughtfulness of this PR and the time you took to make a contribution.

I'm conflicted, however, since Nokogiri has never really followed semver for a variety of reasons, not the least of which is the continually-varying dependencies on libxml2 and xerces/nekohtml; also consider that everyone seems to have a different opinion on what a breaking change is.

I've had dreams of moving to semver after a 2.0 release, but it's been a few years and it's still not well-defined in my mind, and so I'm hesitant to suggest to anyone that we might be following semver when we're not ... see https://twitter.com/veganstraightedge/status/1008899145904545792 for the typical complaint about this.

What are your thoughts about this? I'm happy to note that we rarely have backwards-incompatibilities, but don't want to promise (or even pursue) a semver policy. Dependabot is pretty interesting, that's great data to have.

greysteil:
Ah, interesting. I can see how those libxml2 and xerces/nekohtml dependencies make things tricky in theory, although in practice it looks like they very rarely cause (compatibility) issues.

I think it would be really useful to have a paragraph in the readme mentioning that. It sounds like for almost all users, Nokogiri is backwards compatible between minor and patch releases, and that's all that most people need to know. Maybe that could be a footnote to this badge, or I could change the badge to just say `stability`?

(I've re-written this comment several times as I think it over - sorry if it's ended up incoherent!)

flavorjones:
Sorry for the delay. I thought about it, and hey, why not. Merging.

Issue #1785, opened 2018-08-15
Investigate Ubuntu libxml2 patches in USN-3739-1 and USN-3739-2


flavorjones:
This issue is to drive investigation and potential action around a set of upstream patches that Canonical judged valuable enough to port to their distributions.

References:

* https://usn.ubuntu.com/3739-1/
* https://usn.ubuntu.com/3739-2/

-----

Summary:

Two upstream patches, not yet available in an official libxml2 release, are candidates for patching in Nokogiri's vendored libxml2. A pull request has been created at https://github.com/sparklemotion/nokogiri/pull/1786 for comments.

flavorjones:
## USNs

https://usn.ubuntu.com/3739-1/ which addresses:

* CVE-2016-9318
* CVE-2017-16932
* CVE-2017-18258
* CVE-2018-14404
* CVE-2018-14567

and https://usn.ubuntu.com/3739-2/ which addresses:

* CVE-2016-9318
* CVE-2018-14404

Note that 3739-2 addresses a subset of 3739-1.

## CVEs

### CVE-2016-9318

Permalink is https://people.canonical.com/~ubuntu-security/cve/2016/CVE-2016-9318.html

Description:

> libxml2 2.9.4 and earlier, as used in XMLSec 1.2.23 and earlier and other products, does not offer a flag directly indicating that the current document may be read but other files may not be opened, which makes it easier for remote attackers to conduct XML External Entity (XXE) attacks via a crafted document.

Canonical rates this vulnerability as "Priority: Low"

The CVE report indicates this is the patch that addresses the vulnerability:

> https://git.gnome.org/browse/libxml2/commit/?id=ad88b54f1a28a8565964a370b5d387927b633c0d

Looking at libxml upstream:

```
$ git tag --contains ad88b54f1a28a8565964a370b5d387927b633c0d
v2.9.8
v2.9.8-rc1
```

... we see this was fixed in libxml 2.9.8, which Nokogiri has vendored since v1.8.3.


### CVE-2017-16932

Permalink is https://people.canonical.com/~ubuntu-security/cve/2017/CVE-2017-16932.html

Description:

> parser.c in libxml2 before 2.9.5 does not prevent infinite recursion in parameter entities

Canonical rates this vulnerability as "Priority: Low"

The CVE report indicates that this is the patch that addresses the vulnerability:

> https://github.com/GNOME/libxml2/commit/899a5d9f0ed13b8e32449a08a361e0de127dd961

[...]

PR #1786, opened 2018-08-15
pull in upstream libxml2 patches

flavorjones:
based on USN-3739-1 and -2.

see related #1785.


flavorjones:
If there are no objections in the next 48 hours, I'd like to ship this change as part of v1.8.5. Please do comment if you have opinions!

PR #1788, opened 2018-08-18
Nokogumbo should be able to compile against Nokogiri by finding headers in extensions directory

stevecheckoway:
This is an alternative implementation to https://github.com/sparklemotion/nokogiri/pull/1325. The major differences are
1. Nokogiri headers are placed in `nokogiri/include` inside the nokogiri extension directory rather than directly in `nokogiri`.
2. `libxml2` and `libxslt` headers are placed in `nokogiri/include/libxml2` and `nokogiri/include/libxslt` (and `nokogiri/include/libexslt`) if nokogiri used the packaged rather than system libraries.
3. The presence or absence of the headers is tested by travis-ci.

This will assist downstream projects (like [nokogumbo](https://github.com/rubys/nokogumbo/issues/71)) which need to build against a matching version of these libraries.

cc: @flavorjones

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,32 @@ matrix:
     rvm: jruby-9.1.15.0
   - os: linux
     rvm: rbx-3
+  - name: packaged libraries
+    os: linux
+    rvm: 2.4.2
+    env:
+      - OVERRIDE_GEM_PATH=/tmp/gem_path
+    install:
+      - bundle install --jobs=3 --retry=3 --path vendor/bundle
+      - bundle exec rake
+      - bundle exec rake gem
+      - gem install -i /tmp/gem_path pkg/nokogiri-*.gem
+      - gem install -i /tmp/gem_path minitest
+    script:
+      - bundle exec rake test:package
+  - name: system libraries
+    os: linux
+    rvm: 2.4.2
+    env:
+      - OVERRIDE_GEM_PATH=/tmp/gem_path
+    install:
+      - bundle install --jobs=3 --retry=3 --path vendor/bundle
+      - bundle exec rake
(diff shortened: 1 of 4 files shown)


stevecheckoway:
I just noticed that `Nokogiri::VERSION_INFO["libxml"]["libxml2_path"]` is the path to `libxml2` but that the build process has deleted that directory. Simply not deleting that directory would be sufficient for me. The test could be easily updated to test for the presence/absence of that directory.

flavorjones:
@stevecheckoway Apologies for my slow reply and lack of attention.

This PR looks good, I need to resolve the merge conflicts but will adopt your testing approach. I've got it from here!

I would like to block this PR on https://github.com/sparklemotion/nokogiri/issues/1845, though, which is how I'd like to test gem installation. And then this is targetted for release with v1.11.0 when it ships (see the milestone).

flavorjones:
Update: Things to do to make nokogumbo work well with Nokogiri v1.11.x+:
* merge this PR, keeping headers around
* of particular note: these headers need to be kept around in precompiled native linux gem releases
* we need to restore the `libxml_path` value in some form for VERSION_INFO
  * and if that key path in VERSION_INFO changes, send a PR to nokogumbo to use it.

Of course, we will add a CI task to test installing nokogumbo against the gem release under test to drive the above out.

flavorjones:
@stevecheckoway I've made commits in #2143 to fix this, and I've tested against nokogumbo and it seems like everything's working to me. Likely will cut a v1.11.0.rc4 today or tomorrow for some more tire kicking ahead of a final release by the end of the year.

stevecheckoway:
Great!

PR #1789, opened 2018-08-21
Add support for configuration of XSLT security

maths22:


flavorjones:
@maths22 Thank you for submitting this! And apologies for my slow reply. Will look at it and optimistically would like to ship in v1.11.0 (see milestone).

flavorjones:
I just rebased this off current `main` and added an assertion about `NotImplementedError` to the tests for JRuby.

flavorjones:
Again, apologies for the long delay in reviewing this and providing feedback.

There are a few things about this PR that I'd like to change:

- naming. Currently Nokogiri uses the word "options" (e.g., ParseOptions, SaveOptions) and this PR introduces two new words for the same concept: "prefs" and "config". I see that the [libxslt docs](http://xmlsoft.org/xslt/html/libxslt-security.html) use "prefs" but in the past the places where we've closely mirrored the libxml2 API are generally the parts of the API I regret most. I think we can improve on libxslt's API.
- initial values. It's not clear from this code or the tests what the initial values of these options is, and the setter isn't called by default.
- tests. I'd really like to have some tests exercising each of these libxslt features so we know whether we accidentally break them in a future release. this again is a hard-won lesson from not doing so when implementing ParseOptions or SaveOptions.

If you're not interested in working on this PR, given the long time delay, I understand. But please let me know if you think you'll be able to work on it? Otherwise I'll keep it open and hopefully find time to address some of this feedback.

maths22:
Sorry for the big delay here.  I've just pushed up a new version that I think exposes a more idiomatic ruby API, has more logical defaults, and checks that the underlying library gets all the config params as expected.

(I'm starting to work on getting nokogiri-xmlsec-instructure passing the build inside nokogiri proper, and currently this xmlsec stuff is shoved into that library where it definitely doesn't belong)

PR #1792, opened 2018-09-05
Replace DOM2DTM with DOM2DTMExt the patched version of DOM2DTM (fixes 1749)

### LINKED ISSUE #1749: JRuby XPATH Memory Usage
When using Nokogiri on jruby with a nested XPATH loop, the document memory footpring explodes in size. 

For example, with a document like
`<items>
  <item>
    <value1>
    <value2>
  <item>`
  ... for many thousands of items

And attempting to use (for example) an xpath like: doc.xpath('items').each { |node| node.xpath('value1') }

You'll wind up with a document that could be hundreds of megabytes large due to caching in the CACHED_XPATH_CONTEXT layer. Specifically the nokogiri.internals.XalanDTMManagerPatch winds up with thousands of values in "m_dtm". I'm not an expert in this area and am unclear what that terminology is referencing. In my case I had a document with 4000 items taking 4GB of memory of cached xpath. And there appears to be no way to clear that specific cache. 

This behavior is not present in the mri ruby version.

jvshahid:
Looks like we replaced DOM2DTM with DOM2DTMExt when we fixed
https://github.com/sparklemotion/nokogiri/pull/1320 but forgot to replace it in
the DOM2DTM manager

fixes #1749

### DIFF
--- a/ext/java/nokogiri/internals/XalanDTMManagerPatch.java
+++ b/ext/java/nokogiri/internals/XalanDTMManagerPatch.java
@@ -19,14 +19,14 @@
  */
 package nokogiri.internals;
 
+import javax.xml.transform.dom.DOMSource;
+
 import org.apache.xml.dtm.DTM;
 import org.apache.xml.dtm.DTMWSFilter;
 import org.apache.xml.dtm.ref.dom2dtm.DOM2DTMExt;
 import org.apache.xml.res.XMLErrorResources;
 import org.apache.xml.res.XMLMessages;
 import org.w3c.dom.Node;
-
-import javax.xml.transform.dom.DOMSource;
 
 /**
  * @author kares
@@ -75,8 +75,8 @@ public final class XalanDTMManagerPatch extends org.apache.xml.dtm.ref.DTMManage
         //   Have each DTM cache last DOM node search?
         for(int i = 0; i < m_dtms.length; i++) {
             DTM thisDTM = m_dtms[i];
-            if (thisDTM instanceof org.apache.xml.dtm.ref.dom2dtm.DOM2DTM) {
-                int handle = ((org.apache.xml.dtm.ref.dom2dtm.DOM2DTM) thisDTM).getHandleOfNode(node);
+            if (thisDTM instanceof DOM2DTMExt) {
+                int handle = ((DOM2DTMExt) thisDTM).getHandleOfNode(node);
                 if (handle != DTM.NULL) {
                     return handle;
                 }


jeremyhaile:
@jvshahid thank you so much for this fix. I can confirm it fixes the memory issues with nested XPATH searching. However, there is still a pretty large and possibly exponential performance penalty for nested XPATH searches.

Here is the test script I am using: 

```
require 'nokogiri'

items = %Q(
<div class="test">
  <a href="something">derp</a>
</div>
) * ARGV[0].to_i

html = %Q(
<html>
  <body>
    #{items}
  </body>
</html>
)
doc = Nokogiri::HTML html
puts "pid is #{$$}"

start = Time.now
links = doc.css('html body .test a')
puts "Without nesting got #{links.length} links in #{Time.now - start}s" # This will return immediately

start = Time.now
links = doc.css('html body .test').css('a')
puts "With nesting got #{links.length} links in #{Time.now - start}s" # This will take a LONG time to finish
```

Here is the output with 5000 divs:
```
pid is 49017
Without nesting got 5000 links in 0.225391s
With nesting got 5000 links in 2.666568s
```

And here is the output with 50,000 divs:
```
pid is 49113
Without nesting got 50000 links in 0.7900900000000001s
With nesting got 50000 links in 296.062414s
```

So in both cases the nested version is several orders of magnitude slower than the unnested version. And it seems to get exponentially worse for larger numbers of matched elements.

This should definitely be merged in to fix the memory issue, as crashing the JVM is much worse than being slow. However there is still something wrong with nokogiri java's handling of nested XPATH queries.

Any idea why the performance might be so bad? Thanks again!

jvshahid:
I think this is a result of the `NodeSet` implementation in JRuby which uses RubyArray under the hood. This causes the exponential behavior you are seeing as it tries to calculate the union in `.css('a')` for each node in the result of `.css('html body .test')`. As a workaround you can try using some variation of the following instead:

```
links = doc.css('html body .test').map do |node|
  node.css('a')
end
puts "finished"
links = Nokogiri::XML::NodeSet.new doc, links.flatten
```

I will probably open a new issue to track the exponential behavior of `NodeSet` and merge this PR to close #1749

jeremyhaile:
@jvshahid thanks for the explanation and workaround. Unfortunately for us the badly performant code is in a third-party library that uses nokogiri, although we could fork it and modify. 

Please let me know when you open the new issue and I'll subscribe there too. Thanks again!

jvshahid:
Just a quick update. I looked at libxml source code and it looks like internally it is using an array similar to the current implementation in JRuby. I am still not sure why the JRuby implementation is that slow.

jvshahid:
created https://github.com/sparklemotion/nokogiri/issues/1795 to track the performance of XMLNodeSet. This PR is ready to merge AFAIK.

WoolenWang:
why this old issue has merge and have the same issue #1873

PR #1804, opened 2018-10-16
Ignore test/** into gem for reduce gem size.

huacnlee:
Remove `test/*` for reduce gem size 2.7Mb.


flavorjones:
Although this was closed by the OP, I'd like to note that discussion around doing this has been ongoing at https://github.com/sparklemotion/nokogiri/issues/1711 and https://github.com/sparklemotion/nokogiri/pull/1719

PR #1820, opened 2018-11-17
Fix misuse of MAKEFILE_CONFIG

nobu:
As the values of `RbConfig::MAKEFILE_CONFIG` are intended to use in Makefile and may contain make macros, so a path in it may not match `/gcc/`, nor cannot execute directly.
Use `RbConfig::CONFIG`, whose values are expanded, for such porpose instead.


flavorjones:
Thank you! All tests are green, so I'm merging.

Issue #1822, opened 2018-11-30
NodeSet#each returns zero instead of self


olehif:
According to ruby convention, #each method should return self:
`[1, 2].each {} # => [1, 2]`

While NodeSet#each always return zero instead:
`nodes.each {} # => 0`

As everyone in the ruby community expects #each to always return self (I myself had a hard time debugging seemingly perfect line of code) I think it would make sense to change this behavior even though it might potentially be a breaking change.

flavorjones:
Hey! Thanks for submitting this. I agree it's unexpected behavior to not act like an `Enumerable` or `Array` class, which is what `NodeSet` is intended to behave like.

That said, I'll first (and this is intended to be humorous!) point out that this is not an obvious assumption for "everyone in the ruby community": see https://github.com/sparklemotion/nokogiri/issues/1678 where some people have opined that at least one aspect of `Enumerable`'s behavior is non-intuitive.

Second, I do want to point out that this is slightly more complicated than you've presented. The [`Array#each`](https://ruby-doc.org/core-2.2.0/Array.html#method-i-each) method returns an [`Enumerator`](https://ruby-doc.org/core-2.2.0/Enumerator.html) object when there's no block presented. This PR doesn't address that behavior, **and that's OK**, I just want to point out that there's a complex set of behaviors that despite our best intentions may never get fully implemented in NodeSet.

Further evidence of this complexity is in https://github.com/sparklemotion/nokogiri/issues/1677 where I honestly failed to completely understand the issue(s) being submitted.

In any case, this is an awesome, quick fix for what I consider to be a bug. Thank you! Will merge and this and it should be in the next release! Will give you a mention in the CHANGELOG.

PR #1823, opened 2018-11-30
NodeSet#each to return self instead of zero

olehif:
PR for https://github.com/sparklemotion/nokogiri/issues/1822


flavorjones:
I commented on this at #1822. Will merge and this should be in the next release. Thank you!

PR #1827, opened 2018-12-02
setting attribute value to blank in HTML doc now renders as blank

### LINKED ISSUE #1800: attribute.value = attribute.value changes output
**What problems are you experiencing?**
doing attribute.value = attribute.value makes it "boolean"/"minimized"

**What's the output from `nokogiri -v`?**
# Nokogiri (1.8.4)
    ---
    warnings: []
    nokogiri: 1.8.4
    ruby:
      version: 2.4.4
      platform: x86_64-linux
      description: ruby 2.4.4p296 (2018-03-28 revision 63013) [x86_64-linux]
      engine: ruby
    libxml:
      binding: extension
      source: packaged
      libxml2_path: "/usr/local/lib64/ruby/gems/2.4.0/gems/nokogiri-1.8.4/ports/x86_64-pc-linux-gnu/libxml2/2.9.8"
      libxslt_path: "/usr/local/lib64/ruby/gems/2.4.0/gems/nokogiri-1.8.4/ports/x86_64-pc-linux-gnu/libxslt/1.1.32"
      libxml2_patches:
      - 0001-Revert-Do-not-URI-escape-in-server-side-includes.patch
      libxslt_patches: []
      compiled: 2.9.8
      loaded: 2.9.8

**Can you provide a self-contained script that reproduces what you're seeing?**
```
doc = Nokogiri::HTML('<!doctype html><html><body><img alt=""></body></html>')
doc.xpath('//*/@*').each { |a| a.value = a.value } # this is the code that changes something while it shouldn't?
doc.to_html
```

flavorjones:
previously was rendered as a boolean attribute. also allow for boolean
attributes by setting value to `nil`.

Fixes #1800.

### DIFF
--- a/ext/java/nokogiri/XmlAttr.java
+++ b/ext/java/nokogiri/XmlAttr.java
@@ -130,7 +130,9 @@ public class XmlAttr extends XmlNode {
     @JRubyMethod(name = {"value=", "content="})
     public IRubyObject value_set(ThreadContext context, IRubyObject content){
         Attr attr = (Attr) node;
-        attr.setValue(rubyStringToString(XmlNode.encode_special_chars(context, content)));
+        if (content != null && !content.isNil()) {
+            attr.setValue(rubyStringToString(XmlNode.encode_special_chars(context, content)));
+        }
         setContent(content);
         return content;
     }
--- a/ext/nokogiri/xml_attr.c
+++ b/ext/nokogiri/xml_attr.c
@@ -4,37 +4,39 @@
  * call-seq:
  *  value=(content)
  *
- * Set the value for this Attr to +content+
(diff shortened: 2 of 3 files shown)


flavorjones:
Just a note that the PR enables the following behavior:

```ruby
#! /usr/bin/env ruby

require 'nokogiri'

doc = Nokogiri::HTML('<!doctype html><html><body><div bar="x" foo="x"></div></body></html>')
puts doc.to_html
# => <html><body><div bar="x" foo="x"></div></body></html>

doc.at_css("div").tap do |node|
  node.attributes['bar'].value = nil
  node.attributes['foo'].value = ""
end

puts doc.to_html
# => <html><body><div bar foo=""></div></body></html>
```

PR #1830, opened 2018-12-04
fix dom2dtm to work in j9

flavorjones:
This is a rebase of PR #1760 along with an additional commit to resolve a logical merge conflict.

@jvshahid would love your approval that my additional commit looks OK.


flavorjones:
@jvshahid I ran the repro script from #1749 and it ran OK so I'm going to assume I did it right. :wink:

Merging!

PR #1852, opened 2019-01-05
CSS selector `:has()` matches against deep descendants

### LINKED ISSUE #350: :has() selector should match against all descendants, not just direct children
From http://api.jquery.com/has-selector/:

> The expression `$('div:has(p)')` matches a `<div>` if a `<p>` exists anywhere among its descendants, not just as a direct child.

So I would expect:

```
Nokogiri::CSS.xpath_for("#outer:has(#inner)")  # => ["//*[@id = 'outer' and .//*[@id = 'inner']]"]
```

But, instead:

```
Nokogiri::CSS.xpath_for("#outer:has(#inner)")  # => ["//*[@id = 'outer' and *[@id = 'inner']]"]
```

which doesn't match:

```
<div id="outer">
  <div>
    <div id="inner"/>
  </div>
</div>
```

(I think that's the right way to match all descendants, but I'm not totally solid on XPath.  It works in my testing.)

flavorjones:
Fixes #350.

### DIFF
--- a/test/css/test_parser.rb
+++ b/test/css/test_parser.rb
@@ -63,8 +63,8 @@ module Nokogiri
       end
 
       def test_has
-        assert_xpath  "//a[b]", @parser.parse("a:has(b)")
-        assert_xpath  "//a[b/c]", @parser.parse("a:has(b > c)")
+        assert_xpath  "//a[.//b]", @parser.parse("a:has(b)")
+        assert_xpath  "//a[.//b/c]", @parser.parse("a:has(b > c)")
       end
 
       def test_dashmatch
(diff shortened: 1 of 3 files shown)


test/css/test_parser.rb
```diff
@@ -63,8 +63,8 @@ def test_to_a
       end
 
       def test_has
-        assert_xpath  "//a[b]", @parser.parse("a:has(b)")
-        assert_xpath  "//a[b/c]", @parser.parse("a:has(b > c)")
+        assert_xpath  "//a[.//b]", @parser.parse("a:has(b)")
+        assert_xpath  "//a[.//b/c]", @parser.parse("a:has(b > c)")
```
akostadinov, line 67: But it should match just `a`, that's the point of the `:has` css pseudo class. Otherwise you could just use a normal descendant selection, not `has`. Or am I missing something?

Issue #1855, opened 2019-01-08
please ship specs in the gem


boutil:
Hi

Until 1.8, nokogiri gem contained the spec files. It was useful for Debian (and certainly for other GNU/Linux distributions) to be able to run the test suite when building packages for their archive.

Since 1.9, the spec/ subdir is not included anymore. In Debian, we thought about using the github repository as the new source, but since it has no gemspec file (on purpose, as you explained already), byt has jar files, which we need to filter out, it makes things a bit tricky for packagers.

Therefore, we would like to kindly ask you if you could continue to ship the specs in the gem file.

Thank you in advance.

flavorjones:
Hi,

Thanks for bringing this up, and I apologize if this change caught you off-guard.

We did make reasonable efforts to discuss this publicly and get input into the decision. There was a public discussion on the mailing list and in a couple of different github issues about removing specs from the gem, and was met with general approval from users and downstream library maintainers. I released an RC with this change before shipping the final version. Apologies again if this was surprising, but then I'd encourage you to participate in the mailing list and/or github issues if you're repackaging Nokogiri.

Some history: this change was originally proposed on the [mailing list in January 2018](https://groups.google.com/forum/#!topic/nokogiri-talk/aGBNEhQIcMA). That discussion resulted in https://github.com/sparklemotion/nokogiri/issues/1711 and a PR at https://github.com/sparklemotion/nokogiri/pull/1719. I invite you to read all those threads for context.

Nokogiri has been downloaded over 202 million times, so there is a real cost to shipping files in the gem that nobody uses (those costs are borne by the users of the gem as well as rubygems.org). I think at this point I would need to hear very compelling reasons for putting files in the gem that aren't necessary for installation, runtime, or licensing/legal reasons.

I don't know much about what's done by Debian and other distros to repackage Nokogiri, but I imagine that it wouldn't be too troublesome to either download the [source tarball from the Github Releases page](https://github.com/sparklemotion/nokogiri/releases/tag/v1.10.0) or to use git to access the tagged release in the repo, and run those tests.

[...]

PR #1858, opened 2019-01-09
provide ability to package a gem that includes test files

flavorjones:
Related to #1855.


flavorjones:
As noted in #1855, this branch and PR are not necessary, and so I'm closing this PR. Will delete the branch in a week or so unless there are objections here or in #1855.

Issue #1845, opened 2018-12-17
concourse: automate building gems, testing of gem installation, and `gem push`


flavorjones:
The work under #1719, which modified the contents of the gemfiles, has made me realize that although we test the code pretty well, we don't have any automated test coverage on the gem installation process.

I'd like to set up the following:

__Release Candidate gems__

A manually-triggered pipeline with jobs to:
* build an RC gem for each of the platforms we ship, and upload them to an S3-compatible bucket somewhere
* test that installation of each RC succeeds on the appropriate platform(s), and run the test suite on the installed gem to make sure it's fully functional
* on success of the above, do `gem push` to rubygems.org

__Final release gems__

A manually-triggered pipeline with jobs to:
* build a final version of a gem, and upload to an S3-compatible bucket somewhere
* test that installation succeeds and the test suite passes on the installed gem
* on success of the above, do `gem push` to rubygems.org

__Security__

Keeping the software supply chain secure is obviously super-important. I'd like to take this opportunity to move towards using [CredHub](https://concourse-ci.org/creds.html#credhub) to then keep all credentials related to the S3 buckets and rubygems.org.

__Triggering__

Maybe we can set this up so it only runs when `Nokogiri::VERSION` changes? The easiest thing to do here might be to put that constant in a separate file, and trigger the builds only when that one file changes. If that works, then we don't need separate RC and Final pipelines; a single pipeline would do the trick.

flavorjones:
See https://github.com/sparklemotion/nokogiri/issues/1845 for the first build/install/test cycle that worked.

flavorjones:
JRuby is now also covered. This coverage is good enough that I can start refactoring `extconf.rb`.

Upcoming work:

- [ ] test gem installation on vanilla ubuntu
- [ ] test gem installation on alpine (see #1571)
- [ ] test gem installation on windows - precompiled
- [ ] test gem installation on windows - devkit
- [ ] do all of the gem tests in the PR pipeline

flavorjones:
Note that alpine test coverage was introduced as part of #1984.

flavorjones:
Note to self: I think we should remove this from the v1.11.0 milestone, but create two new issues which are:

- test windows native gem installation in CI
- test windows compile-from-source gem installation in CI

and tag those both for v1.11.0, because I'd like more confidence that I'm not shipping gems that are broken on Windows.

flavorjones:
Done - see #2151, #2152, #2153

PR #1868, opened 2019-01-26
Add an attempt to build upstream libxml2-nokogiri with patches

loz:
Note this is response to a twitter thread with @tenderlove around trying to build some disti packages to speed up install of the gem, if the disti build could already have a custom build.

1. This doesn't sort the consumption of a disti version in the gem build (yet)
2. This hasn't published, and only currently targets alpine, however there is a docker file which could be used to CI test that a package builds, and that nokogiri can install (when integrated)
3. Is hardcoded in the APKBUILD file which could probably be autogenerated from other artefacts in the repo
4. Needs to be integrated into a CI for publish, and in particular publish a PR to alpine for the build (I think) into a blessed alpine REPO.

This is a build/APK setup to build a custom alpine package for libxml

No nokogiri changes, just added some APK stuff

No

If not, that may be OK, just please note it here.

### DIFF
--- a/packages/alpine/APKBUILD
+++ b/packages/alpine/APKBUILD
@@ -1,0 +1,72 @@
+# Contributor: Jonathan Lozinski <jonathan.lozinski@gmail.com>
+# Maintainer: Jonathan Lozinski <jonathan.lozinski@gmail.com>
+pkgname=libxml2-nokogiri
+pkgver=2.9.9
+pkgrel=3
+pkgdesc="XML parsing library, version 2 (Nokogiri Dependent Version)"
+url="http://www.xmlsoft.org/"
+arch="all"
+license="MIT"
+depends=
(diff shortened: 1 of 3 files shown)


flavorjones:
@loz Thanks for submitting this. I have to admit I'm not totally sure what the goal of this PR is, can you maybe start from the beginning and assume I'm not familiar with the aforementioned twitter thread?

This sounds a little like you're trying to precompile libxml? And if that's the case we've already got a plan at https://github.com/sparklemotion/nokogiri/pull/1571 which is simply awaiting additional tests that install the gem and exercise the test suite (see also https://github.com/sparklemotion/nokogiri/issues/1845).

loz:
That is indeed the thread, around pre-installing/using pre-installed native binaries which are safe, and free from XSF etc..  Looks like this is covered, so closing out.

PR #1871, opened 2019-02-01
Remove support for future script macros

ksolo:
libxml2 added support for future script macros based on the below
specification

https://www.w3.org/TR/html401/appendix/notes.html#h-B.7.1

This syntax was reserved, but has yet to be defined. It is still unclear
how these macros will be used and what risks will be created if/when
implemented.

This PR removes support until more information is provided.

### DIFF
--- a/Manifest.txt
+++ b/Manifest.txt
@@ -234,3 +234,4 @@ lib/xercesImpl.jar
 lib/xml-apis.jar
 lib/xsd/xmlparser/nokogiri.rb
 patches/libxml2/0001-Revert-Do-not-URI-escape-in-server-side-includes.patch
+patches/libxml2/0002-Remove-script-macro-support.patch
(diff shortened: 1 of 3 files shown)


flavorjones:
Hey @ksolo! Thanks so much for submitting this, I agree we should remove this support in Nokogiri.

It looks like this PR failed CI because the test you added is being run even when Nokogiri is built against the "system" libxml2 (in our CI, this is Ubuntu). We should skip this test unless we're running with the vendored libxml2.

An example of how to do this is in https://github.com/sparklemotion/nokogiri/blob/master/test/html/test_attributes.rb

Basically, skip the test unless `Nokogiri::VersionInfo.instance.libxml2? && Nokogiri::VersionInfo.instance.libxml2_using_system?`

Does that make sense?

ksolo:
@flavorjones That makes perfect sense. Thank you!

flavorjones:
:+1:

PR #1880, opened 2019-03-04
Fix test class name to match the filename

larskanis:


flavorjones:
This went red on what looks to be a false positive from valgrind. Going to close and reopen to trigger CI again.

PR #1881, opened 2019-03-05
Update Appveyor-CI

larskanis:

It updates the Appveyor setup to check the build and the test suite on supported MRI ruby versions.
Since https://ci.nokogiri.org doesn't do any tests on Windows currently, I propose to add Appveyor in the meantime instead to nokogiris github repository.

Appveyor-CI setup currently runs MRI only and doesn't run JRuby.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,11 +18,10 @@ test_script:
 environment:
   matrix:
     - ruby_version: head-x64
+    - ruby_version: 26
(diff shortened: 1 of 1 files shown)


larskanis:
This is how the Appveyor output currently looks like: https://ci.appveyor.com/project/larskanis/nokogiri/builds/22824137

flavorjones:
Thank you for this! Merging.

PR #1890, opened 2019-04-12
Test for handler functions not interfering with CSS id queries

twalpole:
This PR adds a test for an issue that shows up when using JRuby.  When using JRuby css id queries (at least - maybe others too) where the id partially contains the name of a function in the passed in handler causes the CSS to return no elements.  This does not happen on MRI.  Not really sure where to even start looking for the root cause.

### DIFF
--- a/test/files/tlm.html
+++ b/test/files/tlm.html
@@ -827,6 +827,7 @@ page.<span class="me1">body</span> =~ /&lt;textarea<span class="br0">&#91;</span
 </div>
 
 <div id="abc.123" class='special.character'>Special character div</div>
+<div id="partial_collision_id">Partial collision id</div>
 <div id="footer">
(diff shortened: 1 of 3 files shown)


twalpole:
I think the issue is at https://github.com/sparklemotion/nokogiri/blob/master/ext/java/nokogiri/XmlXpathContext.java#L123 where it appears to indiscriminately replace any matching text in the query whether it's something that could be a function call or not -

flavorjones:
Thanks for submitting this. Tagging @jvshahid so he's aware and maybe we can pair on it.

flavorjones:
This issue goes back to 468c757 in 2010 when XPath custom handler support was first added. It's because a very rough method is being use to try to infer method calls. Will stare at it a bit today.

flavorjones:
See https://github.com/sparklemotion/nokogiri/issues/2147 for a suggested future path around this functionality.

flavorjones:
Will be fixed in v1.11.0.rc4, sorry for the delay.

PR #1895, opened 2019-04-18
Support direct combinators in CSS :has()

### LINKED ISSUE #688: Should support CSS selector: parent:has(>child)
nokogiri supports the CSS selector `parent:has(descendant)` but does not support `parent:has(>child)`. jQuery supports both, so it would be great if nokogiri would also support `parent:has(>child)`.

jonathanhefner:

Adds support for CSS queries like `"a:has(> b)"`, `"a:has(~ b)"`, and `"a:has(+ b)"`.

Closes #688.

Yes, I think so.

No.

### DIFF
--- a/lib/nokogiri/css/xpath_visitor.rb
+++ b/lib/nokogiri/css/xpath_visitor.rb
@@ -51,7 +51,8 @@ module Nokogiri
         when /^comment\(/
           "comment()"
         when /^has\(/
-          ".//#{node.value[1].accept(self)}"
+          is_direct = node.value[1].value[0].nil? # e.g. "has(> a)", "has(~ a)", "has(+ a)"
+          ".#{"//" if !is_direct}#{node.value[1].accept(self)}"
         else
           args = ['.'] + node.value[1..-1]
           "#{node.value.first}#{args.join(', ')})"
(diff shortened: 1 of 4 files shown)


lib/nokogiri/css/xpath_visitor.rb
```diff
@@ -51,7 +51,8 @@ def visit_function node
         when /^comment\(/
           "comment()"
         when /^has\(/
-          ".//#{node.value[1].accept(self)}"
+          is_direct = node.value[1].value[0].nil? # e.g. "has(> a)", "has(~ a)", "has(+ a)"
```
jonathanhefner: Is the comment here gratuitous?  Is there a more self-documenting name for `is_direct`?

flavorjones: I'm fine with both the comment and the name. No worries!

flavorjones:
@jonathanhefner Thanks for submitting this PR! I've added some tests that run CSS queries on a real document to make sure I understand how these pseudoselector queries work. Should be through CI shortly!

flavorjones:
Sorry for the force-push, rebased onto current master.

flavorjones:
Thank you! This will be in v1.11.0 when it ships. Watch the [milestone](https://github.com/sparklemotion/nokogiri/milestone/18) for details.

jonathanhefner:
@flavorjones Thank you for adding those tests and merging this in! :smiley:

PR #1911, opened 2019-06-28
Use private constant to test broken libxml version

ashmaroli:

**Unnecessary memory usage.**

### The issue

While profiling a project for optimizing memory usage, I was pointed to the following line:
```
allocated memory by location
-----------------------------------
  14.61 MB  nokogiri-1.10.3/lib/nokogiri/xml/node.rb:852
```
which is:
https://github.com/sparklemotion/nokogiri/blob/c45658aa86da6cf238216d2cf8bf153dada27f64/lib/nokogiri/xml/node.rb#L852

If the above code is frequently called, then **for every call**:
  - A new array is allocated via `%w[2 6]`
  - An interim array is generated via `LIBXML_VERSION.split('.')`
  - And yet another array is generated via `Array#[]<range>`

### The Solution

The solution is to avoid generating objects and directly check the version string against a substring.
But since its wasteful to repeatedly test a `CONSTANT` string, stash the entire evaluation in a private constant and then read its value as required.

## Have you included adequate test coverage?

This doesn't change behavior nor introduce a new behavior.

## Does this change affect the C or the Java implementations?

No, it doesn't.

### DIFF
--- a/lib/nokogiri/xml/node.rb
+++ b/lib/nokogiri/xml/node.rb
@@ -847,7 +847,7 @@ module Nokogiri
         node_or_tags
       end
 
-      IS_BROKEN_LIBXML_VERSION = LIBXML_VERSION.start_with?('2.6').freeze
+      IS_BROKEN_LIBXML_VERSION = defined?(LIBXML_VERSION) && LIBXML_VERSION.start_with?('2.6').freeze
       private_constant :IS_BROKEN_LIBXML_VERSION
 
       def to_format save_option, options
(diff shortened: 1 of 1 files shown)


flavorjones:
Thanks for pointing out that we could be more efficient with memory here. It looks like this PR is failing on JRuby, I'll take a look this weekend and may end up editing this patchset to make it all work.

ashmaroli:
@flavorjones Thank you for informing me about the failed check.
I've pushed a commit to check if `LIBXML_VERSION` is defined beforehand.
*Note: I didn't use `Nokogiri.uses_libxml?` to avoid invoking a VersionInfo instance.*

flavorjones:
Merged manually! Thank you again for this contribution.

PR #1916, opened 2019-07-26
Reduce array allocations via ary.each_with_object

ashmaroli:

Unnecessary object allocations:
- `array.map` duplicates given Array `array`
- `Array#[]` generates a new Array instance.

Therefore, `Hash[array.map { |e| [e, foo] }]` generates multiple interim arrays on each call and iteration.

Not necessary since this is just refactoring existing implementation.

Does not affect C or Java code


flavorjones:
Merged manually! Thank you!

flavorjones:
(see 63c2b74)

Issue #1764, opened 2018-05-21
investigate using XML_PARSE_BIG_LINES to tell libxml2 to track line numbers bigger than a short int


flavorjones:
## Context

libxml2 has a limitation that it tracks the line numbers in the input with a short int, meaning that long files have odd behavior, either lines having a number of `0` or of `65535`, see related issues #1493, #1617, #1505, etc.

## New Information

In `parser.h` since libxml 2.9.5, there's a parse option `XML_PARSE_BIG_LINES` which stores big line numbers in a separate field in the node struct, `psvi`. I think this might be seamless because Nokogiri's `xml_node.c` calls `xmlGetLineNo` and all the logic we need is in there.

## Idea

Let's explore:

- add `XML_PARSE_BIG_LINES` to `XML::ParseOptions`
- see if we can affect the related issues (linked above) by setting this option.

If it seems to work, then:

- add some tests - xml doc, htmll doc, xsd schema, relaxng schema ... (any others?)
- add that bit to the `XML::ParseOptions::DEFAULT_*` masks to make the tests pass
- update the CHANGELOG
- close all the issues

flavorjones:
See #2309.

PR #1658, opened 2017-06-20
Add API to set line number, fixes #1657

fulldecent:
This adds a way to set XML node line numbers.

Why would you want to do that? Constructed nodes. See one implementation that needs this here https://github.com/rubys/nokogumbo/issues/53#issuecomment-309504802

Sorry, this is maybe a useless contribution, I don't actually know Ruby, just pretending.

This fixes issue #1657.

Issue #1223, opened 2015-01-16
Node#line heuristic under JRuby is hacky and should be smarter


camertron:
With MRI 2.1.5, line numbers are correctly reported. With any recent version of JRuby, they aren't:

```
$> rbenv local ruby-2.1.5
$> bundle console
irb> require 'nokogiri'
irb> puts Gem.loaded_specs['nokogiri'].version
1.6.5
irb> doc = Nokogiri::XML('<root><foo>bar</foo>\n\n<foo>baz</foo></root>')
irb> root = doc.xpath('/root').first
irb> root.child.first.line
1
irb> root.child.last.line
3
```

And then:

```
$> rbenv local jruby-1.7.15
$> bundle console
irb> require 'nokogiri'
irb> puts Gem.loaded_specs['nokogiri'].version
1.6.5
irb> doc = Nokogiri::XML('<root><foo>bar</foo>\n\n<foo>baz</foo></root>')
irb> root = doc.xpath('/root').first
irb> root.child.first.line
1
irb> root.child.last.line
1
```

As you can see, under JRuby nokogiri incorrectly returns `1` for the line number of the second child under `root`.

flavorjones:
Hello @camertron,

Thanks for reporting this! I've captured this in a slightly different repro script:

```ruby
#! /usr/bin/env ruby

require 'nokogiri'

doc = Nokogiri::XML <<EOXML
<root>
  <foo>
    bar
  </foo>
  <foo>
    baz
  </foo>
</root>
EOXML
puts doc.to_xml
doc.css("foo").each { |node| puts node.line }
```

MRI outputs:
```
<?xml version="1.0"?>
<root>
  <foo>
    bar
  </foo>
  <foo>
    baz
  </foo>
</root>
2
5
```

JRuby outputs:
```
<?xml version="1.0"?>
<root>
  <foo>
    bar
  </foo>
  <foo>
    baz
  </foo>
</root>
2
4
```

As you can see, when compared to your script, JRuby appears to use a different heuristic when reporting the line number than MRI does. This is likely due to the underlying implementations being different XML libraries (libxml2 for MRI, and xerces for JRuby).

Indeed, looking at the code, in MRI we call a libxml2 method to get the internal line number from the parser:

> https://github.com/sparklemotion/nokogiri/blob/master/ext/nokogiri/xml_node.c#L1247

And in JRuby, we look at a counter maintained in the JRuby extension (not the Xerces parser):

[...]

PR #1918, opened 2019-08-10
Add `Nokogiri::XML::Node#line=`

stevecheckoway:
Add support for setting the line number on elements from Ruby. This only
works when Nokogiri uses `libxml2`.

Line numbers less than 65535 may be set. Larger line numbers require
`libxml2` to be compiled with a particular option and even then, it is
buggy. See https://bugzilla.gnome.org/show_bug.cgi?id=767138.

---

Thank you for contributing to Nokogiri! To help us prioritize, please take care to answer the questions below when you submit this pull request.

This PR is intended to solve the issues for downstream projects. In particular, https://github.com/gjtorikian/html-proofer/pull/362 in html-proofer which is preventing that from using Nokogumbo for HTML5 parsing.

This implementation is similar to https://github.com/sparklemotion/nokogiri/pull/1658 (which I think was broken because the option, `XML_PARSE_BIG_LINES`, mentioned in https://github.com/sparklemotion/nokogiri/issues/1764 is not used). This version simply limits the number of lines to a maximum of 65534. This limitation could be removed if `XML_PARSE_BIG_LINES` is used, but that could be done later.

Note that the implementation of `set_line()` is slightly different from https://github.com/sparklemotion/nokogiri/pull/1658 which returns `self` rather than the argument. `set_native_content()` and `set_name` return the argument so I followed that approach.

It is tested in the same manner as `Nokogiri::XML::Node#line`. (I think that should be better tested, but JRuby's line number heuristic seems like it would be easy to break and I'm not sure how to use JRuby to test.)

This effects the C version only. The Java version uses a heuristic to get line numbers but that heuristic [doesn't quite work](https://github.com/sparklemotion/nokogiri/issues/1223).

### DIFF
--- a/ext/nokogiri/xml_node.c
+++ b/ext/nokogiri/xml_node.c
@@ -1334,6 +1334,25 @@ static VALUE line(VALUE self)
 
 /*
  * call-seq:
+ *  line=(num)
+ *
+ * Sets the line for this Node. num must be less than 65535.
+ */
+static VALUE set_line(VALUE self, VALUE num)
+{
+  xmlNodePtr node;
+  Data_Get_Struct(self, xmlNode, node);
+
+  int value = NUM2INT(num);
+  if (value < 65535) {
+    node->line = value;
+  }
+
+  return num;
+}
+
+/*
+ * call-seq:
  *  add_namespace_definition(prefix, href)
  *
  * Adds a namespace definition with +prefix+ using +href+ value. The result is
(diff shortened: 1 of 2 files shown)


flavorjones:
Thanks for submitting this, I'll do my best to take a look in the next day or so.

gjtorikian:
Any updates on this?

flavorjones:
@gjtorikian thanks for poking. my personal life is nuts right now and have had limited time to devote to new-feature and PR velocity (have only managed to get security-related updates out for some packages). I'm aligned with this feature but haven't had time to review details and merge it. That said, I expect this will be in v1.11.0 and have tagged it for that milestone. ETA is in the next week or two (optimistically).

gjtorikian:
Ah, no worries. Self care is important and I hope things settle down for you. 💜

flavorjones:
Thanks for your patience. Merging. Will be in v1.11.0, due out in late December (coincident with the Ruby 2.7 release).

flavorjones:
Please note that this was included in v1.11.0.rc1, so if you want to test, you can use that version.

stevecheckoway:
Great, thanks!

PR #1921, opened 2019-08-23
Add RuboCop configuration files

eduardoj:

Adding RuboCop configuration files allows to define the configurable attributes of each RuboCop cop, and to track the files that still don't pass the RuboCop cops. It will make it easier to work with RuboCop. 

*Not required.*

No, it doesn't.


flavorjones:
Hi, @eduardoj. Thanks for submitting this.

Can you help me understand who's using Rubocop and would need this configuration? As a maintainer of Nokogiri I find it very helpful to understand who's using everything in the repository.

I'm asking this question because, to my knowledge, none of the contributors or maintainers to Nokogiri are using RuboCop, and I worry about adding more unused configurations.

Thanks in advance for helping me understand!

flavorjones:
Amendment to earlier statement: we're using the Security cop now, but that's being controlled via commandline arguments in the Rakefile, and not a config file.

eduardoj:
> Amendment to earlier statement: we're using the Security cop now, but that's being controlled via commandline arguments in the Rakefile, and not a config file.

Introducing the security cops was a good idea. I thought it would be good too to take advantage of the rest of the work done in RuboCop, checking the code with the rest of the cops.

I still think that keeping the RuboCop configuration and TODO files, even if they are used only for including the security cops, is the way to work with RuboCop in a project. At a glance every future contributor has an idea of how RuboCop is configured, and what else is left to do.

In both cases, I would like to contribute to this project, using RuboCop to apply the guidelines outlined in the Ruby Style Guide.

flavorjones:
@eduardoj It sounds like you're interested in becoming a contributor? That's great!

I'd like to reflect for a moment that onboarding new contributors has not been something we've ever focused on; and I'm unfamiliar with most of the functionality of RuboCop. If you feel strongly that having RuboCop will help onboarding new contributors, then I'd like to better understand how.

If, on the other hand, you're simply advocating for RuboCop in the abstract, without the intention of joining the small team of contributors, then I'd prefer to wait until I or one of the other contributors sees enough value in it to invest the time. Adding a config file to RuboCop without also having the intention of paying attention to it is not a good outcome, in my opinion.

flavorjones:
Closing. If another maintainer feels good enough about RuboCop to maintain a 1,500-line config file then I'm open to pulling something in. But otherwise this is simply increasing the surface area that needs to be maintained.

PR #1922, opened 2019-08-25
Update readme tutorial link

pkayokay:


flavorjones:
Hi, thank you for submitting this patch!

There's a limitation in MkDocs (which is the rendering platform we use for nokogiri.org) that doesn't allow linking to `/tutorials` and getting redirected to the first subpage. I'd prefer not to link directly to an arbitrary tutorial from the README, because that requires a bit too much knowledge of how the tutorials are laid out (which might change in the future).

Instead, what I'd like to do is detailed at https://github.com/sparklemotion/nokogiri.org/issues/24 which is to create a landing page for tutorials and having the README point there.

I hope this solution will address your original concern, which I'm assuming is that you end up on the nokogiri.org home page and need to click one more time to get to a tutorial. Let me know if I've misunderstood.

Thanks again, and sorry I won't be accepting this pull request at this time.

pkayokay:
Gotcha, thanks @flavorjones !

PR #1924, opened 2019-09-18
Implement a mark function so we're compaction friendly

tenderlove:
If we don't manually mark members of the struct, they can move out from
underneath us.  For now lets just mark them so everything stays stable.


tenderlove:
Oh we can totally delete it I just forgot to hit the button

PR #1930, opened 2019-10-08
[fix] avoid doing (Java) document factory finder lookup

kares:
every time a new XML document is being created, since it causes class-path scanning 
(considerable slow-downs esp. in multi-threaded envs)

samples of captured lock waits (note the stack trace bellow pointing to `XmlDocument.new`) :

nokogiri1

nokogiri2

hard to test multi-threaded behaviour with internal (JVM) locking

change is a Java specific fix that improves Nokogiri performance under JRuby

### DIFF
--- a/ext/java/nokogiri/HtmlDocument.java
+++ b/ext/java/nokogiri/HtmlDocument.java
@@ -70,15 +70,15 @@ public class HtmlDocument extends XmlDocument {
     }
 
     @JRubyMethod(name="new", meta = true, rest = true, required=0)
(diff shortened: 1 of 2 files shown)


flavorjones:
Hi @kares - I'm pretty sure this change was rolled into the PR at #1934, but wanted to confirm before I close this.

kares:
yes, its in https://github.com/sparklemotion/nokogiri/pull/1934/commits/827d7ba64dccc2072ee1ef71c001b7c97018de40 (from #1934) ... thanks for the merge (should make some folks happy as it gets out)!

PR #1934, opened 2019-10-16
[jruby] cleanup/refactor extension code

kares:
There's some anti-patterns regarding JRuby ext / Java to revisit (still an ongoing effort).
The Java ext part is confusing at times - initializing parts twice (e.g. XmlDocument's node).

Relevant changes : 

- avoid ns-cache being double backed by a map + another key list (use just map backing)
  also introduce a cache-key as using arrays was weird - potentially leaky on some scenarios
- do not construct ext wrapper objects went internally XML API objects are sufficient
- cleaner and more explicit way of initializing (document) internals of XmlNodeSets
- align with JRuby's toJava convention (will also now work for any Node not just Document)
- review some of the internals and remove dead-code
- do not silently swallow potentially fatal exception from backend

Nokogiri is a bit heavy on memory on times - there's a few attempts to reduce allocations here.

**NOTE:** includes https://github.com/sparklemotion/nokogiri/pull/1930 (to avoid potential conflict resolution) should be without conflicts after its merged.

This also could use some raw performance numbers.

### DIFF
--- a/ext/java/nokogiri/HtmlDocument.java
+++ b/ext/java/nokogiri/HtmlDocument.java
@@ -47,6 +47,8 @@ import org.w3c.dom.NodeList;
 
 import nokogiri.internals.HtmlDomParserContext;
 
+import static nokogiri.internals.NokogiriHelpers.getNokogiriClass;
+
 /**
  * Class for Nokogiri::HTML::Document.
  *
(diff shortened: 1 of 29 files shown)


flavorjones:
@kares Thanks so much for all the work that went into this patch set. Very much appreciated!

PR #1942, opened 2019-11-15
Do not override encoding set during initialisation

kp666:
encoding set during initialisation of sax parser gets replaced to ASCII when parsing.
Fix: check if encoding has already been set.

### DIFF
--- a/lib/nokogiri/xml/sax/parser.rb
+++ b/lib/nokogiri/xml/sax/parser.rb
@@ -88,9 +88,8 @@ module Nokogiri
 
         ###
         # Parse given +io+
-        def parse_io(io, encoding = "ASCII")
-          @encoding = check_encoding(encoding)
-          ctx = ParserContext.io(io, ENCODINGS[@encoding])
+        def parse_io(io, encoding = @encoding)
+          ctx = ParserContext.io(io, ENCODINGS[check_encoding(encoding)])
           yield ctx if block_given?
           ctx.parse_with(self)
         end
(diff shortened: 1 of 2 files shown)


flavorjones:
Hi, there's a CI failure for this PR that I think is related, as it's an encoding test:

https://ci.nokogiri.org/teams/nokogiri-core/pipelines/nokogiri-pr/jobs/ruby-2.6-system/builds/60

I'd also like this PR to contain a test.

kp666:
@flavorjones I am not sure if to mutate encoding while passing encoding via parse_io.
please give directions here.

flavorjones:
The current CI failures are not due to this commit -- will take a look now.

flavorjones:
ok - remaining failures would be addressed by a rebase onto master, regardless this is OK to merge ...

BUT I want to think about edge cases before doing so.

kp666:
My issue was that even when i set encoding to 'UTF-8' when the parser is initialised, parser.parse was calling parse_io without encoding. This meant that encoding was getting changed to 'ASCII' as per the old code. 
IMO parse_io api seems a bit weird since parse_file and parse_memory apis do not take an additional encoding argument, and neither does mutate the 'encoding' instance variable, where as parse_io is capable of mutating encoding.

flavorjones:
Just rebased onto master to get the checks green ...

flavorjones:
Rebased again to try to get CI green ...

flavorjones:
I've rebased this again onto current `main`

flavorjones:
LGTM! Merging. Thank you for your patience!!!

flavorjones:
I've credited you in the changelog in 243eab5. This will be in v1.14.0.

Issue #1253, opened 2015-02-21
use jar-dependencies for requiring jars


mkristian:
this gives back the control of which jars are loaded when. you can use jbundler or maven or gradle to manage the jars. actually you manage gems and jars almost in the same manner.

it also gives you a chance to work around certain classloader problems when you really need to not delegate the classloading to the parent classloader.

currently I am working on pysch to do the same
https://github.com/mkristian/psych/blob/jruby-build/Rakefile#L41
since we want have psych as default gem in jruby-9k

jruby-openssl is an example to use jar-dependencies with still packing the jars but use jar-dependencies when available (so older jrubies work as well)
https://github.com/jruby/jruby-openssl/blob/master/jruby-openssl.gemspec#L23
https://github.com/jruby/jruby-openssl/blob/master/lib/jopenssl/load.rb#L8

mkristian:
see also https://github.com/guyboertje/jrjackson/issues/35

flavorjones:
Hi there,

Sorry again, but you'll need to fully explain what the problem you're trying to solve is. Not being a full-time Java developer, I have no idea what `jar-dependencies` is, what problem it is supposed to help, what the tradeoffs are to supporting it, nor what alternative solutions were considered.

Let's please just start with a clear description of the problem? We can then proceed to a discussion of solutions. Thanks for your patience.

mkristian:
in jruby you can use gems and jars almost in the same manner as they
provide libraries your code can use.

gems are managed by rubygems/bundler and they usually declare their
gem-dependencies inside their gemspec - to vendor gems is not that common
(any more).

in the java world jar can have dependenies and they can be declared as well
- used by maven, buildr, ivy, gradle, etc

BUT gems like nokogiri vendor their jars, i.e. they get loaded at some
point in time into the jruby-classloader. there is no way to find out which
jars with which version was loaded since in java the jars do not contain
such meta information (those tools above provide thise meta data on side).

let say I have a java library depending on xerces or xalan and my jruby
project is using it along nokogiri. tools like jbundler or maven or gradle
can handle the java libraries for your project but they can NOT inlcude
those vendored jars of nokogiri into account. so you end up with loading
the same jar TWICE into the jruby classloader - maybe even with different
versions !

so jar-dependencies first gives a way to "declare" the jar inside the
gemspec in a similar manner as you would declare gem dependencies:
https://github.com/jruby/jruby-openssl/blob/master/jruby-openssl.gemspec#L23

[...]

PR #1967, opened 2020-01-14
Start using jar-dependencies in the gemspec.

headius:
This change affects the Java gem and is an attempt to start using Maven artifacts for all the jar dependencies in Nokogiri.

See #1253. The commented versions can't be found on Maven Central. Until we can resolve them, they would have to remain as bare jar files in the repository and gem.

Next step for this PR would be to get the gem building using jar-dependencies and remove the properly-versioned jars from the repository.


jvshahid:
For future reference, I tried `org.codelibs:nekohtml,` but the latest version `2.1.0` only supports Java version 11 and up.  When I ran the tests against version `2.0.2`, the newest version that supports Java 8, I saw four tests fail.

I also submitted a patch to the Nekohtml project on [sourceforge](https://sourceforge.net/p/nekohtml/patches/24/). I am not sure if that will go anywhere, given that the mailing list hasn't been active in a few years.

jvshahid:
@headius or @mkristian can either one of you help me understand how to use `jar-dependencies` in development?  I tried to check out this branch locally, delete the jars declared in `Rakefile` from `lib/,` and run `bundle exec rake compile.`  I expected the jars to be available on the classpath for the compiler. Instead the java compiler failed because it couldn't find the classes.

headius:
@jvshahid Will have to wait for mkristian...I'm not clear on how to use jar-deps in dev either.

mkristian:
Will make same time this week to integrate jar-dependencies. Will do it in a way it fits with rake work flow. Will ping you guys.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

mkristian:
I have something working but I am unsure what the best/easiest way to manage this.

one way would be to add a rake tasks which updates the jars and otherwise wire up jar-dependencies to use the installed jars.

mkristian:
@headius @jvshahid please see my branch
https://github.com/mkristian/nokogiri/tree/jar-deps

* load all jar via jar-dependencies using it maven coordinate
* add rake task to update jar using Jarfile as source
* using Jarfile to declare jar deps in gemspec

a few jars are not part of maven central so I copies them to right place, i.e. they will be still part of the duplicate jar warning/detection of jar dependencies. the jing.jar will start up but produces NullPointerException in tests. Still need to look into this - as I did see the same errors when using the original jar. 

running tests I used `rake test` or is there something else as I do see errors on master as well.

flavorjones:
The underlying rakefiles and gemspec have evolved enough that I'm going to close this and start a new PR. I'll tag @headius and @mkristian for review and feedback on it.

PR #1969, opened 2020-01-14
Change XmlNodeSet#to_a return type to RubyArray

### LINKED ISSUE #1968: XmlNodeSet.to_s return value conflicts with JRuby 9.2.9
XmlNodeSet defines a `to_a` method here: https://github.com/sparklemotion/nokogiri/blob/84a554f0c7b78bf82bdd928d27ae038cf29dd0bb/ext/java/nokogiri/XmlNodeSet.java#L395

JRuby originally defined a superclass `to_a` method (in RubyBasicObject) that did not take a `ThreadContext` parameter. However in JRuby 9.2.9, we made another set of changes to pass `ThreadContext` through more methods, including the `to_a` in RubyBasicObject. This added a new signature `RubyArray to_a(ThreadContext)` with the old `RubyArray to_a()` left deprecated.

As a result, the method in XmlNodeSet no longer compiles because it attempts to return a more general type (IRubyObject as opposed to RubyArray).

headius:
The existing signature conflicts with one added to JRuby 9.2.9.
Specifically, the new signature in JRuby returns RubyArray, which
causes a compilation error on this line in Nokogiri because it
attempts to use a more general return type.

We would prefer to keep the specific return type in JRuby.

* If we patch JRuby, then 9.2.9 will never be able to compile any
  version of Nokogiri.
* If we patch Nokogiri, all versions of JRuby can compile current
  and future Nokogiri. Versions prior to 9.2.9 will be able to
  compile all existing releases of Nokogiri.

I do not believe the change in 9.2.9 breaks anything at runtime,
since the JVM does not care about this particular return type
mismatch unless someone actually returns a non-RubyArray object.

Fixes #1968


jvshahid:
The changes look good to me.  I don't think the failures are related to the change.  The failure was in the MRI valgrind tests.  I am waiting to see the tests pass before I merge the PR.

flavorjones:
Yeah, looks like some optimizations were backported to ruby 2.5, so I just added a valgrind suppression and will kick off again.

flavorjones:
@headius I see you based this commit on the `v1.10.x` branch. For future reference, we ask that people send PRs against `master`, and then we can rebase against the v1.10.x branch if we need to cut a release.

I'm not totally familiar with the JRuby 9.2.9 change that is driving this commit, but am I to understand that you're requesting a bug fix release in the v1.10.x series?

flavorjones:
Remaining CI failure is due to #1971, looking into it. We could merge this PR, but I'd like to leave it open for a bit so I can ensure the CI job is stable.

headius:
@flavorjones In JRuby 9.2.8, the only `to_a` defined on our `RubyBasicObject` class was `RubyArray to_a()`. In JRuby 9.2.9, we added `RubyArray to_a(ThreadContext)`. Nokogiri's `XmlNodeSet` previously defined a `IRubyObject to_a(ThreadContext)` signature. Since `XmlNodeSet` descends from `RubyBasicObject`, method signatures it overrides need to be compatible. In this case, because `IRubyObject` is a the more generic interface type than `RubyArray`, it fails to compile.

So the addition of the new `to_a` in JRuby broke compilation of Nokogiri because Java. At runtime it probably doesn't matter since the `XmlNodeSet#to_a` still returns a RubyArray, but the compiler doesn't like the signatures not matching.

This fix should go into any active Nokogiri branches going forward. I did not do the PR against master because it appear to be behind the branch.

flavorjones:
@headius Going forward please submit PRs based on `master`, thanks.

flavorjones:
Merged manually onto master. Will be in v1.11.0. Thank so much, @headius!

flavorjones:
Side note: this actually went out in v1.10.9, see https://my.diffend.io/gems/nokogiri/1.10.8-java/1.10.9-java

headius:
> @headius Going forward please submit PRs based on master, thanks.

A suggestion: put this in the project PR template, perhaps?

Thanks for merge and release!

flavorjones:
@headius Great minds think alike: 787fed8 :rofl: 

Thanks again.

Issue #1965, opened 2020-01-06
Fix code coverage calculation


flavorjones:
Sometime between Sept 8 and Sept 15, 2019, code coverage broke and is being calculated before tests are actually run.

flavorjones:
You can see this at https://codeclimate.com/github/sparklemotion/nokogiri/trends/test_coverage_total

![image](https://user-images.githubusercontent.com/8207/71828450-5e03d000-3070-11ea-867b-a7201a9ee74f.png)

flavorjones:
I've determined this is an issue in Hoe v3.18.0 and higher. Sending a PR.

flavorjones:
PR submitted at https://github.com/seattlerb/hoe/pull/102

flavorjones:
Pinning hoe to v3.17.x temporarily fixes this:

![image](https://user-images.githubusercontent.com/8207/73177821-f7f1f200-40dd-11ea-963a-a39c36a66f8d.png)

zenspider:
Released! Unpin, update, and you should be good.

flavorjones:
Unpinned in f56cbc6 on 2020-02-10. Closing.

PR #1970, opened 2020-01-15
Use the same empty string literal

ashmaroli:

**Avoidable extra memory allocation**

While profiling a project to optimize memory usage, the following came to my notice:
```
Allocated String Report
-----------------------------------
    154375  ""
     88132  nokogiri-1.10.7/lib/nokogiri/xml/node.rb:767
```
which is:
https://github.com/sparklemotion/nokogiri/blob/e6b3229ec53ddf70f1d198bba0d3fc13fde842a8/lib/nokogiri/xml/node.rb#L767

The explanation is that **`'a string of any length' * 0 == ""`**.

`String#*` generates a *new string* even if string literals are frozen via the pragma.
Therefore, if the above line was executed (with `indent_times == 0`) a 100 times, there would be an allocation of 100 empty strings (irrespective of `indent_text`) .

This PR replaces this deterministic allocation with a single frozen string literal.
(*Hope the native method doesn't mutate it*).

No change in behavior.

N/A

### DIFF
--- a/lib/nokogiri/xml/node.rb
+++ b/lib/nokogiri/xml/node.rb
@@ -763,10 +763,15 @@ module Nokogiri
         end
         indent_text   = options[:indent_text] || ' '
 
+        # Any string times 0 returns an empty string. Therefore, use the same
+        # string instead of generating a new empty string for every node with
+        # zero indentation.
+        indentation = indent_times.zero? ? '' : (indent_text * indent_times)
+
         config = SaveOptions.new(save_options.to_i)
         yield config if block_given?
 
-        native_write_to(io, encoding, indent_text * indent_times, config.options)
+        native_write_to(io, encoding, indentation, config.options)
       end
 
       ###


flavorjones:
@ashmaroli Thanks for submitting this! The failing test is related to coverage diffs, and that's related to the config issue documented in #1965. This will be in v1.11.0!

ashmaroli:
@flavorjones Glad to know that this will be included in `v1.11.0`. Thank you for informing :)

ashmaroli:
Strange.. this didn't have an affect at all.. :frowning_face: 

```
Allocated String Report
-----------------------------------
    154390  ""
     88132  nokogiri-1.10.7/lib/nokogiri/xml/node.rb:767
```
https://github.com/sparklemotion/nokogiri/blob/e6b3229ec53ddf70f1d198bba0d3fc13fde842a8/lib/nokogiri/xml/node.rb#L767

---

```
Allocated String Report
-----------------------------------
    154393  ""
     88132  nokogiri-1.11.0.rc1/lib/nokogiri/xml/node.rb:780
```
https://github.com/sparklemotion/nokogiri/blob/8ce75d120acfe6a04d3e2c158d48050286ad3816/lib/nokogiri/xml/node.rb#L780

flavorjones:
@ashmaroli are you saying you didn't test it both ways before submitting the PR? or are you saying the change didn't make it into 1.11.0? i'm a little confused by your response.

flavorjones:
OK - Please explain to me how you're profiling so I can either revert or fix.

flavorjones:
I've opened a new issue, #1986, to investigate. @ashmaroli I'd love your help there.

Issue #1959, opened 2019-12-20
nokogiri builds incorrectly (exports libxml2's symbols)


x-yuri:
**Describe the bug**

The rumor has it that `nokogiri` builds [incorrectly][1]. From what I can gather it exports (not hides) `libxml2`'s symbols. As a result, if `libxml2.so` is loaded before `nokogiri.so`, `nokogiri` is bound to use the system library, not the one embedded in the binary. And as a resullt it has to [check][2] whether that's the case.

The solution is supposedly to add [`-Wl,--exclude-libs,ALL` or `-fvisibility=hidden`][3].

[1]: https://stackoverflow.com/a/59419800/52499
[2]: https://github.com/sparklemotion/nokogiri/blob/v1.10.7/lib/nokogiri/version.rb#L40
[3]: https://stackoverflow.com/questions/2222162/how-to-apply-fvisibility-option-to-symbols-in-static-libraries

**To Reproduce**

`docker-compose.yml`:

```yaml
version: '3'

services:
  ruby:
    build: .
    volumes:
      - ./:/app
```

`Dockerfile`:

```dockerfile
FROM ruby:2-alpine3.10
RUN apk add build-base imagemagick6-dev
WORKDIR /app
```

`Gemfile`:

```ruby
source 'https://rubygems.org'
gem 'nokogiri', '1.10,7'
gem 'rmagick', '4.0.0'
```

`1.rb`:

```ruby
require 'rmagick'
require 'nokogiri'
```

```
$ docker-compose run ruby sh

/app # cat /etc/issue
Welcome to Alpine Linux 3.10
Kernel \r on an \m (\l)

/app # apk info -vv | grep libxml
libxml2-2.9.9-r3 - XML parsing library, version 2

/app # bundle
Fetching gem metadata from https://rubygems.org/.............
Resolving dependencies...
Using bundler 2.1.2
Fetching mini_portile2 2.4.0
Installing mini_portile2 2.4.0
Fetching nokogiri 1.10.7
Installing nokogiri 1.10.7 with native extensions
Fetching rmagick 4.0.0
Installing rmagick 4.0.0 with native extensions
Bundle complete! 2 Gemfile dependencies, 4 gems now installed.
Bundled gems are installed into `/usr/local/bundle`

/app # bundle exec ruby 1.rb
WARNING: Nokogiri was built against LibXML version 2.9.10, but has dynamically loaded 2.9.9
```

Alternatively,

`1.sh`:

```sh
[...]

PR #1979, opened 2020-02-01
don't export libxml2 symbols

flavorjones:

See #1959 

I'm concerned that without exported symbols, nokogumbo will break. We don't have tests for that. Otherwise, existing test coverage suffices.

Pinging @stevecheckoway for awareness and to ask for validation and input from the nokogumbo project maintainers.

Will only affect CRuby.


flavorjones:
Closing, this breaks Nokogumbo. See details at #1959.

Issue #1983, opened 2020-02-02
Ship precompiled gems for linux


flavorjones:
**Hello!** If you're arriving here to give feedback on the v1.11.0 Release Candidates, please make sure to:

* tell us which linux distro you're using!
* (If you can) a docker image in which we can reproduce the failure you're seeing!

Thanks so much for using Nokogiri and for giving us feedback on this experiment!

-----

See #1571 for the PR from @larskanis that enables building precompiled libraries for linux.

This issue is for discussion on how to ship those precompiled libraries in a gem for linux users in a way that won't break the installation experience for people.

Some things that we'll need to figure out were mentioned in #1571 by @larskanis:

> 1. Bundler isn't clever at selection of unsupported ruby versions. See https://github.com/bundler/bundler/pull/6247 . This requires users to explicit opt-out of binary gems, when they want to run nokogiri on ruby-trunk.
> 
> 2. There is a known issue on Alpine Linux: https://github.com/rake-compiler/rake-compiler-dock/issues/20 . I did a manual test on the alpine docker image and it reproducible fails at `Nokogiri::XML::Reader#namespaces` somewhere within the tests: https://gist.github.com/larskanis/94920901e6466c9383360ee1346f9f34 . I didn't analyze the reason for the segfault so far, but I think we should before the release.
> 

And he later mentioned:

> Issue 1 in the above comment will probably finally be fixed by the next bundler release. See rubygems/bundler#7522 . This way bundler will select the platform=ruby gem automatically, if fat binary gem doesn't fit to the ruby release.
> 
> I think we'll probably also be able to fix / work around issue 2, so that I hope we'll be able to ship binary releases some day.
> 

Finally, I'd like for `VersionInfo` to contain some indication that the user is using precompiled libraries. A patch that worked at one point on the #1571 branch (but now needs to be rewritten) is:

```patch
From 20f69908732b50258be0ee18e1d78b841d462dc4 Mon Sep 17 00:00:00 2001
From: Mike Dalessio <mike.dalessio@gmail.com>
Date: Sun, 13 Jan 2019 01:15:06 -0500
Subject: [PATCH] `nokogiri -v` emits info about precompiled binaries

It may be useful information to be able to tell if the libraries were
compiled on installation, or whether they were precompiled in a fat
binary native gem.
---
 Rakefile                | 3 +++
 ext/nokogiri/nokogiri.c | 6 ++++++
 lib/nokogiri/version.rb | 6 ++++++
[...]

PR #1989, opened 2020-02-04
Remove dependency to "mini_portile2" from binary gems

larskanis:
It is only necessary at build in extconf.rb.

As noted in [this issue](https://github.com/sparklemotion/nokogiri/issues/1983#issuecomment-581850547) binary gems depend on "mini_portile2" gems, although it isn't necessary.

There are no automated tests for binary gems, so far.

Only the C implementation.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -275,6 +275,7 @@ else
 Nokogiri is built with the packaged libraries: #{libs}.
       EOS
       spec.files.reject! { |path| File.fnmatch?('ports/*', path) }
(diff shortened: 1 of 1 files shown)


flavorjones:
Thank you, @larskanis!

flavorjones:
@larskanis :thinking: This change doesn't seem to have any effect. I don't see the `cross_compiling` block being executed when compiling native linux gems (though it is executed when cross-compiling windows gems).

PR #1990, opened 2020-02-07
Fix binary nokogiri gems on musl based Linux

larskanis:

As described in #1983 there are two issues when using the nokogiri binary gem on musl based Linux.

The first commit fixes a segfault, the second commit fixes a failing test case. The the commit description for more details.

Not yet. Tests on Alpine were run per `docker run --rm -it ruby:alpine /bin/sh`.

The C version only.

### DIFF
--- a/ext/nokogiri/xml_reader.c
+++ b/ext/nokogiri/xml_reader.c
@@ -28,35 +28,24 @@ static int has_attributes(xmlTextReaderPtr reader)
 static void Nokogiri_xml_node_namespaces(xmlNodePtr node, VALUE attr_hash)
 {
   xmlNsPtr ns;
-  static char buffer[XMLNS_BUFFER_LEN] ;
-  char *key ;
-  size_t keylen ;
+  VALUE key;
 
   if (node->type != XML_ELEMENT_NODE) return ;
 
   ns = node->nsDef;
   while (ns != NULL) {
 
-    keylen = XMLNS_PREFIX_LEN + (ns->prefix ? (strlen((const char*)ns->prefix) + 1) : 0) ;
-    if (keylen > XMLNS_BUFFER_LEN) {
-      key = (char*)malloc(keylen) ;
-    } else {
-      key = buffer ;
-    }
-
+    key = rb_enc_str_new_cstr(XMLNS_PREFIX, rb_utf8_encoding());
     if (ns->prefix) {
-      sprintf(key, "%s:%s", XMLNS_PREFIX, ns->prefix);
-    } else {
-      sprintf(key, "%s", XMLNS_PREFIX);
+      rb_str_cat_cstr(key, ":");
+      rb_str_cat_cstr(key, (const char*)ns->prefix);
     }
 
+    key = rb_str_conv_enc(key, rb_utf8_encoding(), rb_default_internal_encoding());
     rb_hash_aset(attr_hash,
-        NOKOGIRI_STR_NEW2(key),
+        key,
         (ns->href ? NOKOGIRI_STR_NEW2(ns->href) : Qnil)
     );
-    if (key != buffer) {
-      free(key);
-    }
     ns = ns->next ;
   }
 }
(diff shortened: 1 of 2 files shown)


larskanis:
The test failure is also present, when nokogiri is native compiled on Alpine Linux. It is not specific to the binary gems.

ext/nokogiri/xml_reader.c
```diff
@@ -28,35 +28,24 @@ static int has_attributes(xmlTextReaderPtr reader)
...
+    key = rb_enc_str_new_cstr(XMLNS_PREFIX, rb_utf8_encoding());
     if (ns->prefix) {
-      sprintf(key, "%s:%s", XMLNS_PREFIX, ns->prefix);
-    } else {
-      sprintf(key, "%s", XMLNS_PREFIX);
+      rb_str_cat_cstr(key, ":");
+      rb_str_cat_cstr(key, (const char*)ns->prefix);
     }
 
+    key = rb_str_conv_enc(key, rb_utf8_encoding(), rb_default_internal_encoding());
```
zenspider: I like how much your code deletes! But this seems like a wide departure from the original, with extra string concatenation for something with known size. EIther the string should be allocated ahead of time (which doesn't look that easy to do with cruby's API), or the concatenation should be avoided (easy). Here's the (rough + untested) cruby equivalent to the original code + your encoding differences:
```c
if (ns->prefix) {
  key = rb_enc_sprintf(rb_utf8_encoding(), "%s:%s", XMLNS_PREFIX, ns->prefix);
} else {
  key = rb_enc_sprintf(rb_utf8_encoding(), "%s", XMLNS_PREFIX);
}
key = rb_str_conv_enc(key, rb_utf8_encoding(), rb_default_internal_encoding());

**Comment by zenspider** [comment]
THIS looks awesome. I would love to have nokogiri binary gems on alpine.

File: ext/nokogiri/xml_reader.c (hunk 1 above)
Comment (larskanis) [reply ext/nokogiri/xml_reader.c]: Thank you  @zenspider for reviewing! `rb_enc_str_new_cstr()` allocates a [struct RString](https://github.com/ruby/ruby/blob/master/include/ruby/ruby.h#L952-L965) with an embedded string length of 23 bytes at maximum. So each string has a capacity of 23 ASCII characters from the start. Since namespace keys tend to be shorter, usually no reallocation is required. Given that `sprintf` is usually slower due to format string interpretation, I think the proposed implementation is best how it is.

**Comment by flavorjones** [comment]
Just a note - I want to add alpine to the CI pipeline (per note in #1845) before merging this; I'd like to go red (see the segfault reliably on alpine) first, then merge and go green.

**Comment by flavorjones** [comment]
I've added two new things to the CI pipelines (master and PRs):

* run tests and valgrind on a `musl` system (alpine)
  * https://ci.nokogiri.org/teams/nokogiri-core/pipelines/nokogiri/jobs/ruby-musl-system
  * https://ci.nokogiri.org/teams/nokogiri-core/pipelines/nokogiri/jobs/ruby-musl-valgrind
* build a native linux gem and run it against all supported rubies and a `musl` system (alpine)
  * https://ci.nokogiri.org/teams/nokogiri-core/pipelines/nokogiri/jobs/native-gem-test

These should go red, and I'll manually merge these two commits to make each red build go green.

**Comment by flavorjones** [comment]
Red #1: https://ci.nokogiri.org/teams/nokogiri-core/pipelines/nokogiri/jobs/ruby-musl-system/builds/1

```
  1) Error:
Nokogiri::HTML::TestDocumentEncoding#test_encoding_without_charset:
ArgumentError: invalid byte sequence in UTF-8
    /tmp/build/4d9a0f57/nokogiri/test/html/test_document_encoding.rb:26:in `test_encoding_without_charset'
```

**Comment by flavorjones** [comment]
First commit merged: ff99fa5

**Comment by flavorjones** [comment]
Second red build: https://ci.nokogiri.org/teams/nokogiri-core/pipelines/nokogiri/jobs/native-gem-test/builds/1

```

flavorjones:
Second commit merged: a3d9438

flavorjones:
OK - additional test coverage is all green, after manual merges. Closing.

PR #1995, opened 2020-02-15
fix css parser cache

### LINKED ISSUE #1935: Race condition in Nokogiri::CSS::Parser
```
#!/usr/bin/env ruby

require 'nokogiri'

parser = Nokogiri::CSS::Parser.new

puts Nokogiri::CSS::Parser.cache_on?

threads = []

threads << Thread.new do
	Nokogiri::CSS::Parser.without_cache do
		sleep 1
	end
end

threads << Thread.new do
	sleep 0.1
	
	Nokogiri::CSS::Parser.without_cache do
		sleep 2
	end
end

threads.each(&:join)

puts Nokogiri::CSS::Parser.cache_on?

# Prints false, but should be true.
```

flavorjones:
Fixes #1935


lib/nokogiri/css/parser_extras.rb
```diff
@@ -32,33 +38,22 @@ def clear_cache
...
-        ###
-        # Parse this CSS selector in +selector+.  Returns an AST.
-        def parse selector
-          @warned ||= false
-          unless @warned
-            $stderr.puts('Nokogiri::CSS::Parser.parse is deprecated, call Nokogiri::CSS.parse(), this will be removed August 1st or version 1.4.0 (whichever is first)')
-            @warned = true
-          end
-          new.parse selector
+          set_cache tmp
```
eregon: I think it would be good to have this line in `ensure`, so even if the block raises the state is restored correctly.

flavorjones: Right on, thanks.

flavorjones: See e2c694f

PR #2000, opened 2020-02-25
support "keyword attributes" via new methods on Nokogiri::XML::Node

flavorjones:
Nokogiri::XML::Node currently supports some convenience methods for dealing with CSS classes (`#classes`, `#add_class`, `#append_class`, and `#remove_class`).

There are other HTML attributes, though, that are interpreted as sets of keywords; for example `rel` (see https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel).

Further, some of this keyword-attribute functionality has been built into `Loofah` in a really inconvenient way (see https://github.com/flavorjones/loofah/issues/173).

This PR creates a set of keyword-attribute convenience methods on Node:

* `#kwattr_values`
* `#kwattr_add`
* `#kwattr_append`
* `#kwattr_remove`

and re-implements the CSS class convenience methods using these new, more general, methods.

Finally, this PR improves the documentation for both sets of convenience classes.

### DIFF
--- a/lib/nokogiri/xml/node.rb
+++ b/lib/nokogiri/xml/node.rb
@@ -576,10 +576,81 @@ module Nokogiri
         kwattr_remove("class", names)
       end
 
+      # Retrieve values from a keyword attribute of a Node.
+      #
+      # A "keyword attribute" is a node attribute that contains a set
+      # of space-delimited values. Perhaps the most familiar example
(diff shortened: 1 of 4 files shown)


flavorjones:
This PR is still in progress ... need to finish docs for the kwattr methods.

PR #2005, opened 2020-02-28
Update mini_portile2 gem to 2.5.0

alejandroperea:

Keeping the dependencies updated.

It is not needed, all of the current ones should pass.

It shouldn't.


flavorjones:
Thank you! I've noted your contribution in the CHANGELOG.

PR #2008, opened 2020-03-03
Add a workaround for missing call to cross_compile block

### LINKED ISSUE #1991: precompiled gems should not have a runtime dependency on `mini_portile2`
As pointed out by @voxik in #1983, any gems that are precompiled don't need `mini_portile2` and so we should remove that as a dependency from those gemspecs.

larskanis:
The cross_compiling block wasn't called for the x86_64-linux platform, leading to an additional dependency to mini_portile2 and missing post install message.

Related patch to rake-compiler is here: https://github.com/rake-compiler/rake-compiler/pull/171

Fixes #1991

### DIFF
--- a/tasks/cross-ruby.rb
+++ b/tasks/cross-ruby.rb
@@ -166,6 +166,7 @@ namespace "gem" do
       RakeCompilerDock.sh <<-EOT, platform: plat
         gem install bundler --no-document &&
         bundle &&
+        find /usr/local/rvm/gems -name extensiontask.rb | while read f ; do sudo sed -i 's/callback.call(spec) if callback/@cross_compiling.call(spec) if @cross_compiling/' $f ; done &&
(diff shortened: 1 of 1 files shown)


flavorjones:
@larskanis Do you think the change in https://github.com/rake-compiler/rake-compiler/pull/171 will make it into a `rake-compiler` release sometime soon?

This change, while effective, feels dirty and so I _think_ I'd prefer to bump the dependency in Nokogiri (or even just build locally using rake-compiler `master`) ... but I don't feel strongly about that.

larskanis:
@flavorjones IMHO this patch isn't different to other patches we apply to upstream libraries. I'll take care that this issue will be fixed in the next rake-compiler version, but I wouldn't want to block nokogiri until then.

The grpc.gem have had a whole bunch of such patches to the rake-compiler-dock container. In the meantime I solved most of them in the upstream projects (rake-compiler, rake-compiler-dock and ruby), so that only one is left.

PR #2011, opened 2020-03-13
Update Actions CI

MSP-Greg:
1. Test on Ubuntu, macOS, & Windows, three Ruby versions each.

2. Windows - test with both MSYS2 packages and 'vendored' packages


MSP-Greg:
Something needs to change in this PR.  If you're interested, I'll update it, otherwise please close.

MSP-Greg:
@flavorjones

I've got this updated & passing, running Ubuntu, macOS, & Windows, Ruby 2.5 thru 3.0 & head/trunk.  Each also runs with packaged/vendored & system libxml2 & libxslt.

Actions macOS is having some issues in that all jobs don't seem to be starting.

Currently, it's one workflow file with three matrices, one for each OS.  The six 'groups' (3 x OS, vendored vs system) can be divided any way you want.

Current run:
https://github.com/MSP-Greg/nokogiri/actions/runs/460113261

flavorjones:
Hi @MSP-Greg - thanks for bumping this. I appreciate the work you put into this, but I'm honestly not sure whether I want or need to have all of these CI runs in Github Actions.

I already have a significant investment in our Concourse pipelines at this point, which cover several flavors of Linux, and which are doing some more complex builds which I'm not sure how or whether they can be replicated using Actions. I like the full control I have within Concourse, and I haven't had time to learn more about Actions.

Potentially I could see moving our Windows builds to Actions, which is why I left this open. But Appveyor is OK and isn't a pain point, so what little time I've had this year to work on Nokogiri didn't get spent on looking into this.

If it's OK with you, I'm going to leave this open with the intention of coming back to it in a bit in the hopes of instructions and learning. But honestly if I move all our CI to Actions it's going to need a ton of work to reproduce the builds we already have and so these commits may not get used directly.

MSP-Greg:
@flavorjones - Happy New Year...

The main benefit of Actions is the twenty concurrent Ubuntu/Windows jobs, and five macOS jobs.

I know you've used Concourse and other CI, and Actions doesn't have Windows 32 bit CI.  But, GitHub has done a lot with it. and many people are very familiar with it.  You can always ping me if issues come up or you have questions.  And, it can be organized anyway you want.

I'll push what I've got (update & rebase), the macOS jobs are slowly running.  I did switch it from 11.0 to 10.15, trying to see if they ran...



JFYI, I've used Nokogiri for quite a while as a small piece generating https://msp-greg.github.io/.  Also, I've kept any eye on popular extension gems and Windows, so...

MSP-Greg:
@flavorjones

Long running CI can be bothersome.

Currently, AppVeyor is running eight 64 bit jobs (2.5, 2.6, 2.7, head) and two 32 bit jobs (2.7).  It takes a little less that an [hour](https://ci.appveyor.com/project/flavorjones/nokogiri/builds/37091506).
 
 AppVeyor images are not updated often, Ruby 3.0 isn't yet available, and the MSYS2 tool chain is outdated.

Those two issues are not present with Actions.  I modified the AppVeyor script to run six 32 bit jobs (head, 2.7, & 2.5), and it takes about 35 minutes.  The AppVeyor run is [here](https://ci.appveyor.com/project/MSP-Greg/nokogiri/builds/37092266).  This is assuming that Actions is running the 64 bit jobs.

I haven't pushed the AppVeyor changes to this PR...

As to other types of jobs you use, they can probably be ported over to Actions, but I haven't looked for or at the scripts.  The main difference with Actions is that every step runs in a new process.

Anyway, due to the concurrency, you might consider moving more jobs to Actions.

@larskanis - any thoughts, Windows or otherwise?

larskanis:
I just opened [an issue on Appveyor](https://github.com/appveyor/ci/issues/3598) to add ruby-3.0 to their CI image.

Regarding which CI to use, I don't have a strong opinion. The only issue I have with github actions is that it doesn't work on my smartphone, which is bad since I often push and check on mobile later, when I'm away from the computer.

flavorjones:
@MSP-Greg I really appreciate that you're paying attention to these details. Can you help me understand how Actions will support our most complex CI jobs? Specifically I'm thinking of [cruby-native-gem-test](https://ci.nokogiri.org/teams/nokogiri-core/pipelines/nokogiri/jobs/cruby-native-gem-test/builds/70).

If Actions can completely replace Concourse or completely replace Appveyor, I'm interested. But otherwise spending effort to move/maintain additional CI jobs is not an attractive option for me at this time.

MSP-Greg:
> Specifically I'm thinking of cruby-native-gem-test.

I'll have a look.  A while ago I had scripts to generate pre-compiled gems on AppVeyor.  Similar to what you're doing, it built the gem, then tested it on all versions, and saved the gem as an artifact.

I've thought about how to run them on Actions, as it's somewhat tricky, but I may be back.  Not in hours...

I don't expect Actions to ever have Windows mingw 32 bit.

flavorjones:
I've rolled much of this PR into #2244 (and given you co-author credit). Closing.

PR #2027, opened 2020-04-26
Clean up deprecated calls into JRuby

headius:
This PR removes all deprecation warnings from the Java extension by using newer preferred APIs in JRuby.

The bulk of changes are to use `RubyException.toThrowable()` to construct a Java throwable for a Ruby exception, rather than the deprecated `RaiseException` constructor. This change is the likely the most controversial since it breaks compatibility with JRuby versions older than 9.2.

The other changes are to use RuntimeError instead of NativeException for raising unexpected Java XML errors, and to move away from overriding the old `to_s19` form from JRuby 1.7.

### DIFF
--- a/ext/java/nokogiri/XmlSaxParserContext.java
+++ b/ext/java/nokogiri/XmlSaxParserContext.java
@@ -94,7 +94,8 @@ public class XmlSaxParserContext extends ParserContext {
             parser = createParser();
         }
         catch (SAXException se) {
-            throw RaiseException.createNativeRaiseException(runtime, se);
+            // Unexpected failure in XML subsystem
+            throw runtime.newRuntimeError(se.getMessage());
         }
     }
 
@@ -208,7 +209,8 @@ public class XmlSaxParserContext extends ParserContext {
                 parser.setFeature(FEATURE_CONTINUE_AFTER_FATAL_ERROR, true);
             }
             catch (Exception e) {
-                throw RaiseException.createNativeRaiseException(runtime, e);
+                // Unexpected failure in XML subsystem
+                throw runtime.newRuntimeError(e.getMessage());
             }
         }
     }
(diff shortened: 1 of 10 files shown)


headius:
The failed build is for ruby-head... I assume that's not my fault.

ext/java/nokogiri/XmlSaxParserContext.java
```diff
@@ -94,7 +94,8 @@ protected void initialize(Ruby runtime) {
             parser = createParser();
         }
         catch (SAXException se) {
-            throw RaiseException.createNativeRaiseException(runtime, se);
+            // Unexpected failure in XML subsystem
+            throw runtime.newRuntimeError(se.getMessage());
```
kares, line 98: since this path was using the deprecated `NativeException` to wrap a Java throwable,
we could make it a bit closer to resembling the previous format (and set cause) :
```suggestion
            RaiseException ex = runtime.newRuntimeError(se.toString());
            ex.initCause(se);
            throw ex;
```
.. probably worth extracting to a helper for re-use.
also these changes might need inspecting caller and potentially using a custom wrapping Ruby exception if these were `rescue SAXException` anywhere.

headius: Yeah good call.

headius:
I believe this is ready after the change by @kares.

kares (approved):
:+1:

flavorjones:
Great, thank you both! Will merge this onto master for release in the next v1.11.0 release candidate.

@headius Please let me know if you want this pulled into a v1.10.x (stable) release; I'm assuming it's OK to just ship in a v1.11.0 final later this year.

headius:
@flavorjones Yeah there's no rush on this. 👍

PR #2031, opened 2020-05-10
Kernel#open -> URI#open

neko314:
I made this PR to update sample code.

Run sample codes with Ruby 2.7.0 and later and get decricated warning like below.

```
> Nokogiri::HTML(open('https://nokogiri.org/tutorials/installing_nokogiri.html'))
(irb):33: warning: calling URI.open via Kernel#open is deprecated, call URI.open directly
```

So, I edited the codes not to use `Kernel#open` but `URI#open`.

Tests has no changes because I only changed sample codes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ require 'nokogiri'
 require 'open-uri'
 
 # Fetch and parse HTML document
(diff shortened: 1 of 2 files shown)


flavorjones:
Thank you for your contribution!

PR #2058, opened 2020-08-03
Follow WHATWG guidance for incorrectly-closed HTML comments

flavorjones:
This changeset introduces test coverage for both JRuby and CRuby implementations with respect to handling HTML parse errors within comments.

It then introduces a libxml2 patch to comply with the behavior suggested by the [WHATWG "Living Standard" doc](https://html.spec.whatwg.org/multipage/parsing.html#parse-error-incorrectly-closed-comment) with respect to incorrectly-closed comments.

It's important to note that WHATWG's guidance is non-normative, but is what's followed by most popular browsers. Further, the difference in behavior between browsers and server-side sanitization done by a library relying on Nokogiri and libxml2 allows opportunity for XSS attacks (https://owasp.org/www-project-top-ten/OWASP_Top_Ten_2017/Top_10-2017_A7-Cross-Site_Scripting_(XSS)).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,12 +16,12 @@ We bump `Major.Minor.Patch` versions following this guidance:
 `Minor`:
 
 - Features and bugfixes.
-- Dropping support for EOLed Ruby versions, updating vendored libraries.
-- Updating vendored libraries for non-security-related reasons.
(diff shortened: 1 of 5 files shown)


flavorjones:
Also see related upstream merge request: https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/82

Issue #2063, opened 2020-08-18
Explore precompiled binaries for Darwin (OSX)


flavorjones:
This is a placeholder for a future conversation (with myself?) about precompiling Darwin libraries, similarly to how we've shipped precompiled libraries for Windows since ~v1.4 and how we will hopefully [ship for Linux in v1.11.0](https://github.com/sparklemotion/nokogiri/issues/1983).

cc @larskanis @tenderlove tell me if you think I'm crazy for thinking we can do this.

---

### Here's what I've learned so far

(I'll keep this section up to date as I go along)

__Architecture names relevant to us__

- RUBY_PLATFORM
  - used to identify the ruby architecture
  - which on my linux system is "x86_64-linux"
  - using OSX system Ruby is "universal.x86_64-darwin19"
  - using OSX homebrew- or rvm-installed Ruby is "x86_64-darwin19"
- Gem::Platform.new(RUBY_PLATFORM)
  - used by rake-compiler to name native tasks
  - on my linux system is "x86_64-linux"
  - using OSX system Ruby is "universal.x86_64-darwin-19" (note the addition of a "-" before the "19")
  - using OSX homebrew- or rvm-installed Ruby is "x86_64-darwin-19" (note the "-" here also)
- Gem::Platform.local
  - used by rubygems to search for an appropriate native platform gem
  - on my linux system is "x86_64-linux"
  - using OSX system Ruby is "universal-darwin-19"
  - using OSX homebrew- or rvm-installed Ruby is "x86_64-darwin-19"

__Needed changes to Nokogiri to support OSX__

Invoke the properly-named gem task:

```patch
diff --git a/tasks/cross-ruby.rb b/tasks/cross-ruby.rb
index 35c098f..1d5c2fd 100644
--- a/tasks/cross-ruby.rb
+++ b/tasks/cross-ruby.rb
@@ -179,7 +181,7 @@ namespace "gem" do
       task "guest" do
         # use Task#invoke because the pkg/*gem task is defined at runtime
         Rake::Task["native:#{plat}"].invoke
-        Rake::Task["pkg/#{HOE.spec.full_name}-#{plat}.gem"].invoke
+        Rake::Task["pkg/#{HOE.spec.full_name}-#{Gem::Platform.new(plat).to_s}.gem"].invoke
       end
     end
   end
```

__Compatibility__

[...]

PR #2073, opened 2020-09-07
Darwin native gems

flavorjones:

Proposed work for #2063 -- precompiled native gems for OSX/Darwin.

Needs:
- [x] implement verify_dll for the darwin .bundle

No functional changes -- only developer changes in how we build gems.


tasks/cross-ruby.rb
```diff
@@ -92,78 +125,117 @@ def dlls
         end),
         "libc.so.6",
       ]
+    when DARWIN_PLATFORM_REGEX
+      [
+        "/usr/lib/libSystem.B.dylib",
+        "/usr/lib/liblzma.5.dylib",
+        "/usr/lib/libobjc.A.dylib",
```
flavorjones, line 132: @larskanis @tenderlove I could use a sanity check on these remaining references. This seems reasonable to me. I'll probably just ship an RC and see if we get feedback, but wanted to highlight that I made a decision with incomplete information here.

flavorjones:
I'm going to merge this so I can ship an RC and get feedback. Would still love to hear anybody's feedback about this!

PR #2087, opened 2020-09-28
Reduce allocations from hack in document_fragment

ashmaroli:

Reduce object allocations from a hack in the constructor of `Nokogiri::HTML::DocumentFragment`.

Nokogiri 1.11 requires at least Ruby 2.4.0. Therefore, we can safely use `Regexp#match?` instead of `String#=~`.
The advantages are:
- `Regexp#match?` doesn't allocate `MatchData` objects.
- Modifying the regex itself allows not using `String#strip` which duplicates the given string.

This change should be covered by existing tests.

No, it doesn't.

### DIFF
--- a/lib/nokogiri/html/document_fragment.rb
+++ b/lib/nokogiri/html/document_fragment.rb
@@ -33,7 +33,7 @@ module Nokogiri
           self.errors = document.errors - preexisting_errors
         else
           # This is a horrible hack, but I don't care
-          if tags.strip =~ /^<body/i
+          if /^\s*?<body/i.match?(tags)
             path = "/html/body"
           else
             path = "/html/body/node()"
--- a/lib/nokogiri/html/document_fragment.rb
+++ b/lib/nokogiri/html/document_fragment.rb
@@ -4,26 +4,26 @@ module Nokogiri
     class DocumentFragment < Nokogiri::XML::DocumentFragment
       ####
       # Create a Nokogiri::XML::DocumentFragment from +tags+, using +encoding+
-      def self.parse tags, encoding = nil
+      def self.parse(tags, encoding = nil)
         doc = HTML::Document.new
 
         encoding ||= if tags.respond_to?(:encoding)
-                       encoding = tags.encoding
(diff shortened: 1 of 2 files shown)


flavorjones:
@ashmaroli Hi! Thanks for submitting this.

Do you have any metrics around the reduced number of allocations? I'm curious how you performed your analysis.

lib/nokogiri/html/document_fragment.rb
```diff
@@ -33,7 +33,7 @@ def initialize document, tags = nil, ctx = nil
           self.errors = document.errors - preexisting_errors
         else
           # This is a horrible hack, but I don't care
-          if tags.strip =~ /^<body/i
+          if /^\s*?<body/i.match?(tags)
```
flavorjones, line 36: Can you tell me why you've inserted `\s*` at the front of this? That's a functional change that I wouldn't expect based on your description; and there aren't any tests provided demonstrating why this is introduced.

flavorjones:
I'll need to look into why the jruby gem tests failed -- I don't think it's related to this PR.

lib/nokogiri/html/document_fragment.rb
ashmaroli: The original code `tags.strip =~ /^<body/i` removes whitespace around string `tags` and tests if `tags` starts with `<body`.
So I thought instead of allocating a sanitized version of `tags` and testing, it'd be better if test was to check if `<body` substring exists preceded by an optional amount of whitespace from the start of string.

In other words, to do away with `tags.strip`...

ashmaroli:
> I'm curious how you performed your analysis

I used `memory_profiler` gem to parse and serialize an HTML source via GitHub Actions on my fork.
I didn't open a pull request with those because I checked in the `*.gemspec` into version control for easy access to latest `nokogiri` gem..

ashmaroli:
> I'll need to look into why the jruby gem tests failed

The reason behind this is `rubocop-ast-0.7.0` requiring `strscan` (native-extension gem).
A `v0.7.1` has been shipped without that requirement. So, restarting just the JRuby pipeline should now pass..

Issue #1603, opened 2017-02-09
resurrect the GC / memory pressure test suite


flavorjones:
The file `test/test_memory_leak.rb` and the teardown behavior controlled by the `NOKOGIRI_GC` env var could be useful.

Consider resurrecting these.

Acceptance criteria:

* suite needs to finish in a reasonable amount of time (currently many are looping literally forever)
* suite needs to be runnable under valgrind (so, not a separate rake task, but possibly controlled by an environment variable for use with the `test:valgrind` task from `hoe-debugging`)
* tests must clearly pass or fail. previously I'd just run a single test and watch the process heap size to see if it increased. It would be great to run under valgrind and use those metrics for something.
* suite should be run in the main concourse pipeline

rubarb666:
[[

flavorjones:
After a little bit of work in #2658 I'm forming some more concrete opinions on this.

I think we could make this a runnable suite if we use:

- the value returned from `MemInfo.rss` (introduced in af99c52)
- and Minitest::BenchSpec (introduced in 46dcc60)

to ensure that memory usage isn't increasing after some number of iterations.

flavorjones:
ZOMG see #3050

PR #2088, opened 2020-09-30
Profile memory usage from processing HTML contents

ashmaroli:

This pull request adds a GitHub Actions Workflow that profiles memory usage from using Nokogiri to parse and serialize HTML content.

The core of the workflow is split into two steps:
- Parsing an entire HTML page with `Nokogiri::HTML::Document` and dumping it back into HTML string.
- Parsing just the `<body>` tag contents with `Nokogiri::HTML::DocumentFragment` and dumping it back into HTML string.

The source material for both steps are from a physical HTML file (in turn sourced from the current state of https://nokogiri.org/).
Having a static fixture ensures consistent sample across multiple runs of the workflow.


flavorjones:
Please ignore the failed CI status -- that was due to my config mistake related to #2089. This PR was green and should be considered green.

ilyazub:
@ashmaroli How to evaluate memory profiling results? I mean, you wouldn't go checking CI logs on every commit and manually compare memory usage between commits.

ashmaroli:
> you wouldn't go checking CI logs on every commit and manually compare memory usage between commits.

@ilyazub Honestly, I was planning on doing just that &mdash; compare manually between commits of interest &mdash; say, the current `master` stats vs stats for a proposed memory optimization.

Theoretically, if there was a way to get the log for an action that ran on `master`, then one could just execute a script to extract the necessary numbers from that log, compare it with the numbers on an action run in a pull request of interest, and post it as a comment to the said pull request (much like how `AppVeyorBot` did on this PR).

flavorjones:
@ashmaroli Thanks for taking the time to put this together, and sorry for not commenting before now.

I think there's value in being able to look at data like this, but (as @ilyazub is implying) I worry that flagging memory leaks requires more rigor that just emitting a log file.

I'd ideally like to embed this approach in a CI test that emits pass/fail status. Can you think of a way to integrate this data better in the testing cycle?

I'm also curious about combining this approach with the long-defunct memory test suite (see https://github.com/sparklemotion/nokogiri/issues/1603) and also incorporate testing with compaction and either valgrind or ASAN. Any thoughts about that?

ilyazub:
For the CI integration
1. The simplest thing is to check memory usage limits: fail if script allocated more than allowed and pass otherwise.
2. The more advanced thing may be to store historical memory usage data in a file (eg., JSON or Yaml) and fail the CI job if the script used more than `prev_run + N%` of memory. To store that file we can use [build artifacts in Concourse CI](https://concourse-ci.org/resources.html) (not sure), separate repository with a single file ([deno.land approach](https://github.com/denoland/benchmark_data)), GitHub gist, etc.
3. Same as the second option, but to use the server to write, read and compare historical data of memory usage (kinda [Lighthouse CI Server approach](https://github.com/GoogleChrome/lighthouse-ci/blob/8b3fde99a22082f9c4ed8ce355f151a794f42fbb/docs/server.md)). Seems to be overkill.

Regarding #1603, we can store historical data for each test in `test/test_memory_leak.rb` (the second approach). The output of each test case could be stored in an array (like for `test_leaking_namespace_node_strings`). But I don't know what to do with endless loops (eg., `test_leak_on_node_replace`).

@flavorjones @ashmaroli What do you think?

ashmaroli:
Thanks for the feedback on this idea, @flavorjones and @ilyazub.
So, the first step is to handle passing data across CI builds. I'm not familiar with Concourse CI, but if suggestion#2 from @ilyazub is implementable, that'd be nice to have.

Regarding increasing the rigor of profiling memory, we may need to look into other options.
To my knowledge, `memory_profiler` only measures Ruby objects. It won't detect allocations from native extensions.
Once we have a satisfactory way to pass data across builds, we could modify the profiling script to parse the artifact data, compare and issue an exit status accordingly.

flavorjones:
It's been a few years, and some things have changed for the better:

- we use `ruby_memcheck` in CI to find memory leaks
- the memory test suite has been resurrected and runs in CI and has a good variety of operations that we analyze (see https://nokogiri.org/CONTRIBUTING.html#cruby-advanced-usage)

Github actions provides the ability to upload and download artifacts with https://github.com/actions/upload-artifact and https://github.com/actions/download-artifact so what's described above is possible.

If someone has the interest to pick this back up again, I think we're in a place where it makes sense. I likely won't get to it this year.

PR #2104, opened 2020-10-26
Add explicit Racc depenency.

### LINKED ISSUE #1988: Explicit dependency on Racc
Since Nokogiri depends on Racc in runtime, it would be nice if the dependency was explicitly specified. I know that Racc is part of Ruby StdLib as default gem, OTOH one day Racc will release incompatible version and Nokogiri won't be usable with certain older versions of Ruby without older gemified version of Racc.

voxik:
Nokogiri always have had dependency on Racc, but it was never stated explicitly. This is possible issue for several reasons:

1) There is no way RubyGems/Bundler could avoid loading wrong version of Racc if other package in application specifies different version of Racc then Nokogiri needs.
2) If there is released incompatible Racc 2.x, all old versions of Nokogiri will be broken just because it does not specify its dependencies properly.
3) If Ruby decides to drop Racc from StdLib, all Nokogiri versions will be broken (unnoticeably). Or Ruby will be forced to carry Racc around around although it is possibly obsolete.

Fixes #1988

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@ This release ends support for:
 
 #### Gems
 
+* Explicitly add racc as a runtime dependency. [[#1988](https://github.com/sparklemotion/nokogiri/issues/1988)] (Thanks, [@voxik](https://github.com/voxik)!)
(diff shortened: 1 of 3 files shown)


flavorjones:
I've rebased and added a CHANGELOG commit. When this goes green, I'll merge and this will be in the next v1.11.0 release candidate.

flavorjones:
And thank you for the PR!

PR #2111, opened 2020-11-12
Pathname special case on parse fixes #2110, #1821

### LINKED ISSUE #2110: Pathname objects are treated as IO and corrupt input on parse
I just ran into a situation where `Pathname` objects get treated as `IO` instead of a file path.

The problem is that a `Pathname` object will be passed to [`io_read_callback`](https://github.com/sparklemotion/nokogiri/blob/104a3459cb00546dc06661b8c97c9109b59195a2/ext/nokogiri/xml_io.c#L14) where `read` is called on it, but since `Pathname` objects have no state, the first ~4000 (4096?) bytes of the file get read in over and over until silently kiboshed by the parser (assuming default flags).

The solution would be to put a branch in the parser logic to pull out `Pathname` objects for special treatment, either by flattening them to locations or explicitly opening them as file handles. This is a straightforward fix I am offering to do; I am registering this issue beforehand to gauge interest.

doriantaylor:

The mangled behaviour of `Pathname` objects; closing #2110 and #1821.

I hope so? The logic on `Nokogiri.parse` is a little funky but it seemed a little funky to begin with.

From what I can tell this change operates above the underlying implementation.

### DIFF
--- a/lib/nokogiri/html/document.rb
+++ b/lib/nokogiri/html/document.rb
@@ -1,4 +1,8 @@
 # frozen_string_literal: true
+
+# for the constant; pathname is standard lib
+require 'pathname'
+
 module Nokogiri
   module HTML
     class Document < Nokogiri::XML::Document
@@ -165,6 +169,10 @@ module Nokogiri
           options = Nokogiri::XML::ParseOptions.new(options) if Integer === options
           # Give the options to the user
           yield options if block_given?
+
+          # coerce to binary filehandle if this is a Pathname object
+          string_or_io = string_or_io.expand_path.open('rb') if
+            string_or_io.is_a? Pathname
 
           if string_or_io.respond_to?(:encoding)
             unless string_or_io.encoding.name == "ASCII-8BIT"
--- a/lib/nokogiri/html/document.rb
+++ b/lib/nokogiri/html/document.rb
@@ -164,10 +164,11 @@ module Nokogiri
         # Nokogiri::XML::ParseOptions::RECOVER.  See the constants in
         # Nokogiri::XML::ParseOptions.
         def parse string_or_io, url = nil, encoding = nil, options = XML::ParseOptions::DEFAULT_HTML
-
           options = Nokogiri::XML::ParseOptions.new(options) if Integer === options
-          # Give the options to the user
+
           yield options if block_given?
+
+          url ||= string_or_io.respond_to?(:path) ? string_or_io.path : nil
 
           if string_or_io.respond_to?(:encoding)
             unless string_or_io.encoding.name == "ASCII-8BIT"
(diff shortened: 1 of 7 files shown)


doriantaylor:
lol code climate you are dumb

doriantaylor:
> **What problem is this PR intended to solve?**
> 
> The mangled behaviour of `Pathname` objects; closing #2110 and #1821.
> 
> **Have you included adequate test coverage?**
> 
> I hope so? The logic on `Nokogiri.parse` is a little funky but it seemed a little funky to begin with.
> 
> **Does this change affect the C or the Java implementations?**
> 
> From what I can tell this change operates above the underlying implementation.

flavorjones:
Don't worry about Code Climate, I marked it as WONTFIX. Looking into the other build failure now.

flavorjones:
The failing test is a flake, I kicked it and it should go green shortly. Looking at the proposed changes now.

doriantaylor:
I'm kinda lukewarm on having three distinct sites where `is_a? Pathname` is tested but I fear consolidating it would entail rethinking the overall parser dispatch logic. That said I can probably get rid of the one in `nokogiri.rb`.

flavorjones:
@doriantaylor I agree, and the reason I haven't moved yet on my promise in #1821 to write my own PR is because I intended to overhaul the parsing logic and try to consolidate it into fewer places; and I haven't had the time; and it's been a low priority because of the ease of the workaround.

I'm cleaning up a couple of issues and will force-push to this branch in a bit along with an explanation of what I'm changing. Hope that's OK, and I'll keep your name on the commits.

lib/nokogiri/html/document.rb
```diff
@@ -174,6 +177,12 @@ def parse string_or_io, url = nil, encoding = nil, options = XML::ParseOptions::
 
           if string_or_io.respond_to?(:read)
             url ||= string_or_io.respond_to?(:path) ? string_or_io.path : nil
+
+            if string_or_io.is_a?(Pathname)
+              # resolve the Pathname to the file and open it as an IO object, see #2110
+              string_or_io = string_or_io.expand_path.open
```
flavorjones, line 183: @tenderlove can I get your eyes on this? I'm pretty sure `File` objects autoclose when they are GCed (and I tested that), but figured it was worth asking explicitly.

doriantaylor: (My inclination was to throw `expand_path` in there because you can have legitimate stuff like `Pathname('~/foo.xml')` and the overhead is minimal.)

doriantaylor: oh btw putting the `url` assignment _after_ the `Pathname` test was deliberate because `Pathname` does not respond to `path` but `Pathname#open` will produce a `File` which _will_ respond to `path`.

flavorjones: @doriantaylor I don't have any concerns about `#expand_path`, I just want to confirm with @tenderlove that the `File` object we're creating will get cleaned up and file descriptors will get closed during object finalization.

Your `url` comment makes sense -- what confused me was that after your initial patch set (which I've squashed), it was done one way in HTML::Document and the other way in XML::Document. I'll add some test coverage to make sure it's working the way we expect.

doriantaylor:
Nice! Thanks for merging so promptly!

Issue #2123, opened 2020-12-01
enable shipping native gems that support ruby 3.0.0-preview1


flavorjones:
I'd like the next v1.11.0 release candidate gem to have support for ruby 3.0.0-preview1 (or whatever the current RC is at that time) so early adopters can kick the tires.

I spiked on this a bit and a `rake-compiler/rake-compiler-dock` patch that seemed to get me somewhere looked like:

```patch
diff --git a/Dockerfile.mri.erb b/Dockerfile.mri.erb
index 87710bb..6d4504f 100644
--- a/Dockerfile.mri.erb
+++ b/Dockerfile.mri.erb
@@ -70,8 +70,8 @@ USER rvm
 
 COPY build/patches2 /home/rvm/patches/
 # Patch rake-compiler to build and install static libraries for Linux rubies
-RUN cd /usr/local/rvm/gems/ruby-2.5.7/gems/rake-compiler-1.1.0 && \
-    ( git apply /home/rvm/patches/rake-compiler-1.1.0/*.patch || true )
+RUN cd /usr/local/rvm/gems/ruby-2.5.7/gems/rake-compiler-1.1.1 && \
+    ( git apply /home/rvm/patches/rake-compiler-1.1.1/*.patch || true )
 RUN cd /usr/local/rvm/gems/ruby-2.5.7/gems/hoe-3.20.0 && \
     ( git apply /home/rvm/patches/hoe-3.20.0/*.patch || true )
 
@@ -89,7 +89,7 @@ RUN curl -SL http://cache.ruby-lang.org/pub/ruby/2.7/ruby-2.7.0.tar.xz | tar -xJ
 USER rvm
 
 
-ENV XRUBIES 2.7.0:2.6.0:2.5.0:2.4.0:2.3.0:2.2.2
+ENV XRUBIES 3.0.0-preview1:2.7.0
 
 # Build all xruby versions in parallel
 # Then cleanup all build artifacts
```

So the things left to do are:

- [x] test that ruby 3 works with an image built the above way
- [x] generate unofficial docker images and push them to dockerhub `flavorjones/nokogiri-test` tagged appropriately
- [x] update CI pipelines and get to green
- [x] build the gems!

flavorjones:
Tagging @davidmisshula because he paired with me on the spike above! Thank you!

flavorjones:
Closed by #2126

PR #2126, opened 2020-12-03
add native gem support for ruby 3

flavorjones:

See #2123, we want to ship a release candidate of native gems that support the current Ruby 3.0.0-preview1 and potentially the upcoming Ruby 3.0.0 final release.

Yes, additional CI task has been created to ensure that the native gem builds and packages a `nokogiri.so` that supports ruby 3.0, but also exercises that the gem can be installed and runs correctly.

This change only impacts the build and packaging of native gems.

### DIFF
--- a/.cross_rubies
+++ b/.cross_rubies
@@ -1,3 +1,4 @@
+3.0.0:x86_64-darwin
 2.7.0:i686-w64-mingw32
 2.7.0:x86_64-w64-mingw32
 2.7.0:i686-linux-gnu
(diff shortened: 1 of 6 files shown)


flavorjones:
Note that background work I did here to build the `rake-compiler-dock` images with Ruby 3.0.0-preview1 is preserved in amber at https://github.com/rake-compiler/rake-compiler-dock/pull/40, and those images have been pushed to my dockerhub account, e.g. https://hub.docker.com/r/flavorjones/rake-compiler-dock-mri-x86_64-linux

Issue #2079, opened 2020-09-08
Native gems: create releases for additional darwin platforms


flavorjones:
Nokogiri v1.11.0.rc3 only supports OSX `x86_64-darwin19` platform.

At least one commented at #2075 has asked about `x86_64-darwin18`. If that's a common platform we should consider shipping that, as well.

I'll be honest and say that my inability to virtualize OSX is a real challenge here. I have a work laptop that's darwin19, and I have an 2009 mac mini that might have darwin18 on it (which [looks like it's x86_64](http://cpuboss.com/cpu/Intel-Core2-Duo-P7350)).

flavorjones:
Note that @drbrain says: "Travis has older Xcode for linking on older Darwin, I use it for building work stuff for MacOS" at https://twitter.com/drbrain/status/1303369620099751936

drbrain:
Here are the [available macOS versions on Travis-CI](https://docs.travis-ci.com/user/reference/osx/#macos-version)

flavorjones:
It looks like this might be helpful, too: https://github.com/tpoechtrager/osxcross which is a cross-compiler toolchain for OSX.

flavorjones:
@luislavena I'm going to use osxcross to experiment with creating a rake-compiler-dock-compatible image for OSX!

flavorjones:
Example of how this was used in the past in a docker container: https://github.com/hone/mruby-cli-docker/blob/master/Dockerfile which is used by the https://github.com/hone/mruby-cli project.

flavorjones:
Aaaand a more complete Dockerfile using osxcross at https://github.com/appPlant/docker-images/blob/mruby-cli/Dockerfile

flavorjones:
Did some poking around this morning. I noticed that the `grpc` gem [only ships a `universal-darwin` gem](https://rubygems.org/gems/grpc/versions), and that this Just Works™ on my mac running either ruby `universal.x86_64-darwin19` or `x86-64-darwin19`. This works because of [this logic in `Gem::Platform`](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/platform.rb#L163-L175).

**Lesson**: we don't need to ship individual gem files for each version of darwin.

**Next step**: let's take a look at how grpc is built. Optionally consider setting up osxcross in a linux container.

flavorjones:
@odlp - I've created a `universal-darwin` gem and put it in a public storage bucket here:

[...]

Issue #25, opened 2009-04-24
Add documentation to Nokogiri::XML::SAX::Parser


tenderlove:
Add documentation to Nokogiri::XML::SAX::Parser

PR #2142, opened 2020-12-22
WIP: Cross compile for x86_64-darwin and arm64-darwin per rake-compiler-dock

larskanis:
This requires latest rake-compiler-dock from master branch.

Fat binary gems for MacOS are currently built on MacOS native. This complicates the build and release cycle. It is desirable to build all binary gems per cross compiler.

Related to #2079

No CI, but I tested the resulting `x86_64-darwin` gem on MacOS-10.14.4 with both the system ruby-2.3.7 (after extending the `required_ruby_version` of nokogiri) and ruby-2.7.2 installed per rvm. The Mac is running on https://github.com/foxlet/macOS-Simple-KVM .

I'm unable to execute the `arm64-darwin` on any computer. I just did some static inspection of the built `nokogiri.bundle` file.

C only.


flavorjones:
@larskanis This is exciting! Thanks for doing this work. I'll spend some time tomorrow trying it out, and I'll ask some questions inline.

larskanis:
I played a bit with universal libraries to bundle x86_64+arm64 code into one package, but that didn't work. IMHO these universal builds require a lot of additional work compared to a clear separation of CPU architectures.

flavorjones:
I'm kicking the tires today and everything looks good, including the darwin x86_64 gem. I'd like to ship an rc4 in the next day or so so we can get feedback on the arm64 build in particular in time for a v1.11.0 final release by year's end.

@larskanis Are you planning to release rake-compiler-dock v1.1.0 soon so I can merge this?

larskanis:
> Are you planning to release rake-compiler-dock v1.1.0 soon so I can merge this?

Yes, I'm planning to release rake-compiler-dock-1.1.0 soon after ruby-3.0 and before end of the year.

flavorjones:
I've pushed a few related commits. I'll likely keep rebasing this to master HEAD so that I can cut and test release candidates. Sorry for the inconvenience.

larskanis:
IMHO rake-compiler-dock-1.1.0 is ready to be released. I switched to [manylinux](https://github.com/pypa/manylinux/tree/manylinux2014) for the linux cross builds as sugested [here](https://github.com/rake-compiler/rake-compiler-dock/issues/33). I'm not fully convinced about this change, but I think the advantages outweigh the disadvantages. Is this OK for you?

I would like to add the CrossRuby class of Nokogiri to rake-compiler-dock (in a somewhat more generic way). It helps to build and verify other gems similar to nokogiri. Not in 1.1.0, but in a later release.

flavorjones:
@larskanis Thanks for asking - I think that's fine, I've built those images locally and am trying the builds out tonight.

If you'd like to re-use the CrossRuby class, I'd like to suggest changing the file format to be something more easily parseable (like YAML, for instance) because I end up reading it in a few different places (e.g., where I generate the CI pipelines). Maybe we can open up an issue for future discussion? I'm happy to help.

flavorjones:
I used a rebased version of this branch to ship v1.11.0.rc4 just now, and things look like they're going OK! See https://github.com/sparklemotion/nokogiri/issues/2149#issuecomment-752161818

larskanis:
I released rake-compiler-dock-1.1.0 and all the images some hours ago. I think this PR can be merged now.

flavorjones:
OK, I've rebased onto master and fixed some of my commits; waiting for this to go green and I'll ship it.

flavorjones:
@larskanis it looks like the move to centos may have introduced an issue running on musl-based systems, see [cruby-native-gem-test-32bit #25.1 - Concourse](https://ci.nokogiri.org/teams/nokogiri-core/pipelines/nokogiri-pr/jobs/cruby-native-gem-test-32bit/builds/25.1#L5fc5b11c:212). I need to do a bit more digging to be sure and reproduce locally.

larskanis:
> it looks like the move to centos may have introduced an issue running on musl-based systems

Oh, that's not good. And it reminds me that I should have done this musl test locally before release. It's too late for today, but I can check this tomorrow.

flavorjones:
@larskanis I've pushed an additional commit that makes extconf.rb support a `--prevent-strip` argument so that I could get see what's going on in the debugger.

After running the test suite about a dozen times and examining the stack walkback for each, the crashes are happening while libxml2 is trying to call either `isnan` or `isinf` (from `xpath.c` functions `xmlXPathIsNaN` or `xmlXPathIsInf`).

Looking at the shared object generated in the centos container, we can see these end up being method calls:

```
bash-5.0# nm /usr/local/bundle/gems/nokogiri-1.11.0.rc4-x86_64-linux/lib/nokogiri/3.0/nokogiri.so | egrep "isnan|isinf"
                 U __isinf@@GLIBC_2.2.5
                 U __isnan@@GLIBC_2.2.5
```

but in a shared object generated in a debian-based container, there are no such symbols, undefined or other. I _think_ this is because it's defined as a macro on musl and debian?

I feel like we could work around this if necessary by providing local versions, but I'm curious if you can see an easier way to resolve this issue.

flavorjones:
I've pushed a commit that patches libxml2 to avoid calling `isinf` or `isnan` which will let me ship this release. I'd be happy to figure out a more permanent fix, but let's not block on it. I'll open a rake-compiler-dock issue for this.

flavorjones:
Merged! Thanks again, @larskanis!

PR #2154, opened 2020-12-30
CHANGELOG: Fix Markdown formatting

olleolleolle:

This makes the copy-and-paste section more pasteable.

No.


flavorjones:
Right on! Thanks for catching that and sending a PR.

PR #2162, opened 2021-01-04
ci: switch to the ruby/setup-ruby action

parndt:

This switches the setup-ruby action to the one created by the
[ruby organisation](https://github.com/ruby/setup-ruby) which has more
advanced features than the one already in use.

Most notably, this handles installing Bundler, bundling, and caching the
resulting bundle.

Yes, which is to say it hasn't changed any code or tests, just how they
are run.

No, just how CI runs, and you can see the results on this PR 😄.

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -12,26 +12,15 @@ jobs:
     runs-on: macos-latest
     steps:
       - uses: actions/checkout@master
-      - uses: actions/setup-ruby@v1
-        with:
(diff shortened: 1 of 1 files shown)


flavorjones:
@parndt Sweet! Thanks very much for showing us how to do this. Merging!

parndt:
@flavorjones you're very welcome; thanks for all the work on Nokogiri! Happy to have helped. 🎉

PR #2186, opened 2021-02-02
Allow NodeSet to contain Nodes from multiple Documents (fix for #1952)

### LINKED ISSUE #1952: segfault in node_set.rb

flavorjones:

fix: allow NodeSets to contain Nodes from multiple Documents

Specifically, this means we've added a mark callback for the NodeSet which marks each of the contained objects.

Previously we skipped explicitly marking the contained objects due to the assumption that all the nodes would be from the same document as the NodeSet itself.

Fixes #1952.

Yes, additional tests created.

This change allows the C implementation to do what the Java implementation has always allowed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 
 ### Fixed
 
+* [CRuby] `NodeSet` may now safely contain `Node` objects from multiple documents. Previously the GC lifecycle of the parent `Document` objects could lead to contained nodes being GCed while still in scope. [[#1952](https://github.com/sparklemotion/nokogiri/issues/1952)]
 * [CRuby] `{XML,HTML}::Document.parse` now invokes `#initialize` exactly once. Previously `#initialize` was invoked twice on each object.
 * [JRuby] `{XML,HTML}::Document.parse` now invokes `#initialize` exactly once. Previously `#initialize` was not called, which was a problem for subclassing such as done by `Loofah`.
 
--- a/ext/nokogiri/xml_namespace.c
+++ b/ext/nokogiri/xml_namespace.c
@@ -1,5 +1,27 @@
 #include <xml_namespace.h>
 
+/*
+ *  The lifecycle of a Namespace node is more complicated than other Nodes, for two reasons:
+ *
+ *  1. the underlying C structure has a different layout than all the other node structs, with the
+ *     `_private` member where we store a pointer to Ruby object data not being in first position.
+ *  2. xmlNs structures returned in an xmlNodeset from an XPath query are copies of the document's
+ *     namespaces, and so do not share the same memory lifecycle as everything else in a document.
+ *
+ *  As a result of 1, you may see special handling of XML_NAMESPACE_DECL node types throughout the
+ *  Nokogiri C code, though I intend to wrap up that logic in ruby_object_{get,set} functions
+ *  shortly.
+ *
+ *  As a result of 2, you will see we have special handling in this file and in xml_node_set.c to
+ *  carefully manage the memory lifecycle of xmlNs structs to match the Ruby object's GC
+ *  lifecycle. In xml_node_set.c we have local versions of xmlXPathNodeSetDel() and
+ *  xmlXPathFreeNodeSet() that avoid freeing xmlNs structs in the node set. In this file, we decide
+ *  whether or not to call dealloc_namespace() depending on whether the xmlNs struct appears to be
+ *  in an xmlNodeSet (and thus the result of an XPath query) or not.
+ *
+ *  Yes, this is madness.
+ */
+
 VALUE cNokogiriXmlNamespace ;
 
 static void dealloc_namespace(xmlNsPtr ns)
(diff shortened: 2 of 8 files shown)


flavorjones:
The failing test at https://ci.nokogiri.org/builds/14968#L60108339:410 looks legit, and so I'll want to dig in a bit more to understand that segfault before I'm happy with this change.

flavorjones:
OK, I understand why this segfaulted and have a fix, I only lack the time to polish and push it.

flavorjones:
Note to self, next steps:

- [ ] ship the segfault fix to the PR and CI it
- [ ] can we remove `@namespace_cache` from `NodeSet` now that we have a real mark function?
    - [ ] make sure that we have a GC test for namespaces and that we run it ...

I'm unhappy with the complexity of the code we have sprinkled throughout the codebase to manage ruby object lifecycle, wrapping of nodes, and marking of nodes. I'd like to keep going on this branch to address some of those concerns:

- [ ] assert that namespace nodes are not passed into `Nokogiri_wrap_xml_node` to set a baseline
- [ ] for all calls to `Nokogiri_wrap_xml_node` that pass a class, rewrite it to pass Qnil and then assert on the returned object's class.
    - if that doesn't explode, then we can omit that first `klass` argument
- [ ] inside `Nokogiri_wrap_xml_namespace` assert that the `context` struct member matches the passed document
    - if that doesn't explode, then we can omit that first `doc` argument
- [ ] now we can combine those two functions?
- [ ] explore: can we also perform the `nokogiri_root_node` call within that combined function? when wouldn't we want to?
- [ ] explore: what does it look like to use `ruby_object_{get,set}` and `ruby_class_get` everywhere?

Issue #2191, opened 2021-02-10
support `--disable-static` option on TruffleRuby


flavorjones:
See  for more context.

TruffleRuby, by default, sets the `--use-system-libraries` option (via [these lines in mkmf.rb](https://github.com/oracle/truffleruby/blob/d0c380361d65fd9cc89f98efdf7c7f956f073720/lib/mri/mkmf.rb#L52-L54)). Because of how Nokogiri's `extconf.rb` is currently implemented, this setting takes precedence over other options like `--disable-static` which would allow users to compile the packaged, patched versions of libxml2/libxslt and use them as shared libraries (instead of the system libraries).

I've proposed, in https://github.com/Shopify/oracle-truffleruby-collab/issues/11, to change Nokogiri's extconf.rb to essentially follow this logic:

```ruby
static_p = enable_config("static", true) # defaults to true, overridden to false by `--disable-static`
if defined?(::TruffleRuby)
  if static_p
    # use system libraries - sane default to make installation easy
  else
    # use packaged libraries, built as shared objects - use the patched packaged versions, which might be hard for some systems to compile
  end
end
```

Essentially, this means that the environment variable `NOKOGIRI_USE_SYSTEM_LIBRARIES` would be presumed to be the default behavior on TruffleRuby; but allow it to be overridden.

eregon:
I've made a small change locally to allow overriding `NOKOGIRI_USE_SYSTEM_LIBRARIES` with command-line arguments (PR incoming).

With that, it's possible to install nokogiri in all 3 variants. I'm using the latest TruffleRuby nightly build.

Current default, using system libraries:
```
$ time gem install --local pkg/nokogiri-1.11.1.gem    
Building native extensions. This could take a while...
Successfully installed nokogiri-1.11.1
1 gem installed
 34.24s user 2.19s system 424% cpu 8.579 total
```

Compiling the vendored libxml2/libxslt as shared libraries:
```
$ time gem install --local pkg/nokogiri-1.11.1.gem -- --use-system-libraries=false --disable-static 
Building native extensions with: '--use-system-libraries=false --disable-static'
This could take a while...
Successfully installed nokogiri-1.11.1
1 gem installed
99.45s user 12.71s system 281% cpu 39.789 total
```

Compiling the vendored libxml2/libxslt as static libraries (the default on CRuby):
```
$ time gem install --local pkg/nokogiri-1.11.1.gem -- --use-system-libraries=false                  
[...]

PR #2193, opened 2021-02-17
Allow overriding env var with --use-system-libraries=false and default to --disable-static on TruffleRuby

eregon:
See https://github.com/sparklemotion/nokogiri/issues/2191

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 * Reduce the number of object allocations needed when parsing an HTML::DocumentFragment. [[#2087](https://github.com/sparklemotion/nokogiri/issues/2087)] (Thanks, [@ashmaroli](https://github.com/ashmaroli)!)
 * [JRuby] Update the algorithm used to calculate `Node#line` to be wrong less-often. The underlying parser, Xerces, does not track line numbers, and so we've always used a hacky solution for this method. [[#1223](https://github.com/sparklemotion/nokogiri/issues/1223)]
 * Introduce `--enable-system-libraries` and `--disable-system-libraries` flags to `extconf.rb`. These flags provide the same functionality as `--use-system-libraries` and the `NOKOGIRI_USE_SYSTEM_LIBRARIES` environment variable, but are more idiomatic. [[#2193](https://github.com/sparklemotion/nokogiri/issues/2193)] (Thanks, [@eregon](https://github.com/eregon)!)
+* [TruffleRuby] `--disable-static` is used by default on TruffleRuby, when `--disable-system-libraries` is set, which is more flexible and compiles faster, see [#2191](https://github.com/sparklemotion/nokogiri/issues/2191#issuecomment-780724627). [[#2193](https://github.com/sparklemotion/nokogiri/issues/2193)] (Thanks, [@eregon](https://github.com/eregon)!)
 
 
 ## 1.11.1 / 2021-01-06
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,7 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 
 * Reduce the number of object allocations needed when parsing an HTML::DocumentFragment. [[#2087](https://github.com/sparklemotion/nokogiri/issues/2087)] (Thanks, [@ashmaroli](https://github.com/ashmaroli)!)
 * [JRuby] Update the algorithm used to calculate `Node#line` to be wrong less-often. The underlying parser, Xerces, does not track line numbers, and so we've always used a hacky solution for this method. [[#1223](https://github.com/sparklemotion/nokogiri/issues/1223)]
+* Introduce `--enable-system-libraries` and `--disable-system-libraries` flags to `extconf.rb`. These flags provide the same functionality as `--use-system-libraries` and the `NOKOGIRI_USE_SYSTEM_LIBRARIES` environment variable, but are more idiomatic. [[#2193](https://github.com/sparklemotion/nokogiri/issues/2193)] (Thanks, [@eregon](https://github.com/eregon)!)
 
 
 ## 1.11.1 / 2021-01-06
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -149,6 +149,18 @@ HELP
 #
 #  utility functions
 #
+def config_clean?
+  enable_config('clean', true)
+end
+
+def config_static?
+  enable_config("static", true)
+end
+
+def config_cross_build?
+  enable_config("cross-build")
+end
+
 def config_system_libraries?
   enable_config("system-libraries", ENV.key?("NOKOGIRI_USE_SYSTEM_LIBRARIES")) do |_, default|
     arg_config('--use-system-libraries', default)
@@ -524,7 +536,7 @@ def do_clean
       FileUtils.rm_rf(dir, verbose: true)
     end
 
-    if enable_config('static')
+    if config_static?
       # ports installation can be safely removed if statically linked.
       FileUtils.rm_rf(root + 'ports', verbose: true)
     else
@@ -618,10 +630,10 @@ if config_system_libraries?
 else
   message "Building nokogiri using packaged libraries.\n"
 
-  static_p = enable_config("static", true)
+  static_p = config_static?
   message "Static linking is #{static_p ? 'enabled' : 'disabled'}.\n"
 
-  cross_build_p = enable_config("cross-build")
+  cross_build_p = config_cross_build?
   message "Cross build is #{cross_build_p ? 'enabled' : 'disabled'}.\n"
 
   require 'yaml'
@@ -876,7 +888,7 @@ end
 
 create_makefile('nokogiri/nokogiri')
 
-if enable_config('clean', true)
+if config_clean?
   # Do not clean if run in a development work tree.
   File.open('Makefile', 'at') do |mk|
     mk.print(<<~EOF)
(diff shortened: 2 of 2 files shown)


eregon:
@flavorjones Could you review and integrate this? It'd be awesome to have a release with it, as then it will be possible to install nokogiri with or without bitcode on TruffleRuby (https://github.com/sparklemotion/nokogiri/issues/2191#issuecomment-780724627).

flavorjones:
@eregon Absolutely, sorry for not ACKing earlier, it's been a busy week. Will absolutely look and try to have this merged for a patch release in v1.11.x.

flavorjones:
@eregon I've finally got some time to take a look, and I rewrote this a bit to clean up the `extconf.rb` file and how we do configuration (keeping your co-author credit). Notably, I've opted for `--enable-system-libraries` and `--disable-system-libraries` rather than checking for a specific value of the `NOKOGIRI_USE_SYSTEM_LIBRARIES` environment variable.

So now this works OK for me when I have "static" linking on (`--enable-static`) but I cannot get it to work when I am using "shared" linking (`--disable-static`), which is the default after this patch is applied. I suspect there's something wrong with my toolchain, but I can't figure out what it is.

You can reproduce what I'm seeing using the `flavorjones/truffleruby:nightly` docker image, which at this moment is `truffleruby 21.1.0-dev-df94e415`.

Here's what I'm running and what I see:

```
# bundle exec rake compile -- --disable-system-libraries
mkdir -p tmp/x86_64-linux/nokogiri/2.7.2
cd tmp/x86_64-linux/nokogiri/2.7.2
/usr/local/bin/truffleruby -I. ../../../../ext/nokogiri/extconf.rb  -- --disable-system-libraries
checking for whether -g is accepted as CFLAGS... yes
checking for whether -Winline is accepted as CFLAGS... yes
checking for whether -Wmissing-noreturn is accepted as CFLAGS... yes
Building nokogiri using packaged libraries.
Static linking is disabled.
Cross build is disabled.
checking for gzdopen() in -lz... yes
Using mini_portile version 2.5.0
checking for iconv... yes
---------- IMPORTANT NOTICE ----------
Building Nokogiri with a packaged version of libxml2-2.9.10.
Configuration options: --host\=x86_64-unknown-linux-gnu --enable-static --disable-shared --libdir\=/nokogiri/ports/x86_64-unknown-linux-gnu/libxml2/2.9.10/lib --with-iconv\=yes --without-python --without-readline --with-c14n --with-debug --with-threads --enable-shared --disable-static CFLAGS\=-O2\ -U_FORTIFY_SOURCE\ -g
The following patches are being applied:
  - 0001-Revert-Do-not-URI-escape-in-server-side-includes.patch
  - 0002-Remove-script-macro-support.patch
  - 0003-Update-entities-to-remove-handling-of-ssi.patch
  - 0004-libxml2.la-is-in-top_builddir.patch
  - 0005-Fix-infinite-loop-in-xmlStringLenDecodeEntities.patch
  - 0006-htmlParseComment-treat-as-if-it-closed-the-comment.patch
  - 0007-use-new-htmlParseLookupCommentEnd-to-find-comment-en.patch
  - 0008-use-glibc-strlen.patch
  - 0009-avoid-isnan-isinf.patch
  - 0010-parser.c-shrink-the-input-buffer-when-appropriate.patch

The Nokogiri maintainers intend to provide timely security updates, but if
this is a concern for you and want to use your OS/distro system library
instead, then abort this installation process and install nokogiri as
instructed at:

  https://nokogiri.org/tutorials/installing_nokogiri.html#installing-using-standard-system-libraries

Note, however, that nokogiri cannot guarantee compatiblity with every
version of libxml2 that may be provided by OS/package vendors.

Extracting libxml2-2.9.10.tar.gz into tmp/x86_64-unknown-linux-gnu/ports/libxml2/2.9.10... OK
Running git apply with /nokogiri/patches/libxml2/0001-Revert-Do-not-URI-escape-in-server-side-includes.patch... OK
Running git apply with /nokogiri/patches/libxml2/0002-Remove-script-macro-support.patch... OK
Running git apply with /nokogiri/patches/libxml2/0003-Update-entities-to-remove-handling-of-ssi.patch... OK
Running git apply with /nokogiri/patches/libxml2/0004-libxml2.la-is-in-top_builddir.patch... OK
Running git apply with /nokogiri/patches/libxml2/0005-Fix-infinite-loop-in-xmlStringLenDecodeEntities.patch... OK
Running git apply with /nokogiri/patches/libxml2/0006-htmlParseComment-treat-as-if-it-closed-the-comment.patch... OK
Running git apply with /nokogiri/patches/libxml2/0007-use-new-htmlParseLookupCommentEnd-to-find-comment-en.patch... OK
Running git apply with /nokogiri/patches/libxml2/0008-use-glibc-strlen.patch... OK
Running git apply with /nokogiri/patches/libxml2/0009-avoid-isnan-isinf.patch... OK
Running git apply with /nokogiri/patches/libxml2/0010-parser.c-shrink-the-input-buffer-when-appropriate.patch... OK
Running 'configure' for libxml2 2.9.10... OK
Running 'compile' for libxml2 2.9.10... ERROR, review '/nokogiri/tmp/x86_64-linux/nokogiri/2.7.2/tmp/x86_64-unknown-linux-gnu/ports/libxml2/2.9.10/compile.log' to see what happened. Last lines are:
========================================================================
  CC       xmlreader.lo
  CC       relaxng.lo
  CC       dict.lo
  CC       SAX2.lo
  CC       xmlwriter.lo
  CC       legacy.lo
  CC       chvalid.lo
  CC       pattern.lo
  CC       xmlsave.lo
  CC       xmlmodule.lo
  CC       schematron.lo
  CC       xzlib.lo
  CCLD     libxml2.la
  CCLD     xmllint
clang-10: error: no such file or directory: './.libs/libxml2.so'
make[2]: *** [Makefile:1152: xmllint] Error 1
make[2]: Leaving directory '/nokogiri/tmp/x86_64-linux/nokogiri/2.7.2/tmp/x86_64-unknown-linux-gnu/ports/libxml2/2.9.10/libxml2-2.9.10'
make[1]: *** [Makefile:1479: all-recursive] Error 1
make[1]: Leaving directory '/nokogiri/tmp/x86_64-linux/nokogiri/2.7.2/tmp/x86_64-unknown-linux-gnu/ports/libxml2/2.9.10/libxml2-2.9.10'
make: *** [Makefile:893: all] Error 2
```

When I look in the directory `/nokogiri/tmp/x86_64-linux/nokogiri/2.7.2/tmp/x86_64-unknown-linux-gnu/ports/libxml2/2.9.10/libxml2-2.9.10/.libs` I see that `libxml2.so` is a symlink to `libxml2.so.2.9.10` which doesn't exist:

```
# ls -lt /nokogiri/tmp/x86_64-linux/nokogiri/2.7.2/tmp/x86_64-unknown-linux-gnu/ports/libxml2/2.9.10/libxml2-2.9.10/.libs
total 9912
lrwxrwxrwx 1 root root      13 Mar  9 03:18 libxml2.la -> ../libxml2.la
-rw-r--r-- 1 root root     975 Mar  9 03:18 libxml2.lai
lrwxrwxrwx 1 root root      17 Mar  9 03:18 libxml2.so -> libxml2.so.2.9.10
lrwxrwxrwx 1 root root      17 Mar  9 03:18 libxml2.so.2 -> libxml2.so.2.9.10
-rw-r--r-- 1 root root   66552 Mar  9 03:18 xzlib.o
...
-rw-r--r-- 1 root root   84512 Mar  9 03:18 entities.o
-rw-r--r-- 1 root root   20260 Mar  9 03:18 SAX.o
```

I'm wondering if we could omit the commit to disable-static from this PR for now, and treat it as a separate concern? Otherwise this seems good to go.

flavorjones:
@eregon Hmm, OK, well since we can't reproduce with a gem, then I'm going to include this in v1.11.2 which should go out today.

eregon:
Actually, installing from a `.gem` with `gem install --local pkg/nokogiri-1.11.1.gem -- --disable-system-libraries` does not work with latest TruffleRuby unfortunately.

It seems due to the fact recently TruffleRuby prepends the toolchain to PATH and libxml2's configure/Makefile not liking it, and in fact disabling that works (as it used to):
`TRUFFLERUBYOPT="--experimental-options --cexts-prepend-toolchain-to-path=false" gem install --local pkg/nokogiri-1.11.1.gem -- --disable-system-libraries`.

`--disable-system-libraries --enable-static` does indeed seem to work even with the toolchain in PATH.

So it seems best to hold off on this a bit longer until we figure out what's going on, or we decide to go for `--enable-static`.
Sorry, I did not envision it would break between opening my PR and now.

flavorjones:
@eregon Thanks for finding the root cause. I'm going to remove the "static" default commit from this PR but pull in the "system libraries" flags to unblock the Shopify folks.

flavorjones:
And I'll create a new PR with that commit on it!

eregon:
@flavorjones Agreed, that sounds safe, useful and lets us try all ways to install nokogiri.

eregon:
I figured it out, the issue was that `ld` is used by libxml2 and the GraalVM LLVM toolchain's `ld` was not properly built in the TruffleRuby standalone. I'll fix that in TruffleRuby, it's trivial. With that fix, `--disable-static` works well.

@flavorjones So I believe it's fine to include the commit to `--disable-static` by default on TruffleRuby as well in the nokogiri release. There are [several advantages](https://github.com/sparklemotion/nokogiri/issues/2191#issuecomment-780724627) about it so it seems worth it.

eregon:
@flavorjones I've cherry-picked the commit having `--disable-static` by default on TruffleRuby, and added an entry to the CHANGELOG about it. It should all be ready now :)

flavorjones:
Just kicked CI, it was stuck.

PR #2195, opened 2021-02-26
ci: specify 30-x64 now that Appveyor supports it

flavorjones:

`appveyor.yml` specifies "head-x64" but "30-x64" is now recognized (https://www.appveyor.com/docs/windows-images-software/#ruby).

### DIFF
--- a/.github/workflows/macos.yml
+++ b/.github/workflows/macos.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - main
   pull_request:
+    types: [opened, synchronize]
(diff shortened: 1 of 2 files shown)


larskanis:
Do you intend to remove Ruby-head tests? I think it's useful to keep it and would rather consider to drop Ruby-2.6 and 2.7 if test duration matters.

flavorjones:
Ah! Good point, I just force-pushed to keep "head" and only add 3.0.

Issue #2207, opened 2021-03-15
Understand whether we can move all of CI to Github Actions


flavorjones:
### Context

Current CI configuration, and the advantages Concourse provides, are explained at https://github.com/sparklemotion/nokogiri/blob/25e7c49c87bda2cb1e2dd15677af11c74503dc04/CONTRIBUTING.md#how-continuous-integration-ci-is-configured

### What problems would we solve by moving to GA?

- Appveyor is slow.
- The Concourse environment is relatively expensive to keep running.
- Knowledge of Concourse is low in the Ruby community.
- Easier integration of the Nokogumbo test suite (see #2204)

### Goals of this issue

I simply want to build confidence that we can go all-in on GA, and remove either Appveyor or Concourse or both; so that I can accept PRs like #2011 without feeling like I'm spreading myself even-thinner.

### Unknowns and risks

- [x] How to use custom containers (e.g., rake-compiler-dock containers and the nokogiri-test containers)
- [x] How to move large payloads between jobs (e.g., between gem-build and gem-install-and-test)
- [x] How will GA perform on memory-intensive and CPU-intensive jobs like valgrind tests?
- [x] Can we cover all the Windows use cases in GA? (MSVC, mingw, 32-bit, 64-bit ... ?)

I've started reading https://docs.github.com/en/actions/learn-github-actions/managing-complex-workflows

### Related conversations / issues / PRs

* https://github.com/sparklemotion/nokogiri/issues/2015#issuecomment-779992560
* [Update Actions CI by MSP-Greg · Pull Request #2011 · sparklemotion/nokogiri](https://github.com/sparklemotion/nokogiri/pull/2011)

---

cc @MSP-Greg

MSP-Greg:
@flavorjones

I sense your frustration here (and in past messages) with CI.  GA is still missing features that exist in Travis & AppVeyor.  Re the issues you've listed:

> How to use custom containers (e.g., rake-compiler-dock containers and the nokogiri-test containers)

I haven't worked much with them.  See https://github.com/ruby/ruby/blob/master/.github/workflows/compilers.yml for one example with a container.

> How to move large payloads between jobs (e.g., between gem-build and gem-install-and-test)

I believe it can be done, but I haven't implemented it in GA.  I have PowerShell scripts to build gems (and test the install) with AppVeyor, but those used one job.  Being able to do so on GA and take advantage of parallel jobs would be useful.

[...]

PR #2226, opened 2021-04-29
ci: check how fast/slow valgrind is on GA

flavorjones:

Let's experiment with a `rake test:valgrind` run on Github Actions. This action is taking ~1.5h on my Concourse instance on GCP lately, and I'm not sure why (it used to only take ~30 minutes).

Related to #2207

### DIFF
--- a/.github/workflows/valgrind.yml
+++ b/.github/workflows/valgrind.yml
@@ -1,0 +1,24 @@
+# this is a work in progress!
+
+name: valgrind
+
+on:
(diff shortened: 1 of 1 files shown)


flavorjones:
The test ran in ~28 minutes total, much faster than the concourse deployment on GCP. That's pretty compelling.

flavorjones:
Closing, will be rolling this into a new PR shortly.

Issue #2241, opened 2021-05-18
[bug] using windows system libxml 2.9.12 segfaults on exit


flavorjones:
**Please describe the bug**

When Appveyor and Github Actions upgraded their system libraries to libxml 2.9.12, the tests started segfaulting on exit.

- Appeyor example: https://ci.appveyor.com/project/flavorjones/nokogiri/builds/39213892/job/0hrn9p1k20adtjsv#L668
- GA example: https://github.com/sparklemotion/nokogiri/runs/2614266657?check_suite_focus=true#step:7:70

Note specifically that the same version of Nokogiri running against system libxml 2.9.10 does not do this; nor does Nokogiri running against packaged libxml 2.9.12.

**Additional context**

There were some changes introduced into libxml 2.9.11/12 related to cleanup of windows DLLs at unload/exit time, which seem likely to be contributing causes here. Relevant links:

- https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/66
  - https://gitlab.gnome.org/GNOME/libxml2/commit/c7c526d6d0f605ed090f8fc1bbede9e439d3185c
- https://gitlab.gnome.org/GNOME/libxml2/-/issues/153
- https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/72
  - https://gitlab.gnome.org/GNOME/libxml2/-/commit/9fa3200cb366c726f7c8ef234282603bb9e8816d
- https://github.com/sparklemotion/nokogiri/issues/2059
  - https://gitlab.gnome.org/GNOME/libxml2/-/commit/956534e02ef280795a187c16f6ac04e107f23c5d

PR #2242, opened 2021-05-18
Remove the default allocator function from Document

tenderlove:
We expect that XML documents will be of type T_DATA as they are backed
by a C extension.  It was possible for people to call `allocate` on
Document and end up with a T_OBJECT.  This commit removes the default
allocator so the only way to get a document object is via the
constructors we've defined.

### DIFF
--- a/ext/nokogiri/xml_document.c
+++ b/ext/nokogiri/xml_document.c
@@ -663,6 +663,8 @@ noko_init_xml_document()
    */
   cNokogiriXmlDocument = rb_define_class_under(mNokogiriXml, "Document", cNokogiriXmlNode);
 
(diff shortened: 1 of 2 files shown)


flavorjones:
Windows tests are failing because of https://github.com/sparklemotion/nokogiri/issues/2241 which I'm working on

flavorjones:
Chatted with @tenderlove and it sounds like we want to put this into draft mode until all the C-struct classes either define allocate or undef it.

flavorjones:
I've rebased which should address the failing windows tests.

flavorjones:
Superseded by #2249

PR #2268, opened 2021-06-16
CI: Add UCRT based test on Appveyor

larskanis:
and use a wider range of ruby versions

I would like to test nokogiri on rubyinstaller-head which has switched to a newer C runtime called UCRT. Unfortuantely https://github.com/MSP-Greg/setup-ruby-pkgs doesn't support UCRT yet, so that I reactivated appveyor x64 build for it. While being over there, I would suggest to use a wider range of ruby versions, so that the latest and newest supported version is checked.

The CI run.

Only C.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -33,12 +33,22 @@ test_script:
 
 environment:
   matrix:
-    - ruby_version: 30
-    - ruby_version: 30
+    - ruby_version: head-x64
+    - ruby_version: head-x64
+      INSTALL_PACKAGES: "mingw-w64-ucrt-x86_64-libxslt"
+      EXTCONF_PARAMS: "--use-system-libraries"
+
+    - ruby_version: head-x86
+    - ruby_version: head-x86
       INSTALL_PACKAGES: "mingw-w64-i686-libxslt"
       EXTCONF_PARAMS: "--use-system-libraries"
 
-    - ruby_version: 27
-    - ruby_version: 27
+    - ruby_version: 25
+    - ruby_version: 25
       INSTALL_PACKAGES: "mingw-w64-i686-libxslt"
       EXTCONF_PARAMS: "--use-system-libraries"
+
+matrix:
+  allow_failures:
+    - ruby_version: head-x64
+    - ruby_version: head-x86


flavorjones:
@larskanis is this why ruby-head on windows is failing at the moment? (example: https://github.com/sparklemotion/nokogiri/runs/2837044316?check_suite_focus=true)

flavorjones:
@larskanis also, we're testing every version of Ruby on windows:

image

What versions were you thinking of when you say " I would suggest to use a wider range of ruby versions"?

flavorjones:
@larskanis One other note, in addition to my comments above: the real value that the appveyor builds are providing right now (given that we're testing windows thoroughly in github actions) is the 32-bit windows builds, which have been removed in this PR. If you'd like to *add* UCRT coverage on ruby-head, that's great. But please restore the 32-bit tests.

Is there a plan to work with @MSP-Greg on UCRT support on Github Actions?

MSP-Greg:
I just opened an [issue](https://github.com/actions/virtual-environments/issues/3597) re adding the ucrt packages to the Actions Windows images, and I'll be looking  at the code in `setup-ruby` and `setup-ruby-packages`.  So, yes, we're working on it...

larskanis:
@flavorjones Sorry for the noise - the CI didn't run without opening a pull request. But now that it run, I noticed the [(failing) test run ](https://github.com/sparklemotion/nokogiri/runs/2837044316?check_suite_focus=true) in the "upstream" github action, that you already mentioned above.

So since there is already a rubyinstaller-head test run in the CI, so this PR can be considered obsolete. The reason why the "upstream" test fails, is that the wrong PATH is set and the wrong packages are installed. The PATH should be
  `C:/msys64/ucrt64/bin/`
instead of
  `C:/msys64/mingw64/bin/`
and the package should be
  `mingw-w64-ucrt-x86_64-libxslt`
instead of
  `mingw-w64-x86_64-libxslt`
.

However both issues should be solved in https://github.com/MSP-Greg/setup-ruby-pkgs .

larskanis:
>  " I would suggest to use a wider range of ruby versions"?

What I meant is, that ruby-2.7 and 3.0 are close together, while ruby-2.4 and ruby-head might trigger more potential issues on the 32 bit x86 platform.

flavorjones:
@MSP-Greg Thanks for opening that issue!

@larskanis Thanks for the additional context! I'll close this, but will also take your advice on adding versions for the 32-bit build (I was being lazy).

MSP-Greg:
@larskanis

It's not quite that simple.  The path manipulation happens in `setup-ruby`, not `setup-ruby-pkgs`.  I think `setup-ruby` can be patched to set the correct path (without a breaking change), and same for patching `setup-ruby-pkgs`.

Currently on Windows, `setup-ruby` always enables the mingw paths for access to the tools.  For mswin and older MSYS Rubies, it  additionally adds the 'paths' for their tools.

To clarify, `setup-ruby-pkgs` is designed to simplify cross-platform package installation and updates.  If that isn't needed or desired, `setup-ruby` is all that is needed.  But, `setup-ruby` also assumes that packages/tools are installed on the Actions image, and currently the `ucrt` packages/tools are not.

Also, `setup-ruby-pkgs` uses the code in `setup-ruby`.  So, if we change the code in `setup-ruby` to set the correct path, and I also change the code in `setup-ruby-pkgs` to install `ucrt` packages with `ruby-head`, it will work for any people using `setup-ruby-pkgs`.

But, if they're only using `setup-ruby`, it will break because none of the packages are installed.

Hence, let's wait and see the response to adding ucrt to the Actions image? Regardless, I'll get the code ready.

MSP-Greg:
@flavorjones & @larskanis

I added Windows ruby-head (the ucrt build) to Actions.

Please see https://github.com/MSP-Greg/nokogiri/actions/runs/944724538

[windows (enable, head)](https://github.com/MSP-Greg/nokogiri/runs/2844491787) compiled with a lot of warnings, failed one test.

[windows (disable, head)](https://github.com/MSP-Greg/nokogiri/runs/2844491896) also compiled with a lot of warnings, but passed all tests.

Feel free to look at the test logs.

As to the code, I've got debug output in it, etc.  It involves patches to both `setup-ruby` and `setup-ruby-pkgs`, and since there are no ucrt packages installed on the Actions image, it is updating a lot more that needed, and it isn't setup to do that on a particular job, it does so for all.

Or, maybe we can wait to see if MSFT/GitHub will be adding the ucrt packages...

flavorjones:
@MSP-Greg Thanks for that! I've posted a branch that will build using those tags at https://github.com/sparklemotion/nokogiri/pull/2272

- [ruby-head (windows, enable)](https://github.com/sparklemotion/nokogiri/runs/2848463493?check_suite_focus=true)
- [ruby-head (windows, disable)](https://github.com/sparklemotion/nokogiri/runs/2848463519?check_suite_focus=true)

PR #2272, opened 2021-06-17
ci: try UCRT for windows ruby-head builds

flavorjones:

The 3.1.0.dev windows ruby installer uses UCRT, see #2268 for context. This gives Nokogiri and @larskanis a feedback loop.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -58,7 +58,6 @@ jobs:
         with:
           path: ports/archives
           key: tarballs-ubuntu-${{hashFiles('**/dependencies.yml')}}
-          restore-keys: tarballs-
       - run: bundle install --local || bundle install
       - run: bundle exec rake compile -- --${{matrix.sys}}-system-libraries
       - run: bundle exec rake test
@@ -80,7 +79,6 @@ jobs:
         with:
           path: ports/archives
           key: tarballs-ubuntu-${{hashFiles('**/dependencies.yml')}}
-          restore-keys: tarballs-
       - run: bundle install --local || bundle install
       - run: bundle exec rake compile -- --${{matrix.sys}}-system-libraries
       - run: bundle exec rake test:valgrind
@@ -136,7 +134,6 @@ jobs:
         with:
           path: ports/archives
           key: tarballs-ubuntu-${{hashFiles('**/dependencies.yml')}}
-          restore-keys: tarballs-
       - run: bundle install --local || bundle install
       - run: bundle exec rake compile -- --${{matrix.sys}}-system-libraries
       - run: bundle exec rake test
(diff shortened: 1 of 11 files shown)


flavorjones:
@larskanis This looks like it might be a genuine failure on UCRT? 

> https://github.com/sparklemotion/nokogiri/runs/2848463493?check_suite_focus=true#step:7:50

larskanis:
The failure happens when calling the XLST function `date:date()`. So the issue resides in libxslt pretty sure. In contrast to MSVCRT, UCRT has non ANSI-C compatible time functions. This already needed a [fix in ffi](https://github.com/ffi/ffi/commit/4924308429dd5535cbd7795ff503315cdddad06a). I guess that libxslt simply disables date+time functions on UCRT. I'll try to fix that.

flavorjones:
@larskanis That's interesting! Looking at `libexslt/date.c` it seems like `HAVE_TIME` is a compile-time macro:

```c
#if defined(HAVE_TIME_H)					\
    && (defined(HAVE_LOCALTIME) || defined(HAVE_LOCALTIME_R))	\
    && (defined(HAVE_GMTIME) || defined(HAVE_GMTIME_R))		\
    && defined(HAVE_TIME)
#define WITH_TIME
#endif
```

I'll also note that it's fine when Nokogiri compiles it (`--disable-system-libraries`), but not when we use the UCRT system libraries (`--enable-system-libraries`), which I don't fully understand.

Anything I can do to help?

flavorjones:
Also: I'd be fine with not fixing this. For example, when using a version of libxml2 that does not have iconv support compiled in, we skip the tests and expose that configuration as part of VERSION_INFO.

flavorjones:
WDYT of the following solution:

- during Nokogiri initialization, call `exsltDateDate(NULL)`
- if the return value is NULL, set a ruby constant like `NOKOGIRI_LIBXSLT_TIME_SUPPORT` to false
- else set it to true
- expose it in `VERSION_INFO`

Then we can either rewrite the failing test to not depend on date/time functionality, and/or conditionally run tests based on the value of the constant.

larskanis:
I opened a PR on Mingw-packages addressing the date/time issue: https://github.com/msys2/MINGW-packages/pull/8957

flavorjones:
@larskanis In the meantime, I have a commit on this PR to detect whether libxslt is built with date-time functionality, and skip the assertion if not.

larskanis:
The latest MSYS2 package `mingw-w64-ucrt-x86_64-libxslt-1.1.34-4` is now in the official repository and it fixes the date/time issue on UCRT64.

flavorjones:
@larskanis yup! looks good in this build: https://github.com/sparklemotion/nokogiri/runs/2873137424?check_suite_focus=true#step:7:35

OMERKHALIDFATHIFATHI:
وت

PR #2277, opened 2021-06-19
improve CI caching of compiled dependencies

flavorjones:

Quite a bit of CI time is spent compiling our packaged dependencies (libxml2, libxslt, libiconv, zlib). For a given platform, we should be able to cache these libraries once compiled.

This PR does two things:

- move libgumbo out of `/ports`
- caches `/ports` for most jobs that were previously caching just `/ports/archives` (which only contains the tarballs)

I'll run CI twice on this branch and compare the timing to previous runs to make sure this is actually effective.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,8 +56,8 @@ jobs:
       - uses: actions/cache@v2
         if: matrix.sys == 'disable'
         with:
-          path: ports/archives
-          key: tarballs-ubuntu-${{hashFiles('**/dependencies.yml')}}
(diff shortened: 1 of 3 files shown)


flavorjones:
Pretty compelling before/after timings for the `rake compile` job:

|              | before | after |
|--------------|--------|-------|
| linux 2.7    | 1m12s  | 18s   |
| linux head   | 1m29s  | 17s   |
| macos head   | 2m10s  | 34s   |
| windows head | 9m47s  | 1m42s |

Issue #2204, opened 2021-03-13
Epic: merge Nokogumbo into Nokogiri


flavorjones:
The maintainers of Nokogiri and Nokogumbo are planning to merge the two gems together so that Nokogiri assimilates Nokogumbo's HTML5 parsing functionality.

This issue is intended to be a "parent" issue which can be followed to understand the plan and how it's progressing, as the work will likely take O(weeks) and will be on a feature branch driven by multiple PRs.

This description will be edited as we go to reflect current state and progress.

## Background

This work has previously been discussed:

* #2064
* https://github.com/rubys/nokogumbo/issues/170

## Goals

Here's what success looks like:

* Nokogiri v1.12 supports HTML5 parsing using Nokogumbo's API, code, and fork of Gumbo.
* The Nokogumbo maintainers feel welcomed into the Nokogiri maintainer team.
* Upgrade paths for users are easy to navigate, and users don't get "stuck" on old versions.
* The Gumbo parser ships as a precompiled library in all the native platform gems supported by Nokogiri v1.11.
* The end-state of Nokogumbo is a (mostly) empty gem that prompts users to replace their dependency on Nokogumbo with Nokogiri.

For more specific objectives, see the "Punchlist" section below.

## Note: No JRuby support at this time

> :warning: **JRuby support is *not* going to be addressed as part of this work.** However, we feel it is important to think about and we hope to work on this in the future. If you're interested in helping with HTML5 support on JRuby, please comment on this issue or ping the maintainers on [the mailing list or the Discord channel](https://nokogiri.org/tutorials/getting_help.html#ask-for-help).

The Nokogumbo code relies on a parser implemented in C, and a C extension that is tightly coupled to libxml2. As a result, the `Nokogiri::HTML5` module will not be immediately available on JRuby, which uses Xerces in place of libxml2.

We ask that all downstream libraries be aware of this platform limitation as they consider using `HTML5` parsing methods post-merger.

## Risks

This work doesn't feel very risky, but if I had to name the riskiest bits:

* Precompiling Gumbo for all our platforms might be challenging.
* Gumbo is no longer supported upstream by Google, so we are assuming responsibility for keeping it secure, performant, and up-to-date with the HTML5 spec.
* There's no way to prevent using a new Nokogiri with an old Nokogumbo and seeing lots of annoying "constant redefined" warning messages.

Some things that could have been risky but aren't:

* License compatibility.
[...]

PR #2281, opened 2021-06-26
Update html5 tests

stevecheckoway:

html5lib-tests has updated since I forked it to fix a bunch of errors. Some of the updates include new tests to test functionality that changed in the spec. This updates the code to parse the new tests, updates the submodule to include my fixes to the latest version of html5lib-tests, and updates the gumbo parser to accommodate the spec changes.

The new tests were mentioned in https://github.com/sparklemotion/nokogiri/issues/2204#issuecomment-863550189.

Yes. The changes to the parser include parser-specific tests and the update to html5lib-tests includes end-to-end tests to test the behavior.

This only affects the C implementation as the gumbo parser isn't use for the Java implementation.

### DIFF
--- a/gumbo-parser/src/parser.c
+++ b/gumbo-parser/src/parser.c
@@ -4437,7 +4437,13 @@ static void handle_in_foreign_content(GumboParser* parser, GumboToken* token) {
     ) {
       pop_current_node(parser);
     }
-    handle_in_body(parser, token);
+    // XXX: The spec currently says to handle this using the in body insertion
+    // mode rules. That seems wrong. See
+    // <https://github.com/whatwg/html/issues/6808>. Instead, use the current
+    // insertion mode which seems like it works.
+    //
+    // handle_in_body(parser, token);
+    handle_html_content(parser, token);
     return;
   }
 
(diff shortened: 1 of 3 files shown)


stevecheckoway:
@craigbarnes You'll probably want the changes to gumbo in this PR. The spec changed in its handling of some misformed HTML.

flavorjones:
@stevecheckoway Just as an FYI in case it ended up being not-obvious, here are the gumbo-related test runs for this PR:

- gumbo (linux): https://github.com/sparklemotion/nokogiri/runs/2922756048?check_suite_focus=true
- gumbo (macos): https://github.com/sparklemotion/nokogiri/runs/2922756061?check_suite_focus=true
- gumbo (windows): https://github.com/sparklemotion/nokogiri/runs/2922756054?check_suite_focus=true

And the html5lib tests run in every CI job (see https://github.com/sparklemotion/nokogiri/blob/5e65d56ba49e938b422bf048243ed34460dcc993/.github/workflows/ci.yml#L84-L86 for an example of where we make sure the submodule is checked out, and https://github.com/sparklemotion/nokogiri/runs/2922760696?check_suite_focus=true for an example of the 50,000+ assertions running!).

Everything is green! Go ahead and hit that merge button when you're ready.

flavorjones (approved):

stevecheckoway:
I don't believe I have access to merge.

craigbarnes:
@stevecheckoway Thanks for the heads up!

PR #2283, opened 2021-07-01
Declare functions to take no arguments to save a few bytes

XrXr:
In C, function declarations that have an empty parameter list means
the function can take any number of arguments. Declare functions
that are called with no arguments as taking no arguments instead.

Under the System V calling convention for x86-64, calling a function
that accepts any number of arguments takes more code, so making this
change saves a few bytes with GCC 11. Before this change:

    $ size tmp/x86_64-linux/stage/lib/nokogiri/nokogiri.so
       text	   data	    bss	    dec	    hex
    2236245	  58240	   6632	2301117	 231cbd

After:

    $ size tmp/x86_64-linux/stage/lib/nokogiri/nokogiri.so
       text	   data	    bss	    dec	    hex
    2236181	  58240	   6632	2301053	 231c7d

See also: C99 draft N1256 §6.7.5.3 paragraph 14

It's a trivial change.

No.


flavorjones:
TIL! Thanks, @XrXr. Just kicked off CI and will merge when it goes green.

stevecheckoway:
In case anyone is wondering, the reason is that the SysV ABI requires `al` to hold the number of arguments passed in vector registers in variadic functions. You can see it in action [here](https://godbolt.org/z/8K7rrx634)  (and if you add floating point arguments, you can see `eax` change values from 0 up to 8). And the [psABI-x86_64 specification](https://www.uclibc.org/docs/psABI-x86_64.pdf) describes this starting at the last paragraph of page 20.

PR #2284, opened 2021-07-07
libxslt: tests should handle changed XSLT template resolution

flavorjones:

Upstream libxslt started failing after https://gitlab.gnome.org/GNOME/libxslt/-/commit/b0074eeca3c6b21b4da14fdf712b853900c51635 which addresses https://gitlab.gnome.org/GNOME/libxslt/-/issues/55.

Update the test to order the templates appropriately to reflect this change to how templates are resolved.

This change affects only tests, not functional behavior.

This change to the tests is forward-looking, as the next version of libxslt is likely to include the related change to the xslt template resolver.


flavorjones:
See https://github.com/sparklemotion/nokogiri/runs/3006765122?check_suite_focus=true for the failing test suite that drove this change.

PR #2288, opened 2021-07-09
ci(upstream): run against latest html5lib-tests master

flavorjones:

As the `html5lib-tests` repo changes over time, I think it might be helpful to automatically run against the latest so that if any failures are introduced, we can address them proactively.

This PR adds a job to the `upstream` CI pipeline, which runs periodically every few days, running against the latest versions of some of Nokogiri's dependencies.

I'll rely on @stevecheckoway to tell me if this is _actually_ useful.

This PR introduces new test coverage.

No functional changes.


stevecheckoway (approved):
LGTM

Issue #2276, opened 2021-06-19
[bug] valgrind errors spotted in gumbo parser


flavorjones:
**Please describe the bug**

A CI test run caught some memory errors in gumbo.

The build log is at https://github.com/sparklemotion/nokogiri/pull/2272/checks?check_run_id=2862967128

Here's the relevant snippet:

```text
# Running tests with run options --seed 25442:

==489== Invalid read of size 8
==489==    at 0x8A170F4: parse_args_mark (gumbo.c:307)
==489==    by 0x490A58B: gc_mark_stacked_objects (gc.c:4737)
==489==    by 0x490A58B: gc_mark_stacked_objects_all (gc.c:4777)
==489==    by 0x490A58B: gc_marks_rest (gc.c:5653)
==489==    by 0x490AE9F: gc_marks (gc.c:5713)
==489==    by 0x490AE9F: gc_start (gc.c:6493)
==489==    by 0x490AE9F: gc_start (gc.c:6417)
==489==    by 0x490B5E2: garbage_collect (gc.c:6413)
==489==    by 0x490B5E2: gc_start_internal (gc.c:6722)
==489==    by 0x4A42198: vm_call_cfunc_with_frame (vm_insnhelper.c:1918)
==489==    by 0x4A42198: vm_call_cfunc (vm_insnhelper.c:1934)
==489==    by 0x4A4CEED: vm_exec_core (insns.def:915)
==489==    by 0x4A51FDC: vm_exec (vm.c:1778)
==489==    by 0x4A58F70: invoke_block (vm.c:979)
==489==    by 0x4A58F70: invoke_iseq_block_from_c (vm.c:1031)
==489==    by 0x4A58F70: invoke_block_from_c_bh (vm.c:1049)
==489==    by 0x4A58F70: vm_yield (vm.c:1094)
==489==    by 0x4A58F70: rb_yield_0 (vm_eval.c:970)
==489==    by 0x4A58F70: rb_yield_1 (vm_eval.c:976)
==489==    by 0x4A58F70: rb_yield (vm_eval.c:986)
==489==    by 0x487477B: rb_ary_each (array.c:1837)
==489==    by 0x4A42198: vm_call_cfunc_with_frame (vm_insnhelper.c:1918)
==489==    by 0x4A42198: vm_call_cfunc (vm_insnhelper.c:1934)
==489==    by 0x4A4D962: vm_exec_core (insns.def:850)
[...]

PR #2290, opened 2021-07-11
fix(gumbo): use C struct pointer directly instead of wrapping

flavorjones:

#2276 reports some valgrind warnings coming from the object used to wrap gumbo parse arguments.

This PR avoids the creation of an object and passes the C struct pointer directly, casting it as a `VALUE` when necessary.

I spent a few hours trying to reliably reproduce the #2276 warning and couldn't. :shrug: 

No functional changes.

### DIFF
--- a/ext/nokogiri/gumbo.c
+++ b/ext/nokogiri/gumbo.c
@@ -300,35 +300,10 @@ typedef struct {
   xmlDocPtr doc;
 } ParseArgs;
 
-static void
-parse_args_mark(void *parse_args)
-{
-  ParseArgs *args = parse_args;
-  rb_gc_mark_maybe(args->input);
-  rb_gc_mark_maybe(args->url_or_frag);
-}
-
-// Wrap a ParseArgs pointer. The underlying ParseArgs must outlive the
-// wrapper.
-static VALUE
-wrap_parse_args(ParseArgs *args)
-{
-  return Data_Wrap_Struct(rb_cObject, parse_args_mark, RUBY_NEVER_FREE, args);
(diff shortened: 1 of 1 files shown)


stevecheckoway:
I wrote this code and it wouldn't shock me to learn that I misunderstood the requirements of these functions. The problem that I was trying to solve, if I recall correctly, was that an exception could be thrown in the middle of the parse which would leak memory.

I opted not to take the approach in #2276 because it wasn't clear to me what the Ruby runtime was going to do with the `VALUE`. If it's expecting it to be a real `VALUE` and trying to access the underlying components of it, then treating a pointer as a `VALUE` seems bad.

I would not expect the stack to unwind beyond the `rb_ensure()` before calling the second function. And I don't see why the runtime would hold on to the passed in `VALUE` beyond those two function calls. Maybe this needs to be paired with `rb_rescue()` to work correctly?

larskanis:
As described in https://github.com/sparklemotion/nokogiri/issues/2276#issuecomment-878091518 this looks good to me.

stevecheckoway:
Oh, I see what happened. I wrote a comment in the PR, not in the linked issue and mixed up the numbers. Sorry for the confusion.

This also looks good to me after @larskanis's explanation. It'd be great if there were documentation for the Ruby API, but indeed casting a pointer to a `VALUE` for use in `rb_ensure` does seem like it works. E.g., https://github.com/ruby/ruby/blob/d09988502ed75cae65b787865465361b675cf1ee/ext/monitor/monitor.c#L180

PR #2294, opened 2021-07-17
feat(cruby): patch libxml2 to handle abruptly-closed HTML comments

flavorjones:

Hackerone user [tehryanx](https://hackerone.com/tehryanx?type=user) reported a method of potentially confusing Loofah sanitizers by taking advantage of the difference between how abruptly-closed comments are handled by libxml2 and how they're handled by WHATWG-guidance-compliant modern browsers.

WHATWG advises to treat `<!-->` and `<!--->` as empty comments, but libxml2 today treats these as the _start_ of a comment.

https://html.spec.whatwg.org/multipage/parsing.html#parse-error-abrupt-closing-of-empty-comment

I've submitted this patch upstream at https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/124

Similar prior art is for incorrectly-closed comments:

- https://github.com/sparklemotion/nokogiri/pull/2058
- https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/82

Yes!

This is a behavior change to libxml2/CRuby. Note that nekoHTML/JRuby already follow WHATWG guidance and so JRuby behavior is unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 
 ### Improved
 
+* [CRuby] Handle abruptly-closed HTML comments as WHATWG recommends for browsers. (Thanks to HackerOne user [tehryanx](https://hackerone.com/tehryanx?type=user) for reporting this!)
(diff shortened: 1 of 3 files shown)


flavorjones:
Tagging @tehryanx for the cross-over mention from Hackerone.

flavorjones:
I'm going to wait for feedback on the upstream merge request before merging this.

flavorjones:
OK, given that Nick has stepped down as a libxml2 maintainer, I'm less confident that the upstream commit will get reviewed or merged, so I'm planning on merging and releasing this in v1.13.0.

PR #2296, opened 2021-07-26
Resolve potential NPE in SchemaResourceResolver

pepijnve:

The LSResourceResolver interface specifies that the systemId parameter may be null. The SchemaResourceResolver uses this argument without checking for this which can result a NullPointerException. This patch adds an extra null check before using the parameter.

I'm seeing this NPE in my code, but I'm afraid I haven't been able to derive a simple reproduction case yet. No test yet either as a consequence.

Only the Java implementation.

### DIFF
--- a/ext/java/nokogiri/XmlSchema.java
+++ b/ext/java/nokogiri/XmlSchema.java
@@ -276,7 +276,7 @@ public class XmlSchema extends RubyObject
                     String systemId,
                     String baseURI)
     {
-      if (noNet && (systemId.startsWith("http://") || systemId.startsWith("ftp://"))) {
+      if (noNet && systemId != null && (systemId.startsWith("http://") || systemId.startsWith("ftp://"))) {
         if (systemId.startsWith(XMLConstants.W3C_XML_SCHEMA_NS_URI)) {
           return null; // use default resolver
         }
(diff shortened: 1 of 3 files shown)


flavorjones:
Hi, @pepijnve, thanks for the PR. I'd love to be able to reproduce what you're seeing -- can you help me understand where the challenge is in finding a repro? Maybe I can help!

pepijnve:
> can you help me understand where the challenge is in finding a repro?

The only issue is time 😄 I simply haven't tried to isolate this just yet.

This issue occurred for me in https://github.com/asciidoctor/asciimath/blob/master/spec/parser_spec.rb#L10 when running the specs under JRuby. I'm loading the MathML schema (https://github.com/asciidoctor/asciimath/blob/master/spec/schema/mathml2/mathml2.xsd) from disk. The root schema file loads fine as an XML document, but when attempting to interpret it as a schema, the NPE happens when attempting to resolve the relative paths to the other schema files.

pepijnve:
I'm going to try loading those schemas from a standalone script to see if I can reproduce the problem and if so see if I can get the same error with a simpler schema. I'll let you know how that goes.

flavorjones:
@pepijnve Thanks for the more detailed information! I'll take a look this weekend.

flavorjones:
I've added a test that isolates this bug to the PR. Let's make sure CI goes green!

flavorjones:
Merging, and targetting v1.13 for this. Thanks again for your contribution!

flavorjones:
This has been released as part of v1.12.4.

Issue #2300, opened 2021-08-03
[bug] `require_relative` requires binary extention in unexpected location


ojab:
**Please describe the bug**

In general binary extensions are residing in a [separate directory](https://github.com/rubygems/rubygems/blob/master/lib/rubygems/basic_specification.rb#L104) that is added to `$LOAD_PATH`, so it works fine. But
https://github.com/sparklemotion/nokogiri/commit/0cee8f01f2a07219407d3ffe5237f46714f71bd9#diff-a610b0ddcdf14966b304b7f6ffc87a1c5866f232f995a0a27d1618b23658b12aR25 changed `require` (which searched this directory) to `require_relative`, which looks only in `gems/nokogiri-*/lib/nokogiri`.

Currently after build on ruby platform we have
```
$ find vendor/bundle/ -name nokogiri.so
vendor/bundle/ruby/3.0.0/extensions/x86_64-linux/3.0.0/nokogiri-1.12.0/nokogiri/nokogiri.so
vendor/bundle/ruby/3.0.0/gems/nokogiri-1.12.0/ext/nokogiri/nokogiri.so
vendor/bundle/ruby/3.0.0/gems/nokogiri-1.12.0/lib/nokogiri/nokogiri.so
```
so it works fine.

But for size-cutting some people are using `find ${LOCAL_BUNDLE_PATH}/gems/ -name '*.so' -delete` that removes seemingly excessive binary exts, after that we have `.so` only in `extension_dir`, so `require_relative` from `gems/nokogiri-*/lib/nokogiri/extension.rb` fails, even though we have ext in `$LOAD_PATH`.
I can provide reproducer Dockerfile if needed and I can make a PR that reverts this particular `require_relative` if it's acceptable.

**Expected behavior**

Nokogiri works fine when binary ext is removed from `gems` directory.

**Environment**

Nokogiri (1.12.0) for `ruby` platform built for x86_86, docker image build from `ruby:3.0.2-slim` with `find ${LOCAL_BUNDLE_PATH}/gems/ -name '*.so' -delete` after gems installation.

flavorjones:
@ojab Thanks for reporting this, and I'm sorry if you're having problems.

Can you help me understand why "some people" are doing this? Specifically, is this officially recommended somewhere to maintain smaller install sizes? I do want to make sure Nokogiri is following best practices for C extensions, and I feel like I need to understand this a bit better.

flavorjones:
[...]

PR #2301, opened 2021-08-03
ext: load C extension files from non-native gems via "require"

flavorjones:

As reported in #2300, we should be loading the C extension relying on `require` and `$LOAD_PATH` and not using `require_relative`.

No. This is a bit challenging to test, as we'd have to modify the gem installation (and delete the `nokogiri.so` files present under "lib/nokogiri"), and the regression value is pretty low (especially given the verbose comment in this PR).

That said, this change has adequate test coverage already in the `gem-install` CI pipeline, to ensure we don't break unmodified installations.

No.

### DIFF
--- a/lib/nokogiri/extension.rb
+++ b/lib/nokogiri/extension.rb
@@ -2,6 +2,7 @@
 
 # load the C or Java extension
 begin
+  # native precompiled gems package shared libraries in <gem_dir>/lib/nokogiri/<ruby_version>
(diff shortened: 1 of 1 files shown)


flavorjones:
Released in [Release 1.12.2 / 2021-08-04 · sparklemotion/nokogiri](https://github.com/sparklemotion/nokogiri/releases/tag/v1.12.2)

PR #2303, opened 2021-08-06
c99 fixes

flavorjones:

#2302 reports a compilation failure on RHEL6, RHEL7, and SLES12. These are older distros (RHEL7 is running gcc 4.8) which seem to compile in a C90-ish mode by default that gives errors on C99-isms.

This PR removed the C99isms (which are errors) as well as some C90isms (which generate warnings). Specifically:

- don't declare a variable in a `for` statement (C99-ism)
- don't mix declarations and code (C90-ism)

This PR also cleans up some C casting related to how we're handling `const` qualifiers. Specifically:

- `gumbo.c` is now using `(const xmlChar *)` in place of `BAD_CAST` (which is an alias for `(xmlChar *)`)
- extracted the const-discard hack `(xmlChar *)(uintptr_t)ptr` into a macro `DISCARD_CONST_QUAL`
- introduce use of `DISCARD_CONST_QUAL` in one additional place in `xml_node.c`

As a result, even when using `-Wcast-qual` on older Rubies that support it, we now see no compiler warnings.

I considered setting up CI using a comparable Centos docker image, but I didn't feel like there was much regression value to that test, given the nature of this failure. I could be argued into doing it, if anyone feels strongly.

No functional changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,18 +4,25 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 
 ---
 
+## 1.12.3 / unreleased
+
+### Fixed
+
+* [CRuby] Fix compilation of libgumbo on older systems with versions of GCC that give errors on C99-isms. Affected systems include RHEL6, RHEL7, and SLES12. [[#2302](https://github.com/sparklemotion/nokogiri/issues/2302)]
+
+
 ## 1.12.2 / 2021-08-04
 
 ### Fixed
 
-* Ensure that C extension files in non-native gem installations are loaded using `require` and rely on `$LOAD_PATH` instead of using `require_relative`. This issue only exists when deleting shared libraries that exist outside the extensions directory, something users occasionally do to conserve disk space. [[#2300](https://github.com/sparklemotion/nokogiri/issues/2300)]
+* [CRuby] Ensure that C extension files in non-native gem installations are loaded using `require` and rely on `$LOAD_PATH` instead of using `require_relative`. This issue only exists when deleting shared libraries that exist outside the extensions directory, something users occasionally do to conserve disk space. [[#2300](https://github.com/sparklemotion/nokogiri/issues/2300)]
 
 
 ## 1.12.1 / 2021-08-03
 
 ### Fixed
 
-* Fix compilation of libgumbo on BSD systems by avoiding GNU-isms. [[#2298](https://github.com/sparklemotion/nokogiri/issues/2298)]
+* [CRuby] Fix compilation of libgumbo on BSD systems by avoiding GNU-isms. [[#2298](https://github.com/sparklemotion/nokogiri/issues/2298)]
 
 
 ## 1.12.0 / 2021-08-02
--- a/ext/nokogiri/gumbo.c
+++ b/ext/nokogiri/gumbo.c
@@ -89,15 +89,19 @@ get_parent(xmlNodePtr node)
 static GumboOutput *
 perform_parse(const GumboOptions *options, VALUE input)
 {
+  GumboOutput *output;
+  const char *status_string;
+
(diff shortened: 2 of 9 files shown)


stevecheckoway (commented):
FWIW, I think this is not a good change. C99 is now more than 20 years old. Unless we're really targeting a compiler that doesn't support it (and we can't be because gumbo itself makes heavy use of C99 features), we should figure out how to pass -std=c99 to the compiler.

(If there were features of C11 that I thought were useful, I'd say we should use -std=c11, but there really aren't for this.)

stevecheckoway:
That should have referred only to https://github.com/sparklemotion/nokogiri/pull/2303/commits/3e3485b16eec427c0b25091eb3349acc01e09127 and not the PR as a whole.

stevecheckoway:
Does adding
```ruby
append_cflags("-std=c99")
```
[here](https://github.com/sparklemotion/nokogiri/blob/main/ext/nokogiri/extconf.rb#L596) not work?

flavorjones:
Hi Steve, thanks for bringing this up, and I'm sorry I didn't wait for your input before whacking the "merge" button. I almost entirely agree with you despite having introduced the change.

But I will point out that @nobu has put effort into keeping the codebase of Ruby, Nokogiri, and other gems "clean" and removing C90-isms because (I think) Ruby 2.6 and earlier still use them, and I want to respect his wishes despite not fully understanding them. I've tagged Nobu here in case he has additional insight.

My only hesitation about passing `-std=c99` or `-std=c11` is that I have an incomplete understanding of compiler behavior, and I was worried that it might break compilation for some obscure platform. But your point about libgumbo using it is a good one and I'm thinking about whether to revert it. I'll prep that revert/c99 change in a separate PR and wait a bit to see if Nobu has an opinion.

stevecheckoway:
Looks like `-pedantic` was removed from Ruby (by @nobu) in [2010](https://github.com/ruby/ruby/commit/d6fe62e9439751f743f37ff4333366157baee5ab) (although I don't claim to understand Ruby's build process).

It looks like `-std=gnu99` was added some point before Ruby 2.6. But again, this area is confusing.

flavorjones:
See #2304

AfirstS:
> **What problem is this PR intended to solve?**
> 
> #2302 reports a compilation failure on RHEL6, RHEL7, and SLES12. These are older distros (RHEL7 is running gcc 4.8) which seem to compile in a C90-ish mode by default that gives errors on C99-isms.
> 
> This PR removed the C99isms (which are errors) as well as some C90isms (which generate warnings). Specifically:
> 
> * don't declare a variable in a `for` statement (C99-ism)
> * don't mix declarations and code (C90-ism)
> 
> This PR also cleans up some C casting related to how we're handling `const` qualifiers. Specifically:
> 
> * `gumbo.c` is now using `(const xmlChar *)` in place of `BAD_CAST` (which is an alias for `(xmlChar *)`)
> * extracted the const-discard hack `(xmlChar *)(uintptr_t)ptr` into a macro `DISCARD_CONST_QUAL`
> * introduce use of `DISCARD_CONST_QUAL` in one additional place in `xml_node.c`
> 
> As a result, even when using `-Wcast-qual` on older Rubies that support it, we now see no compiler warnings.
> 
> **Have you included adequate test coverage?**
> 
> I considered setting up CI using a comparable Centos docker image, but I didn't feel like there was much regression value to that test, given the nature of this failure. I could be argued into doing it, if anyone feels strongly.
> 
> **Does this change affect the behavior of either the C or the Java implementations?**
> 
> No functional changes.

Revert file system link🔎🔎this is now🔎🔎🔗@AfirstSℹ️
68747470733a2f2f6e6f6b6f676972692e6f72672f696d616765732f6e6f6b6f676972692d73657269662d626c61636b2e706e67

PR #2305, opened 2021-08-10
Revert "ci: avoid warnings from setup-ruby-ref"

flavorjones:
This reverts commit a81343b2192ede40db8ddd89dd5a0a3e67c4c7a3.

Ruby-head on windows is failing. It's not clear to me why

The first failure was [here](https://github.com/sparklemotion/nokogiri/actions/runs/1105925545), with an error message indicating it might have been a server-side problem, but I recently introduced a81343b2192ede40db8ddd89dd5a0a3e67c4c7a3 to quash warnings and a different failure is now being raised, which seems to be packages not being found when enabling system libraries, and a segfault when using packaged libraries.

Here I'm just experimentally unwinding that change to see if I can learn what's going wrong.

### DIFF
--- a/.github/workflows/upstream.yml
+++ b/.github/workflows/upstream.yml
@@ -89,6 +89,7 @@ jobs:
           apt-get: "libxml2-dev libxslt1-dev pkg-config"
           mingw: "_upgrade_ libxml2 libxslt pkgconf"
           bundler-cache: true
+          setup-ruby-ref: MSP-Greg/ruby-setup-ruby/00-win-ucrt
       - uses: actions/cache@v2
         if: matrix.sys == 'disable'
         with:


flavorjones:
Editor's note: the ruby-head builds went green on this branch. Now I'm re-running https://github.com/sparklemotion/nokogiri/actions/runs/1114254585 on `main` to double-check that there was anything wrong under our control in the first place :shrug:

flavorjones:
Editor's note: re-running `main` still fails in interesting ways, so I guess this config is needed even though I don't understand what it's doing.

@MSP-Greg If you get a moment, can you help me understand why this config is needed and what it's doing? And why it's emitting warnings?

```yaml
        setup-ruby-ref: MSP-Greg/ruby-setup-ruby/00-win-ucrt
```

flavorjones:
Ah, I get it, that's used to set up ruby based on https://github.com/MSP-Greg/ruby-setup-ruby/blob/00-win-ucrt/dist/index.js

OK, all good, merging.

Issue #1617, opened 2017-03-17
Line number capped at 65535 but libxml2 already returns long.


pakore:
`ext/nokogiri/xml_node.c:1263` contains the line:

`return INT2NUM(xmlGetLineNo(node));`

but the signature of `xmlGetLineNo` according to libxml2 2.8.0  (tree.c:4527) is

```
long
xmlGetLineNo(xmlNodePtr node)
```

so libxml2 is returning a `long` that is passed to a macro function that accepts an `int` , so `long` is converted to `int`, and hence capped to 65535

The solution would be to change `ext/nokogiri/xml_node.c:1263` to

`return LONG2NUM(xmlGetLineNo(node));` 

and thus a problem that's been around for some  years would be fixed.

flavorjones:
The act of casting this return value from a long to an int is not the cause of line number overflow.

First, let's take a look at the relevant libxml2 struct:

```
/**
 * xmlNode:
 *
 * A node in an XML tree.
 */
typedef struct _xmlNode xmlNode;
typedef xmlNode *xmlNodePtr;
struct _xmlNode {
    void           *_private;	/* application data */
    xmlElementType   type;	/* type number, must be second ! */
    const xmlChar   *name;      /* the name of the node, or the entity */
    struct _xmlNode *children;	/* parent->childs link */
    struct _xmlNode *last;	/* last child link */
    struct _xmlNode *parent;	/* child->parent link */
    struct _xmlNode *next;	/* next sibling link  */
    struct _xmlNode *prev;	/* previous sibling link  */
    struct _xmlDoc  *doc;	/* the containing document */

    /* End of common part */
    xmlNs           *ns;        /* pointer to the associated namespace */
    xmlChar         *content;   /* the content */
    struct _xmlAttr *properties;/* properties list */
    xmlNs           *nsDef;     /* namespace definitions on this node */
    void            *psvi;	/* for type/PSVI informations */
    unsigned short   line;	/* line number */
    unsigned short   extra;	/* extra data for XPath/XSLT */
};
```

You can see that internally, libxml2 stores line number as an unsigned short.

Second, the size of `int` on most platforms is 4 bytes:

```
int main() {
  printf("sizeof(int) = %d\n", sizeof(int));
}
```

outputs

```
sizeof(int) = 4
```

But thanks for thinking of us! And thanks for using Nokogiri.

pakore:
Mmm I see. And changing that in libxml2 breaks ABI. Too bad. 

[...]

Issue #1505, opened 2016-07-07
Line number of RelaxNG error seems capped at 65535


mikevm:
When validating an XML document using RelaxNG, the line number of the error is capped at 65535.

test code: https://gist.github.com/mikevm/71095581645c62a971194b3142d09b92

output:
```
$ ruby -v
ruby 2.2.3p173 (2015-08-18 revision 51636) [x86_64-darwin14]
$ ruby ./test.rb
Nokogiri: 1.6.8
XML document with a schema violation on line 400
 0. RelaxNG reports error on line 400: Error validating datatype string
 1. RelaxNG reports error on line 400: Element item failed to validate content
XML document with a schema violation on line 40000
 0. RelaxNG reports error on line 40000: Error validating datatype string
 1. RelaxNG reports error on line 40000: Element item failed to validate content
XML document with a schema violation on line 70000
 0. RelaxNG reports error on line 65535: Error validating datatype string
 1. RelaxNG reports error on line 65535: Element item failed to validate content
```

**expected** the result of `#line` always returns the line on which the error occurred
**actual** the result of `#line` is capped at 65535.

larskanis:
In fact this is a shortcoming in libxml2. We can not fix it in nokogiri and it probably will not be fixed in libxml2. See: https://mail.gnome.org/archives/xml/2012-August/msg00168.html

flavorjones:
This will be fixed in v1.13.

Issue #1493, opened 2016-06-20
line numbers not working as expected


flavorjones:
(reported on nokogiri-talk by @justinthec)

I haven't had time to look into this, but wanted to capture it so that it wouldn't get forgotten.

---

> Hello,
> 
> Preface/libxml2
> 
> I've come across a bug that popped up once before in libxml2 back in 2008 where XML_TEXT_NODEs would always return 0 as their line number.
> 
> Link: https://mail.gnome.org/archives/xml/2008-July/msg00017.html
> 
> It was however patched that day and should have been fixed.
> 
> Commit: https://git.gnome.org/browse/libxml2/commit/?id=45efd0878aa56efa4291eec2ecdcff66d2f52fc3
> 
> Those three lines that he added still exist in the most current version of libxml2 now just under a conditional check for a (ctxt->linenumbers) flag.
> 
> Current Day libxml2: https://git.gnome.org/browse/libxml2/tree/SAX2.c?id=4472c3a5a5b516aaf59b89be602fbce52756c3e9#n1905
> 
> Nokogiri
> 
> This bug is currently present in Nokogiri 1.6.8.
> 
> I've been looking through the source in an attempt to find a fix and I'd appreciate some guidance:
> 
> Possibly relevant files:
> https://github.com/sparklemotion/nokogiri/blob/8e305e4e3d0ab697d1c6bc506398fb3007baf874/ext/nokogiri/xml_node.c#L1258
> https://github.com/sparklemotion/nokogiri/blob/027a151102083258f5e297c4fecfddca56904a2a/ext/nokogiri/xml_sax_parser_context.c#L171
> 
> Any ideas?
> 
> Thanks in advance,
> 
> Justin

justinthec:
Thanks Mike, if you have any ideas but don't have time to try them out I'd be happy to take it on.

The farthest I've gotten so far is `xmlGetLineNo()` being called, but the source for that function is nowhere to be found, most likely in the compiled source of whichever version of libxml you guys are using.

Xliff:
Is anyone still interested in this?

If so, it looks like xmlGetLineNo just returns the results of xmlGetLineNoInternal, which you can find, here:
**https://github.com/GNOME/libxml2/blob/master/tree.c#L4593**

flavorjones:
This will be fixed in v1.13.

Issue #1003, opened 2013-11-14
Nokogiri::XML::Node#line overflows


gwang:
Not sure if anyone reported this before:

The return of the #line method obviously overflows when there are more than 65535 lines in an XML. 

http://nokogiri.org/Nokogiri/XML/Node.html#method-i-line

flavorjones:
This will be fixed in v1.13.

PR #2309, opened 2021-08-14
feat(cruby): support line numbers larger than a short

### LINKED ISSUE #1764: investigate using XML_PARSE_BIG_LINES to tell libxml2 to track line numbers bigger than a short int

flavorjones:

As noted in #1493, #1617, #1505, #1003, and #533, libxml2 has not historically supported line numbers greater than a `short int`. Starting in libxml v2.9.0, setting the parse option `BIG_LINES` would allow tracking line numbers in longer documents.

Specifically this PR makes the following changes:

- set `BIG_LINES` parse option by default which will allow `Node#line` to return large integers
- allow `Node#line=` to set large line numbers on text nodes

Fixes #1764 

Yes!

JRuby's Xerces-based implementation did not suffer from this particular shortcoming, although its line number functionality is questionable in other ways (see #2177 / b32c875).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,13 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 
 ---
 
+## next / unreleased
+
+### Improved
+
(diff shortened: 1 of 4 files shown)


svoop:
@flavorjones I just came across this problem parsing an aeronautical XML file which counts a whopping 581857 lines. For audits, I have to reference the source node lines, so *short* doesn't really cut it. Big thanks for this PR! (And hope a fresh release will be cut soonish.)

flavorjones:
@svoop Yes, a new release will be out in the next week or two at most. (I want one in before we introduce Ruby 3.1 support.)

Issue #1790, opened 2018-08-22
`Nokogiri::XML::Node#add_child_node_and_reparent_attrs` behaves incorrectly if an attribute name has a colon


stevecheckoway:
**What problems are you experiencing?**
`Nokogiri::XML::Node#add_child_node_and_reparent_attrs` uses an incorrect test to see if a node's namespaces need to be reparented (at least, I think that's the purpose of this code).

It's testing `a.name =~ /:/` to decide to reparent. This breaks HTML elements which are allowed to have colons in their attribute names. (Essentially, HTML only allows foreign elements to have explicit namespaces.)

It's a little convoluted to demonstrate the problem using the Nokogiri API.

```ruby
require 'nokogiri'

doc = Nokogiri::HTML::Document.new
html = Nokogiri::XML::Element.new('html', doc)
html['a'] = 'en'
attr = html.attribute('a')
attr.name = 'xml:lang'

puts(html.attribute_nodes[0].inspect)
doc.add_child(html)
puts(html.attribute_nodes[0].inspect)
```

This prints

```
#<Nokogiri::XML::Attr:0x3ff75a84b6d8 name="xml:lang" value="en">
#<Nokogiri::XML::Attr:0x3ff75a84b340 name="lang" namespace=#<Nokogiri::XML::Namespace:0x3ff75a84b2dc prefix="xml" href="http://www.w3.org/XML/1998/namespace"> value="en">
```

I note that using `doc.root = html` _doesn't_ do this reparenting.

Here's a backwards compatible fix.

``` ruby
require 'nokogiri'

module Nokogiri
  module XML
    class Node
      # HTML elements can have attributes that contain colons.
      # Nokogiri::XML::Node#[]= treats names with colons as a prefixed QName
      # and tries to create an attribute in a namespace. This is especially
      # annoying with attribute names like xml:lang since libxml2 will
      # actually create the xml namespace if it doesn't exist already.
      def add_child_node_and_reparent_attrs node
        add_child_node(node)
        node.attribute_nodes.find_all { |a| a.namespace }.each do |attr|
          attr.remove
          node[attr.name] = attr.value
        end
      end
    end
  end
end

doc = Nokogiri::HTML::Document.new
html = Nokogiri::XML::Element.new('html', doc)
html['a'] = 'en'
attr = html.attribute('a')
attr.name = 'xml:lang'

puts(html.attribute_nodes[0].inspect)
doc.add_child(html)
[...]

PR #2310, opened 2021-08-14
refactor: relink_namespaces handles attributes properly

flavorjones:

See #1790 for context.

Previously both implementations of relink_namespaces skipped
attributes if the node itself didn't have a namespace.

This essentially reverts the implementation in 9fd03d8 by correcting
the behavior of relink_namespaces.

There's no functional change, existing tests are mostly sufficient ... but see #1790 for a discussion of whether we should be skipping re-namespacing in HTML documents.

I've make the behavioral change to both native implementations of `relink_namespaces`.


flavorjones:
I've added a commit that introduces test coverage for the HTML5 behavior from 7b3c972. This currently fails, will make it pass shortly.

flavorjones:
Should be green now, I'm curious what you think @stevecheckoway

stevecheckoway:
Hey @flavorjones, first let me apologize if I've missed other PRs or issues in which you've mentioned me. I need to get my GitHub settings squared away. Every time one of my students creates a repository (so once per assignment), GitHub adds it to my watch list. I'm apparently watching 1300 repos right now! The emails are overwhelming.

Namespaces in HTML documents is an area that I'm really pretty weak on. I found the HTML standard confusing and I haven't taken the time to dig in to what browsers do. HTML documents have nodes that live in one of [six namespaces](https://infra.spec.whatwg.org/#namespaces). HTML elements live in the HTML namespace, MathML elements live in the MathML namespace, and SVG elements live in the SVG namespace. The XLink, XML, and XMLNS namespaces are for attributes.

During parsing, `math` and `svg` start tags cause `math` and `svg` elements to be inserted in the MathML and SVG namespaces, respectively. Several places require attributes on foreign (i.e., not HTML) elements to [adjust attributes such that some end up in a namespace](https://html.spec.whatwg.org/multipage/parsing.html#adjust-foreign-attributes). Normally, HTML attributes should not have namespaces. (I have no idea if we handle that correctly or not currently.)

Here's an example showing what I think should happen.
```html
<!DOCTYPE html>
<svg viewBox="0 0 200 100" xmlns="http://www.w3.org/2000/svg">
  <title><span xml:lang="en-US">title</span></title>
  <text xml:lang="en-US" foo:bar="blah">This is some English text</text>
</svg>
<p xml:lang="en-US">More text.
```

- The (parser-inserted) `html`, `head`, and `body` elements are in the HTML namespace with null prefixes.
- The `svg` element is in the SVG namespace with null prefix.
- The `viewBox` attribute has no namespace.
- The `xmlns` attribute is in the XMLNS namespace with null prefix.
- The `title` element is in the SVG namespace with null prefix.
- **The `span` element is in the HTML namespace with null prefix**
- **The `span`'s `xml:lang` attribute has local name `xml:lang` and no namespace.**
- The `text` element is in the SVG namespace with null prefix.
- **The `text`'s `xml:lang` attribute is in the XML namespace with prefix `xml` and local name `lang`.**
- **The `text`'s `foo:bar` attribute has local name `foo:bar` and no namespace.**
- The `p` element is in the HTML namespace with null prefix.
- **The `p`'s `xml:lang` attribute has local name `xml:lang` and no namespace.**

I've bolded the parts that might be surprising.

As you can see, it's quite complicated. Some foreign elements, and only in some circumstances (namely when one of [these conditions](https://html.spec.whatwg.org/multipage/parsing.html#tree-construction) is satisfied), can contain HTML elements whose attributes never have namespaces and can contain colons. And only [some attributes with colons in the name](https://html.spec.whatwg.org/multipage/parsing.html#adjust-foreign-attributes) in foreign elements end up with namespaces.

(It's probably worth noting that this [impacts XPath  and XSLT](https://html.spec.whatwg.org/multipage/infrastructure.html#interactions-with-xpath-and-xslt).)

So the question becomes what behavior should the Nokogiri API have when adding nodes whose attributes contain colons?

As a point of comparison, I played around with JavaScript a little bit. If I create a 'svg' element, it puts it in the HTML namespace unless I explicitly tell it the namespace to use. If I create an attribute with the name `xml:lang`, it treats it as an attribute with no namespace. If I add that attribute to the `svg` element in the SVG namespace, it doesn't adjust the attributes.
```javascript
element = document.createElementNS("http://www.w3.org/2000/svg", "svg");
element.namespaceURI; // "http://www.w3.org/2000/svg"
attr = document.createAttribute("xml:lang");
attr.localName; // "xml:lang"
attr.prefix; // null
attr.namespaceURI; // null
element.setAttributeNode(attr)
element.attributes[0].localName; // "xml:lang"
element.attributes[0].prefix; // null
element.attributes[0].namespaceURI; // null
```

I'd be inclined to follow that behavior. If you want an element/attribute in a namespace, you gotta construct it in that namespace.

flavorjones:
@stevecheckoway Thanks for the context dump, and no need to apologize for notification fatigue.

I'm going to take some time to read everything you wrote and make sure I understand it, because my lack of knowledge of HTML5 is showing.

aditiagarw4722:
reviewed

flavorjones:
Rebased onto currrent `main`

PR #2353, opened 2021-11-02
dev: be specific about which files are auto-formatted

flavorjones:

Narrow the set of files formatted by rubocop and astyle to avoid temporary files.


stevecheckoway (approved):
LGTM

PR #2367, opened 2021-11-12
ci: oci image for alpine should have git installed

flavorjones:
Update the alpine CI image to install git, in order for actions/checkout to clone submodules


flavorjones:
waiting to see if this makes the musl build go green in https://github.com/sparklemotion/nokogiri/pull/2366

PR #2369, opened 2021-11-15
ci: revert buildx workaround in image-building

flavorjones:

Revert the workaround to https://github.com/docker/buildx/issues/834 that was solved by https://github.com/moby/buildkit/pull/2461


flavorjones:
Verified via https://github.com/sparklemotion/nokogiri/actions/runs/1462856106

Issue #2354, opened 2021-11-04
investigate why CI started failing on the Windows Ruby 2.5 builds


flavorjones:
Same commit was fine one day and reliably segfaults the next:

Good:
- windows disable 2.5: https://github.com/sparklemotion/nokogiri/runs/4096026023?check_suite_focus=true
- windows enable 2.5: https://github.com/sparklemotion/nokogiri/runs/4096025753?check_suite_focus=true

Bad:
- windows disable 2.5: https://github.com/sparklemotion/nokogiri/runs/4107938499?check_suite_focus=true
- windows enable 2.5: https://github.com/sparklemotion/nokogiri/runs/4107938236?check_suite_focus=true

All these runs were for the same commit, 3800769e37719beb09dc85718527105d1676dc17

Looking for differences in the logs:

virtual environment:
```patch
--- good.actions.log2   2021-11-04 13:14:27.862287741 -0400
+++ bad.actions.log2    2021-11-04 13:14:23.772273959 -0400
@@ -1,7 +1,7 @@
 Can't find any online and idle self-hosted or hosted runner in the current repository, account/organization  that mat
ches the required labels: 'windows-latest'
 Waiting for a self-hosted or a hosted runner to pickup this job...
 Job is waiting for a hosted runner to come online.
-Job is about to start running on the hosted runner: GitHub Actions 12 (hosted)
+Job is about to start running on the hosted runner: GitHub Actions 5 (hosted)
 Current runner version: '2.284.0'
 ##[group]Operating System
 Microsoft Windows Server 2019
@@ -10,9 +10,9 @@
 ##[endgroup]
 ##[group]Virtual Environment
 Environment: windows-2019
-Version: 20211018.0
-Included Software: https://github.com/actions/virtual-environments/blob/win19/20211018.0/images/win/Windows2019-Readm
e.md
-Image Release: https://github.com/actions/virtual-environments/releases/tag/win19%2F20211018.0
+Version: 20211102.4
+Included Software: https://github.com/actions/virtual-environments/blob/win19/20211102.4/images/win/Windows2019-Readm
e.md
+Image Release: https://github.com/actions/virtual-environments/releases/tag/win19%2F20211102.4
 ##[endgroup]
 ##[group]Virtual Environment Provisioner
 1.0.0.0-master-20211022-1
```

flavorjones:
The diff between those images' readmes:

```patch
--- 20211018.readme	2021-11-04 13:21:39.192621214 -0400
[...]

PR #2375, opened 2021-11-23
Actions - use windows-2022, add windows ucrt job

MSP-Greg:

Use a new Actions windows image, add full testing with Windows head build, which is a MSYS2 UCRT64 build, which will be released in a few weeks as 3.1.

The code used for Ruby setup uses the same branch as [PR 224](https://github.com/ruby/setup-ruby/pull/224) in [ruby/setup-ruby](https://github.com/ruby/setup-ruby).

Used existing...

NA

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,8 +235,8 @@ jobs:
       fail-fast: false
       matrix:
         sys: ["enable", "disable"]
-        ruby: ["2.5", "2.6", "2.7", "3.0", "mingw"]
-    runs-on: windows-latest
+        ruby: ["2.5", "2.6", "2.7", "3.0", "mingw", "head"]
+    runs-on: windows-2022
     steps:
       - name: configure git crlf
         run: |
@@ -245,11 +245,12 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
-      - uses: MSP-Greg/setup-ruby-pkgs@v1
+      - uses: MSP-Greg/setup-ruby-pkgs@win-ucrt-2
         with:
           ruby-version: "${{matrix.ruby}}"
           mingw: "libxml2 libxslt"
           bundler-cache: true
+          setup-ruby-ref: MSP-Greg/ruby-setup-ruby/win-ucrt-1
       - uses: actions/cache@v2
         if: matrix.sys == 'disable'
         with:


MSP-Greg:
Sorry, I'm missed the Windows CI in gem-install.yml...  I'll update it shortly.

flavorjones:
@MSP-Greg Thank you for this! Do you think you'll get to the gem-install and upstream pipelines? If not, just let me know and I'll try to do it.

MSP-Greg:
@flavorjones Sorry.  I got distracted, and then...  I'll finish them.

flavorjones:
@MSP-Greg I really appreciate it. Also, you may see Windows Ruby 2.5 builds failing. Those failures predate this PR and are described at [investigate why CI started failing on the Windows Ruby 2.5 builds · Issue #2354 · sparklemotion/nokogiri](https://github.com/sparklemotion/nokogiri/issues/2354)

My gut is to just drop 2.5 from the test matrix, since it will be unsupported in the next patch release of Nokogiri. I just didn't want you to spend time on trying to debug it.

MSP-Greg:
@flavorjones Thanks.  I saw the issue.  After I update for the PR, I may see if I can repo locally (I use Windows and WSL2/Ubuntu).

Question - much of the CI uses platform designations like `ubuntu-latest`.  With the change from `windows-2019` to `windows-2022`, they changed the installation of MSYS2 tools.  With other platforms, they may make similar changes, although the only other recent change (I recall) was adding `macOS-11`.

What do you think of removing all the `latest` uses and changing to numeric?  So, `ubuntu-latest` would change to `ubuntu-20.04`?  See [GitHub Actions - Available Environments](https://github.com/actions/virtual-environments#available-environments)

flavorjones:
> What do you think of removing all the latest uses and changing to numeric

For Mac and Windows I think it makes sense. For ubuntu, it's less obvious to me what the advantage would be, especially since so many of our tests just use ubuntu as a substrate for running another container, and because libc is usually very stable. I'm open to it if we can parameterize it as a matrix dimension.

Maybe the best thing would be for us to collaborate on a separate PR?

flavorjones:
I've moved this commit onto the 3.1 support branch at https://github.com/sparklemotion/nokogiri/pull/2408 and expanded it, but made sure you maintained co-authorship. Closing this PR in favor of that one.

MSP-Greg:
@flavorjones Cool & thanks & Happy New Year.

Sorry, been working on Actions, which has turned into a wonderful combination of ucrt, windows-2022, Ruby 3.1.0 and Bundler 2.3.4 fun...

flavorjones:
@MSP-Greg I've been following the ruby-build issue, thanks for your work on all of it!

PR #2383, opened 2021-12-10
dep(dev): rake-compiler 1.1.2 broke darwin native builds

flavorjones:

rake-compiler v1.1.2 broke darwin native builds, see https://github.com/sparklemotion/nokogiri/runs/4458677557?check_suite_focus=true for an example

This PR pins us to v1.1.1 for now

cc @larskanis


flavorjones:
I've filed an issue upstream at https://github.com/rake-compiler/rake-compiler/issues/195

PR #2385, opened 2021-12-11
dep: rake compiler ~> 1.1.4

flavorjones:

Use the new version of rake-compiler which fixes https://github.com/rake-compiler/rake-compiler/issues/195


flavorjones:
Closing in favor of #2386

Issue #2376, opened 2021-11-25
[bug] HTML5 foreign element namespaces should not be required by CSS queries


danblaker:
After upgrading to Nokogiri v1.12.0, I am unable to target inline svg elements via css or xpath selectors. Sample console commands:

```
html = <<-HTML.strip_heredoc
<!DOCTYPE html>
<html>
<head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <title>svg test</title>
</head>
<body>
  <div id="svg-container">
    <svg version="1.1" width="300" height="200" xmlns="http://www.w3.org/2000/svg">

      <rect width="100%" height="100%" fill="red" />

      <circle cx="150" cy="100" r="80" fill="green" />

      <text x="150" y="125" font-size="60" text-anchor="middle" fill="white">SVG</text>
    </svg>
  </div>
</body>
</html>
HTML

doc = Nokogiri::HTML5.parse(html)
doc.at('svg') # nil
doc.at_css('svg') # nil
doc.at_xpath('//svg') # nil
```

**Expected behavior**

Using the previous versions of Nokogiri (v 1.10.10) + nokogumbo (1.5.0), I was able to target SVG elements without any problems.

**Environment**

```
# Nokogiri (1.12.5)
    ---
    warnings: []
    nokogiri:
      version: 1.12.5
      cppflags:
      - "-I/usr/local/bundle/gems/nokogiri-1.12.5/ext/nokogiri"
      - "-I/usr/local/bundle/gems/nokogiri-1.12.5/ext/nokogiri/include"
      - "-I/usr/local/bundle/gems/nokogiri-1.12.5/ext/nokogiri/include/libxml2"
      ldflags: []
    ruby:
      version: 2.6.8
      platform: aarch64-linux-musl
      gem_platform: aarch64-linux
      description: ruby 2.6.8p205 (2021-07-07 revision 67951) [aarch64-linux-musl]
      engine: ruby
    libxml:
      source: packaged
      precompiled: false
      patches:
      - 0001-Remove-script-macro-support.patch
      - 0002-Update-entities-to-remove-handling-of-ssi.patch
      - 0003-libxml2.la-is-in-top_builddir.patch
      - 0004-use-glibc-strlen.patch
      - 0005-avoid-isnan-isinf.patch
      - 0006-update-automake-files-for-arm64.patch
      - 0007-Fix-XPath-recursion-limit.patch
[...]

PR #2403, opened 2021-12-25
HTML5 documents should not require namespaces in CSS selector queries

flavorjones:

Discussion at #2376 led to the decision to continue to parse HTML5 foreign elements with namespaces, but not to require that namespace in CSS queries.

So, in the canonical example:

```html
<!DOCTYPE html>
<html>
  <body>
    <div id="svg-container">
      <svg version="1.1" width="300" height="200" xmlns="http://www.w3.org/2000/svg">
        <rect width="100%" height="100%" fill="red" />
        <circle cx="150" cy="100" r="80" fill="green" />
        <text x="150" y="125" font-size="60" text-anchor="middle" fill="white">SVG</text>
      </svg>
    </div>
  </body>
</html>
```

Nokogiri will respond to the following searches:

```ruby
doc.css("svg") # ⇒ returns the svg element
doc.css("nsfoo|svg", { "nsfoo" => "http://www.w3.org/2000/svg" }) # ⇒ also returns the svg element
```

In order to accomplish this, this PR is rather invasive, but leaves a lot of code better factored, better documented, more flexible, and more extensible.

Specific changes:

- `Nokogiri::CSS.xpath_for` allows a `Nokogiri::CSS::XPathVisitor` to be injected for fine-grained control over how CSS selectors are translated into XPath queries
- the CSS cache is extended to account for XPathVisitor configuration
- patch libxml2 to support wildcard namespaces in XPath queries (nonstandard in XPath 1.0)
- `HTML5` documents and fragments translate child-axis searches into a wildcard-namespace query if libxml2 is patched; else use a builtin xpath function if libxml2; else use standard XPath local-name() (this order of preference reflects performance, faster-is-more-preferred)

Oh yesssss.

Also here are the benchmarks of the various implementations:

```text
      //span
           18.923k (± 8.9%) i/s -     93.906k in   5.010792s
      //*[local-name()='span']
            1.849k (± 2.8%) i/s -      9.261k in   5.011560s
      //*[nokogiri-builtin:local-name-is('span')]
            3.191k (± 2.4%) i/s -     16.150k in   5.064798s
      //*:span
           18.016k (± 4.6%) i/s -     89.900k in   5.003444s
    
      Comparison:
      //span:
          18922.5 i/s
      //*:span:
          18016.5 i/s - same-ish: difference falls within error
      //*[nokogiri-builtin:local-name-is('span')]:
           3190.6 i/s - 5.93x  (± 0.00) slower
      //*[local-name()='span']:
           1849.4 i/s - 10.23x  (± 0.00) slower
```

I'll also drop in some graphs of the callstacks below.

HTML5 is already only supported by CRuby, and so much of this functionality is simply not going to be turned on in JRuby. That said, if we need this functionality in JRuby, the fallback use of standard XPath `local-name()` would suffice until a need arose for performance optimizations there.


flavorjones:
Callstacks! You'll notice that the queries that use xpath functions end up spending a lot of time in the xpath function stack, and in allocating memory for the objects on that "stack" including the short-lived return values. The standard query and the wildcard query are faster because the comparisons are being done "inline" in the XPath query engine, without invoking xpath functions.


Callstack for standard child-axis query:
element-name

Callstack for local-name query:
local-name

Callstack for the builtin function query:
builtin

Callstack for the wildcard namespace query:
wildcard

flavorjones:
I could use a review in particular of the libxml2 patch https://github.com/sparklemotion/nokogiri/pull/2403/commits/b52dde7ddbe3b6dcc5a6dd6219160ef7b71f9014, because this feels a little bit risky.

Worth noting, though: with that patch applied, libxml2's full test suite still passes.

stevecheckoway:
I spent a little while looking at the xpath.c patch. Nothing jumps out at me as being wrong, but the file is huge and it's difficult to know what the implications are precisely. I'll try to look more closely this coming week.

flavorjones:
@stevecheckoway Thanks for the spot check. I'm going to merge this and accept a little bit of risk, because both libxml2 and nokogiri's full test suites still pass with that libxml2 patch applied.

stevecheckoway:
I think that makes sense. Sorry for not doing a better review in a reasonable amount of time.

flavorjones:
Oh, please don't apologize! I appreciate your attention.

Issue #2419, opened 2022-01-10
[bug] Regression in 1.13.0 in DocumentFragment#css?


CvX:
The following code used to work before 1.13.0:

```rb
fragments = Nokogiri::HTML5::fragment(html)
fragments.css("a/@href", "img/@src", "source/@src", "track/@src", "video/@poster")
```

Now it raises:

```
Nokogiri::XML::XPath::SyntaxError: ERROR: Invalid expression: .//*:a/*:@href | self::*:a/*:@href
```

Stack trace:

```
lib/nokogiri/xml/searchable.rb:222:in `evaluate'
lib/nokogiri/xml/searchable.rb:222:in `xpath_impl'
lib/nokogiri/xml/searchable.rb:208:in `block (2 levels) in xpath_internal'
lib/nokogiri/xml/searchable.rb:207:in `each'
lib/nokogiri/xml/searchable.rb:207:in `block in xpath_internal'
lib/nokogiri/xml/node_set.rb:23:in `initialize'
lib/nokogiri/xml/searchable.rb:206:in `new'
lib/nokogiri/xml/searchable.rb:206:in `xpath_internal'
lib/nokogiri/xml/node_set.rb:86:in `block in css'
lib/nokogiri/xml/node_set.rb:233:in `block in each'
lib/nokogiri/xml/node_set.rb:232:in `upto'
lib/nokogiri/xml/node_set.rb:232:in `each'
lib/nokogiri/xml/node_set.rb:85:in `inject'
lib/nokogiri/xml/node_set.rb:85:in `css'
lib/nokogiri/xml/document_fragment.rb:107:in `css'
```

**Additional context**

You can see the failure in the wild here: https://github.com/discourse/discourse/runs/4768400926?check_suite_focus=true#step:25:23

I don't know if it's related to https://github.com/sparklemotion/nokogiri/issues/2418

flavorjones:
@CvX Thanks for reporting this! It's unrelated to #2418.

This is an interesting failure -- and is most likely happening because nokogiri's ruby code is running against something besides the packaged/vendored libxml2 (with the patch `0009-allow-wildcard-namespaces.patch` applied).

Is it possible that you're pulling in a libxml2 library from somewhere else (an earlier version of nokogiri or libxml-ruby or the system libxml2)? Can you please reply with:

- the output of `nokogiri -v`
- the output of `bundle show` and `bundle show --paths` if you're using bundler

[...]

PR #2422, opened 2022-01-12
fix: regression with XPath attributes in CSS selectors

flavorjones:

Issue #2419 reported that a regression in v1.13.0 that raised an `Nokogiri::XML::XPath::SyntaxError` when parsing XPath attributes mixed into the CSS query, for example `node.css("img/@href")`.

Although this mash-up of XPath and CSS syntax previously worked _unintentionally_, it is now an **officially supported** feature and is documented as such.

Yes!

No.


CvX:
On my end I can confirm this makes Discourse specs pass again! 🎉

PR #2432, opened 2022-01-19
dep(jruby): use jar-dependencies to manage most java dependencies

flavorjones:

Addresses part of #1253

These deps are managed using jar-dependencies:

- isorelax
- jing
- serializer
- xalan
- xercesImpl
- xml-apis

Note that the following versions were changed:

- jing: com.thaiopensource:jing:20091111 → nu.validator:jing:20200702VNU
- xercesImpl: xerces:xercesImpl:2.12.0 → xerces:xercesImpl:2.12.1

Note that the following new dependences were introduced:

- Saxon-HE 9.6.0-4 (via nu.validator:jing:20200702VNU)

Finally, note that we still use a local jar for nekohtml and nekodtd, which is not the ideal end
state.

Tagging @headius and @mkristian for feedback on this PR since this is all pretty new to me.

Existing tests that include packaging and testing a java platform gem should be sufficient.

No.


flavorjones:
This seems to work, so I'm going to merge it. I'm curious if anyone has an opinion about vendoring-or-not-vendoring the jar files, please comment or ping me directly with your thoughts.

QroV:
start

PR #2448, opened 2022-02-11
improve encoding reader performance

flavorjones:

HTML4 document encoding detection improvements.

Yes, there is a complete perf testing approach introduced as a benchmarking test suite.

No.


Preethi-1606:
Done

Issue #2450, opened 2022-02-13
HTML5 parser moves last child text node of <form> tag outside the tag


npickens:
**Please describe the bug**

HTML5 parser moves the last child text node of a `<form>` tag outside the tag (as next sibling). Example:

```ruby
Nokogiri::HTML5.fragment('<form>Some text here</form>').to_html
# => "<form></form>Some text here"
```

Expected behavior:

```ruby
Nokogiri::HTML5.fragment('<form>Some text here</form>').to_html
# => "<form>Some text here</form>"
```

**Help us reproduce what you're seeing**

```ruby
#!/usr/bin/env ruby

require 'nokogiri'
require 'minitest/autorun'

class Test < MiniTest::Spec
  describe "HTML5::fragment" do
    it "should parse last child text node of <form> tag as last child text node" do
      html = '<form>Some text here</form>'
      doc = Nokogiri::HTML5.fragment(html)
      
      assert_equal(html, doc.to_html)
    end
  end
end
```

**Environment**

```
# Nokogiri (1.13.1)
    ---
    warnings: []
    nokogiri:
      version: 1.13.1
      cppflags:
      - "-I[...]/ruby/3.0.0/gems/nokogiri-1.13.1-x86_64-darwin/ext/nokogiri"
      - "-I[...]/ruby/3.0.0/gems/nokogiri-1.13.1-x86_64-darwin/ext/nokogiri/include"
      - "-I[...]/ruby/3.0.0/gems/nokogiri-1.13.1-x86_64-darwin/ext/nokogiri/include/libxml2"
      ldflags: []
    ruby:
      version: 3.0.3
      platform: x86_64-darwin21
      gem_platform: x86_64-darwin-21
      description: ruby 3.0.3p157 (2021-11-24 revision 3fb7d2cadc) [x86_64-darwin21]
      engine: ruby
    libxml:
      source: packaged
      precompiled: true
      patches:
      - 0001-Remove-script-macro-support.patch
      - 0002-Update-entities-to-remove-handling-of-ssi.patch
      - 0003-libxml2.la-is-in-top_builddir.patch
      - 0004-use-glibc-strlen.patch
      - 0005-avoid-isnan-isinf.patch
      - 0006-update-automake-files-for-arm64.patch
      - 0007-Fix-XPath-recursion-limit.patch
      - '0008-htmlParseComment-handle-abruptly-closed-comments.patch'
      - '0009-allow-wildcard-namespaces.patch'
[...]

PR #2451, opened 2022-02-14
Correctly handle text in form elements

stevecheckoway:
In the usual course of events, a `</form>` tag does _not_ pop all of the
nodes off the stack of open elements until a `form` element is popped.
Instead, it only removes the `form` element itself. Popping in the
normal fashion will flush any text node that is in the process of being
constructed.

This commit adds that flushing when `</form>` is encountered as well.

Fixes: #2450

#2450 

Yes. The Gumbo parser's test suite was updated to test for this.

It affects the C implementation's HTML5.


flavorjones:
@stevecheckoway Thanks for the fast analysis and fix! If you're happy with it, whack that merge button.

npickens:
@flavorjones @stevecheckoway Thanks for jumping on this so quickly! Appreciate it.

flavorjones:
I'm hoping to be able to cut a release in the next week that will contain this fix. Will update this issue when I do.

PR #2453, opened 2022-02-16
ci: Change Ruby setup to standard

MSP-Greg:

Currently, Actions CI is using non-standard branches for Ruby setup.  When the branches were used, changes had not been merged to the various actions repos (MSP-Greg/setup-ruby-pkgs, ruby/setup-ruby).  Updates have been merged.

When tested in my fork, updated CI workflows ran as they do in main

na, as only workflow files have been changed

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -244,12 +244,11 @@ jobs:
       - uses: actions/checkout@v2
         with:
           submodules: true
-      - uses: MSP-Greg/setup-ruby-pkgs@win-ucrt-2
+      - uses: MSP-Greg/setup-ruby-pkgs@v1
         with:
           ruby-version: "${{matrix.ruby}}"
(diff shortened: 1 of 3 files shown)


MSP-Greg:
@flavorjones Some days I really wish MSYS2 had more resources.  The error in the CI workflow job [windows (disable, head)](https://github.com/sparklemotion/nokogiri/runs/5218903488?check_suite_focus=true#step:4:57) is caused by a slow connection:
```
error: failed retrieving file 'msys.db.tar.zst.sig' from mirror.umd.edu : Operation too slow.
Less than 1 bytes/sec transferred the last 10 seconds
```

It did pass in my fork.  I've recently seen similar errors in other repos, involving other files...

flavorjones:
@MSP-Greg No worries, we see that failure relatively often. Merging! And thank you again for everything you're doing behind the scenes to make Windows work well.

PR #2454, opened 2022-02-17
Stop checking encoding names

casperisfine:
Comparing the names is much less efficient than comparing the instance directly.

It may also change in the future: https://bugs.ruby-lang.org/issues/18576


flavorjones:
Thank you, @casperisfine!

Issue #2464, opened 2022-02-22
[install] gumbo.c compilation failure


joyboy67:
I'm trying to install Nokogiri on OpenBSD 7.0 (amd_64) I got a compilation failure even with CFLAGS=-std=c99

Thank you for your help :)

**Have you read and followed the installation tutorial at http://www.nokogiri.org/tutorials/installing_nokogiri.html?**

- [x] Yes!

**What is the complete output of `gem install` ?**

```
current directory: /usr/local/lib/ruby/gems/3.1.0/gems/nokogiri-1.13.3/ext/nokogiri
/usr/local/bin/ruby -I /usr/local/lib/ruby/3.1.0 -r ./siteconf20220222-48874-xtpw65.rb extconf.rb
checking for whether -I/usr/local/include is accepted as CPPFLAGS... yes
checking for whether -std=c99 is accepted as CFLAGS... yes
checking for whether -Wno-declaration-after-statement is accepted as CFLAGS... yes
checking for whether -g is accepted as CFLAGS... yes
checking for whether -Winline is accepted as CFLAGS... yes
checking for whether -Wmissing-noreturn is accepted as CFLAGS... yes
Building nokogiri using packaged libraries.
Static linking is enabled.
Cross build is disabled.
Using mini_portile version 2.8.0
checking for iconv... yes
---------- IMPORTANT NOTICE ----------
Building Nokogiri with a packaged version of libxml2-2.9.13.
Configuration options: --host\=x86_64-unknown-openbsd7.0 --enable-static --disable-shared --libdir\=/usr/local/lib/ruby/gems/3.1.0/gems/nokogiri-1.13.3/ports/x86_64-openbsd7.0/libxml2/2.9.13/lib --with-iconv\=yes --disable-dependency-tracking --without-python --without-readline --with-c14n --with-debug --with-threads --disable-shared --enable-static CFLAGS\=-O2\ -U_FORTIFY_SOURCE\ -g\ -fPIC
The following patches are being applied:
  - 0001-Remove-script-macro-support.patch
  - 0002-Update-entities-to-remove-handling-of-ssi.patch
  - 0003-libxml2.la-is-in-top_builddir.patch
  - 0004-use-glibc-strlen.patch
  - 0005-avoid-isnan-isinf.patch
  - 0006-update-automake-files-for-arm64.patch
  - 0008-htmlParseComment-handle-abruptly-closed-comments.patch
  - 0009-allow-wildcard-namespaces.patch
[...]

PR #2465, opened 2022-02-22
Rename gumbo.h to nokogiri_gumbo.h

stevecheckoway:
In some cases, building Nokogiri can pick up a system-installed
libgumbo. By renaming the one external header, this should no longer
happen.

Fixes: #2464

#2464

No. We might want to test the failing configuration explicitly.

No.


flavorjones (approved):
Works for me in my OpenBSD VM that reproduces #2464

PR #2478, opened 2022-03-17
Remove patch of libxml2 reg. isnan/isinf on musl systems

larskanis:
It is no longer necessary with rake-compiler-dock-1.2.1.

The nokogiri binary gem for linux wan't usable on musl based systems due to unresolved symbols `__isnan` and `__isinf`. This is described here: https://github.com/rake-compiler/rake-compiler-dock/issues/42

Tests on alpine are already part of the CI.

C only.


larskanis:
What a bummer! Now it segfaults on musl-linux: https://github.com/sparklemotion/nokogiri/runs/5582802058?check_suite_focus=true#step:5:669

flavorjones:
Ah, that was built with the docker image from 1.2.0 ... see https://github.com/sparklemotion/nokogiri/runs/5582709245?check_suite_focus=true

Let me update the CI pipeline to use the new image (1.2.1) and see what happens.

flavorjones:
@larskanis Passes with the new docker image! Merging.

larskanis:
Nice! Thank you!

PR #2480, opened 2022-03-18
C: Check whether memory allocation was successful

Garfield96:

The C extension didn't check whether memory allocations were successful (= `malloc`/`calloc` return a value other than `NULL`). In the rare case, that memory allocation fails, this leads to undefined behavior, which usually results in a segmentation fault.

This problem was discovered using the [static analyzer](https://gcc.gnu.org/onlinedocs/gcc/Static-Analyzer-Options.html) (`-fanalyzer`) of GCC.

There is no easy way to test this behavior, hence I omitted test cases.

If no more memory is available `rb_eNoMemError` is raised. As far as I can tell, it is not necessary to change the Java implementation, since Java would automatically raise an `OutOfMemoryError` exception.

### DIFF
--- a/ext/nokogiri/nokogiri.h
+++ b/ext/nokogiri/nokogiri.h
@@ -195,7 +195,7 @@ NOKOPUBFUN VALUE Nokogiri_wrap_xml_document(VALUE klass,
 #define NOKOGIRI_SAX_SELF(_ctxt) ((nokogiriSAXTuplePtr)(_ctxt))->self
 #define NOKOGIRI_SAX_CTXT(_ctxt) ((nokogiriSAXTuplePtr)(_ctxt))->ctxt
 #define NOKOGIRI_SAX_TUPLE_NEW(_ctxt, _self) nokogiri_sax_tuple_new(_ctxt, _self)
-#define NOKOGIRI_SAX_TUPLE_DESTROY(_tuple) free(_tuple)
+#define NOKOGIRI_SAX_TUPLE_DESTROY(_tuple) ruby_xfree(_tuple)
 
 #define DISCARD_CONST_QUAL(t, v) ((t)(uintptr_t)(v))
 #define DISCARD_CONST_QUAL_XMLCHAR(v) DISCARD_CONST_QUAL(xmlChar *, v)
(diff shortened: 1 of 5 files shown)


larskanis:
Wouldn't it make sense to use [ruby_xmalloc](https://github.com/ruby/ruby/blob/cb4e2cb55a59833fc4f1f6db2f3082d1ffcafc80/include/ruby/internal/xmalloc.h#L67-L86) instead?

Garfield96:
Good idea. I don't have much experience with the Ruby API and wasn't aware of this function. I'll change it.

flavorjones:
+1 we should have been using `ruby_xmalloc` all along (an in fact [this is what we configure libxml2 to use internally](https://github.com/sparklemotion/nokogiri/blob/388edb8ac4e91173273d49c49f2c1e1dd9f62732/ext/nokogiri/nokogiri.c#L220)).

flavorjones:
@Garfield96 Thanks so much for submitting this PR. Are you comfortable making the change that Lars suggested? If not, that's OK and I'm happy to do it.

Please also note there exists a `ruby_xcalloc` that would replace `calloc`.

Garfield96:
I started to implement it with ruby allocators and I'll push the new version tomorrow.

Garfield96:
Ruby allocators are now used everywhere, except for `vasprintf`. I tried to adapt the `vasprintf` implementation in `nokogiri.c` but it results in a segmentation fault. To not block this PR, I'll open another PR for `vasprintf`.

flavorjones:
Thanks! I'm on vacation for a few days but will take a look as soon as I
can.

flavorjones:
I pushed a second commit that uses `ruby_xfree` to free the `nokogiriXsltStylesheetTuple` struct in xslt_stylesheet.c, which I think you missed. Otherwise this looks good to me.

I've pinged @larskanis for his review as well, because he's very familiar with memory handling in C extensions.

flavorjones:
Appveyor is green, not sure why that's not reflected here in the PR checks. I'll give Lars a chance to comment, but I'm pretty confident this is good to merge so I'm planning to do that this weekend. Thank you!

flavorjones:
Merged! Thank you for your contribution, @Garfield96. I'll credit you in the CHANGELOG.

PR #2481, opened 2022-03-19
Use Ruby allocators for `vasprintf`

Garfield96:

This is a follow-up PR for #2480, which implements `vasprintf` with Ruby allocators.
Unfortunately, the current implementation results in a segmentation fault. I would highly appreciate it if somebody with more experience with the Ruby API could have a look at this since it looks like it might be a bug in the Ruby implementation of `vsnprintf`.

### DIFF
--- a/ext/nokogiri/nokogiri.c
+++ b/ext/nokogiri/nokogiri.c
@@ -52,12 +52,11 @@ void noko_init_test_global_handlers(void);
 static ID id_read, id_write, id_external_encoding;
 
 
-#ifndef HAVE_VASPRINTF
 /*
  * Thank you Geoffroy Couprie for this implementation of vasprintf!
  */
 int
-vasprintf(char **strp, const char *fmt, va_list ap)
+noko_vasprintf(char **strp, const char *fmt, va_list ap)
 {
   /* Mingw32/64 have a broken vsnprintf implementation that fails when
    * using a zero-byte limit in order to retrieve the required size for malloc.
@@ -65,14 +64,10 @@ vasprintf(char **strp, const char *fmt, va_list ap)
    */
   char tmp[1];
   int len = vsnprintf(tmp, 1, fmt, ap) + 1;
-  char *res = (char *)malloc((unsigned int)len);
-  if (res == NULL) {
-    return -1;
-  }
+  char *res = (char *)ruby_xmalloc((unsigned int)len);
   *strp = res;
   return vsnprintf(res, (unsigned int)len, fmt, ap);
 }
-#endif
 
 
 static VALUE
--- a/ext/nokogiri/nokogiri.c
+++ b/ext/nokogiri/nokogiri.c
@@ -63,7 +63,11 @@ noko_vasprintf(char **strp, const char *fmt, va_list ap)
    * So we use a one byte buffer instead.
    */
   char tmp[1];
-  int len = vsnprintf(tmp, 1, fmt, ap) + 1;
+  va_list tmp_ap;
+  va_copy(tmp_ap, ap);
+  int len = vsnprintf(tmp, 1, fmt, tmp_ap) + 1;
+  va_end(tmp_ap);
+
   char *res = (char *)ruby_xmalloc((unsigned int)len);
   *strp = res;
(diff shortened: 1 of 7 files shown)


flavorjones:
Thank you for submitting this PR! I've kicked off CI ... let's see what the valgrind runs tell us. :crossed_fingers:

flavorjones:
OK! CI crashes as well, and I can repro in my dev environment. I'm going to take some time to dive in with valgrind and see what I can figure out.

ext/nokogiri/nokogiri.c
```diff
@@ -52,27 +52,22 @@ void noko_init_test_global_handlers(void);
...
    * using a zero-byte limit in order to retrieve the required size for malloc.
    * So we use a one byte buffer instead.
    */
   char tmp[1];
   int len = vsnprintf(tmp, 1, fmt, ap) + 1;
-  char *res = (char *)malloc((unsigned int)len);
-  if (res == NULL) {
-    return -1;
-  }
+  char *res = (char *)ruby_xmalloc((unsigned int)len);
```
stevecheckoway, line 71: `ruby_xmalloc()` returns `void *`, the return values from the `malloc` family of functions should not be cast.

With the previous change above, this line can become
```c
char *rest = ruby_xmalloc(len);
```

flavorjones:
OK, so on linux the va_list can't be re-used once va_arg has been called on it. So I've pushed an additional commit to introduce a temporary copy of the va_list for the first call to vsnprintf.

But I think we can actually get rid of vasprintf entirely by simply using `rb_vsprintf` ... going to spike on that now.

ext/nokogiri/nokogiri.c
```diff
@@ -63,7 +63,11 @@ noko_vasprintf(char **strp, const char *fmt, va_list ap)
    * So we use a one byte buffer instead.
    */
   char tmp[1];
-  int len = vsnprintf(tmp, 1, fmt, ap) + 1;
+  va_list tmp_ap;
+  va_copy(tmp_ap, ap);
+  int len = vsnprintf(tmp, 1, fmt, tmp_ap) + 1;
```
stevecheckoway, line 68: The return value for `vsnprintf()` should be checked against `-1`.

```c
int ret = vsnprintf(tmp, 1, fmt, ap);
if (ret == -1)
{
  *strp = NULL;
  return -1;
}
size_t len = ret + 1;
```

ext/nokogiri/nokogiri.c
```diff
@@ -63,7 +63,11 @@ noko_vasprintf(char **strp, const char *fmt, va_list ap)
...
+  va_end(tmp_ap);
+
   char *res = (char *)ruby_xmalloc((unsigned int)len);
   *strp = res;
   return vsnprintf(res, (unsigned int)len, fmt, ap);
```
stevecheckoway, line 73: `len` shouldn't be cast to an `unsigned int`, it's a `size_t`.

stevecheckoway: `va_end` needs to be called before returning the final result.
```c
ret = vsnprintf(res, len, fmt, ap);
va_end(ap);
return ret;
```

stevecheckoway (changes requested):

stevecheckoway:
Before running off to teach, I thought I saw a memory leak where a string was allocated via `noko_vasprintf()` but wasn't being freed. The PR changed while I was writing my review and I can't find that any more.

stevecheckoway:
Also I realized one of my review comments was wrong (about needing to call `ap_end()`; I deleted it).

stevecheckoway:
@flavorjones Will `rb_vsprintf()` work everywhere? It returns a `VALUE`.

flavorjones:
@stevecheckoway Thanks for the feedback on this!

I'm about to ship an alternative PR that removes calls to `vasprintf()` entirely, which I think is preferable. I'm going to hold off on making edits to this PR until we can compare the two.

flavorjones:
See #2489 for an alternative approach

flavorjones:
Closing in favor of #2489

Issue #2441, opened 2022-01-29
update JRuby dependencies to xercesImpl 2.12.2


flavorjones:
This is a placeholder ticket to consider https://github.com/advisories/GHSA-h65f-jvqw-m9fj

Issue #2437, opened 2022-01-26
publish patched NekoDTD to Maven Central


flavorjones:
## Problem summary

Currently Nokogiri's java native gem depends on patched versions of NekoDTD and NekoHTML. The binary artifacts only exist in Nokogiri's git repo and if I needed to reproduce them I would be starting from scratch.

## Historical context

In the past (e.g., 2e5ce32 in 2012 and 6166964 in 2015) we've patched NekoHTML and NekoDTD to fix bugs or parsing errors. These repositories live (I think) at:

- https://github.com/jvshahid/nekohtml
- https://github.com/jvshahid/nekodtd

A conversation at #1253 suggested using `jar-dependencies` to manage Java dependencies, and #2432 starting doing so for the artifacts listed on Maven Central.

## What do we want to do

This issue is intended to drive a conversation with the following potential outcomes:

- [x] ensure shared ownership of the patched NekoHTML and NekoDTD source repositories
- [x] set up a Maven Central groupId (@headius pointed us at https://central.sonatype.org/publish/publish-guide/)
- [x] https://github.com/sparklemotion/nekodtd/issues/1
- [x] ~https://github.com/sparklemotion/nekohtml/issues/3~ _this library is deprecated in nokogiri >= 1.14.0_
- [x] use `jar-dependencies` to manage those dependencies like our other ones

Doing all of the above would allow us to consider further patches on the Neko libraries, e.g., #2436.

cc @jvshahid whose support and assistance I will need

flavorjones:
@jvshahid Would you be open to transferring ownership of those repositories to me or to the sparklemotion org? That would unblock me to do the rest of the work.

flavorjones:
Thanks, @jvshahid! Repositories have been transferred to @sparklemotion/nokogiri-core

- https://github.com/sparklemotion/nekohtml
- https://github.com/sparklemotion/nekodtd

flavorjones:
I've created a JIRA account at OSSRH and requested a new project with the groupId `org.nokogiri`: https://issues.sonatype.org/browse/OSSRH-77736

headius:
@flavorjones As always let me know if I can help. I get a lot of notifications so I don't always see updates here, but I support this move and will try to help as time permits.

flavorjones:
@headius Thanks for checking in! I will be respectful of your time, but may ask for advice or a code review here and there as I work through this and a handful of other JRuby issues for v1.14.0.

[...]

Issue #2436, opened 2022-01-26
[bug][jruby] Strikethrough tag causes incorrect HTML parsing in JRuby


davue:
**Please describe the bug**
If you parse an HTML string containing a `<s></s>` tag. The elements above the `<s></s>` tag are duplicated into a seperate child element without content and the other child with the actual content will have inverted order. Strangely this only seems to happen with `<s></s>` tags, all other tags are parsed just fine.

**Help us reproduce what you're seeing**
Open a JRuby Rails console and run:
```
Nokogiri::HTML4::DocumentFragment.parse("<!DOCTYPE HTML><html><body><strong><s>Test</s></strong></body></html>")
```
The output will show that the parser created an additional `strong` child even though it is not present in the HTML input and the order of tags is inverted for the other child:
```
#(DocumentFragment:0x888 {
  name = "#document-fragment",
  children = [
    #(Element:0x88a { name = "strong" }),
    #(Element:0x88c { name = "s", children = [ #(Element:0x88e { name = "strong", children = [ #(Text "Test")] })] })]
  })
```

**Expected behavior**
I expect the output to look like in the CRuby implementation, where only one child element is present and the order is correct:
```
#(DocumentFragment:0x13254 {
  name = "#document-fragment",
  children = [
    #(Element:0x13268 { name = "strong", children = [ #(Element:0x1327c { name = "s", children = [ #(Text "Test")] })] })]
  })
```

**Environment**
```
# Nokogiri (1.13.1)
    ---
    warnings: []
    nokogiri:
      version: 1.13.1
    ruby:
      version: 2.6.8
      platform: java
      gem_platform: universal-java-1.8
      description: jruby 9.3.2.0 (2.6.8) 2021-12-01 0b8223f905 OpenJDK 64-Bit Server VM
        25.302-b08 on 1.8.0_302-b08 +jit [linux-x86_64]
      engine: jruby
      jruby: 9.3.2.0
    other_libraries:
      xerces: Xerces-J 2.12.0
      nekohtml: NekoHTML 1.9.21
```

flavorjones:
@davue Thanks for reporting this! That's certainly strange behavior, I'll take a look later today.

flavorjones:
OK, so here's a trick I use a lot to diagnose parser behavior like this: check the `doc.errors` to see what warnings and errors the parser noticed along the way.

```ruby
#! /usr/bin/env ruby

require "nokogiri"

[...]

PR #2482, opened 2022-03-22
Bump xerces version to 2.12.2

jsvd:

XercesJ 2.12.2 is a bugfix release that includes important fixes, including a parsing performance issue and a bug in the regex implementation.

This change should only fix bugs in the Java execution and there should be no changes in behavior.


flavorjones:
Thanks! I'm on vacation for a few days but will take a look as soon as I
can.

Note that I had #2441 open to track this work, and there are some related
issues I'm working through with the java dependencies more generally (see
#2436 and #2437) which aren't blocked but I would like to try to fix it all
before making a release.

flavorjones:
Thanks again for this contribution! I've added a couple of commits to update the jar files, rebase onto main, and update the CHANGELOG. If this goes green, I'll merge and it will be in v1.14.0.

PR #2486, opened 2022-03-29
Hoist loads and sink store in hot loop

stevecheckoway:
This is from @sebpop's [PR](https://github.com/google/gumbo-parser/pull/403/).

Since `char *` can alias with everything, the compiler is unable to
prove that `output->data` doesn't alias with `output->length` and
`output` itself which forces a load of `output->data` and
`output->length` and a store to `output->length` in each iteration of
the loop. By hoisting the two loads and sinking the store out of the
loop, we're left with a loop containing just a single store to
`data[length]`.

https://github.com/google/gumbo-parser/pull/403

There should be no observable changes other than this should be slightly faster so no tests were included.

No.

---

I think this change is fine (the tests pass), but it's not clear that the performance gain (which I suspect to be very slight) overcomes the (also quite slight) degradation in readability. Worth testing.

### DIFF
--- a/gumbo-parser/src/string_buffer.c
+++ b/gumbo-parser/src/string_buffer.c
@@ -72,10 +72,13 @@ void gumbo_string_buffer_append_codepoint (
     prefix = 0xf0;
   }
   maybe_resize_string_buffer(num_bytes + 1, output);
-  output->data[output->length++] = prefix | (c >> (num_bytes * 6));
+  size_t length = output->length;
+  char* data = output->data;
+  data[length++] = prefix | (c >> (num_bytes * 6));
   for (int i = num_bytes - 1; i >= 0; --i) {
-    output->data[output->length++] = 0x80 | (0x3f & (c >> (i * 6)));
+    data[length++] = 0x80 | (0x3f & (c >> (i * 6)));
   }
+  output->length = length;
 }
 
 void gumbo_string_buffer_append_string (


stevecheckoway:
By the way, one reason I'm a little skeptical about the performance gain is that the common case is going to be ASCII characters (things like tag names). Thus the body of the loop is going to execute exactly once. In the worst case, it'll execute 4 times per code point, so we're only saving a handful of loads and stores.

sebpop:
Nice to see that the patch is getting merged.
The patch for gumbo-parser was to improve performance of a benchmark in Geekbench.
I forgot the details, and I don't have access to Geekbench anymore after I left Samsung.

stevecheckoway:
Cool! Thanks for submitting the issue.

flavorjones:
I'm going to spend a bit of time writing a benchmark for this to understand the impact a bit better. Sorry for the delay on responding/reviewing!

flavorjones:
Rebased onto current `main`, will run benchmarks today.

flavorjones:
In a real-world document parsing benchmark there doesn't seem to be any change in performance.

Here's the benchmark, html4 parsing is presented simply for comparison (note that HTML4 parsing is not affected by the compiler option changes).

```ruby
#! /usr/bin/env ruby
# coding: utf-8

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "nokogiri", path: "."
  gem "benchmark-ips"
end

require "nokogiri"
require "benchmark/ips"

input = File.read("test/files/tlm.html")
puts "input #{input.length} bytes"

html4_doc = Nokogiri::HTML4::Document.parse(input)
html5_doc = Nokogiri::HTML5::Document.parse(input)

puts RUBY_DESCRIPTION

Benchmark.ips do |x|
  x.time = 10
  x.report("html5 parse") do
    Nokogiri::HTML5::Document.parse(input)
  end
  x.report("html4 parse") do
    Nokogiri::HTML4::Document.parse(input)
  end
  x.compare!
end
```

(note that's parsing [this test file](https://github.com/sparklemotion/nokogiri/blob/main/test/files/tlm.html))

Here's the benchmark results:


_main with default compiler options:_

```
input 70095 bytes
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]
Warming up --------------------------------------
         html5 parse    11.000  i/100ms
         html4 parse    34.000  i/100ms
         html5 parse    125.659  (±17.5%) i/s -      1.210k in  10.013864s
         html4 parse    380.538  (±22.6%) i/s -      3.638k in  10.011955s

Comparison:
         html4 parse:      380.5 i/s
         html5 parse:      125.7 i/s - 3.03x  (± 0.00) slower

```


_with this patch and default compiler options:_

```
input 70095 bytes
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]
Warming up --------------------------------------
         html5 parse    11.000  i/100ms
         html4 parse    33.000  i/100ms
         html5 parse    120.569  (±18.2%) i/s -      1.166k in  10.077918s
         html4 parse    365.071  (±25.5%) i/s -      3.465k in  10.094851s

Comparison:
         html4 parse:      365.1 i/s
         html5 parse:      120.6 i/s - 3.03x  (± 0.00) slower

```

---

_main with -O2 (see #2369):_

```
input 70095 bytes
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]
Warming up --------------------------------------
         html5 parse    22.000  i/100ms
         html4 parse    36.000  i/100ms
         html5 parse    249.689  (±19.2%) i/s -      2.398k in  10.001083s
         html4 parse    378.405  (±20.6%) i/s -      3.672k in  10.065013s

Comparison:
         html4 parse:      378.4 i/s
         html5 parse:      249.7 i/s - 1.52x  (± 0.00) slower
```

_this patch with -O2 (see #2369):_

```
input 70095 bytes
ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [x86_64-linux]
Warming up --------------------------------------
         html5 parse    19.000  i/100ms
         html4 parse    38.000  i/100ms
         html5 parse    247.586  (±22.2%) i/s -      2.356k in  10.081576s
         html4 parse    375.071  (±18.1%) i/s -      3.648k in  10.012846s

Comparison:
         html4 parse:      375.1 i/s
         html5 parse:      247.6 i/s - 1.51x  (± 0.00) slower
```

@sebpop or @stevecheckoway any thoughts? I'd be happy to run a more specific benchmark if you can write one.

flavorjones:
Bumping this - any ideas for a benchmark that will demonstrate this speedup?

PR #2489, opened 2022-04-01
remove calls to vasprintf

flavorjones:

#2481 proposed cleaning up our local implementation of `vasprintf` and using it on all platforms; however this PR takes an alternative approach of removing calls to `vasprintf` in favor of using `rb_vsprintf` since we need to create Ruby String objects from all the errors being handled this way.

Some nice side effects:

- ~TruffleRuby should like this better, since Sulong was missing functionality around varargs~
- we no longer have to do platform-specific detection of `vasprintf` at compile time

Note also that this PR is based on #2480, so that should be considered before this one.

Yes! Some of the error handling code did not have good, or any, coverage. Test coverage has been added in this PR.

No behavioral changes are being introduced.

cc @Garfield96 @stevecheckoway

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,8 @@ This version of Nokogiri uses [`jar-dependencies`](https://github.com/mkristian/
 
 * Compare `Encoding` objects rather than compare their names. This is a slight performance improvement and is future-proof. [[#2454](https://github.com/sparklemotion/nokogiri/issues/2454)] (Thanks, [@casperisfine](https://github.com/casperisfine)!)
 * Avoid compile-time conflict with system-installed `gumbo.h` on OpenBSD. [[#2464](https://github.com/sparklemotion/nokogiri/issues/2464)]
+* Remove calls to `vasprintf` in favor of platform-independent `rb_vsprintf`
+* Prefer `ruby_xmalloc` to `malloc` within the C extension. [[#2480](https://github.com/sparklemotion/nokogiri/issues/2480)] (Thanks, [@Garfield96](https://github.com/Garfield96)!)
 
 
 ## 1.13.3 / 2022-02-21
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -982,8 +982,6 @@ have_func("xmlRelaxNGSetValidStructuredErrors") # introduced in libxml 2.6.21
 have_func("xmlSchemaSetValidStructuredErrors") # introduced in libxml 2.6.23
 have_func("xmlSchemaSetParserStructuredErrors") # introduced in libxml 2.6.23
 
-have_func("vasprintf")
-
 other_library_versions_string = OTHER_LIBRARY_VERSIONS.map { |k, v| [k, v].join(":") }.join(",")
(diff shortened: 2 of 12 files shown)


larskanis:
Thanks for asking! This looks all great! No `RB_GC_GUARD` necessary.

There's a single remaining stdc `free()` call: https://github.com/sparklemotion/nokogiri/blob/1b7f25917e157a59656010e012c14f20630e2c49/ext/nokogiri/gumbo.c#L276 
It should be `gumbo_free()` instead. Or even better [`gumbo_alloc()`](https://github.com/sparklemotion/nokogiri/blob/1b7f25917e157a59656010e012c14f20630e2c49/gumbo-parser/src/util.c#L24) and siblings could be converted to `ruby_xmalloc` equally.

flavorjones:
@larskanis I saw that `free` call, but left it for now because `gumbo_alloc` and `gumbo_free` aren't public.

I agree it would be nice to add a feature to libgumbo to allow us to set the alloc/free functions (like `xmlMemSetup` does for libxml2), so I've created #2490 to drive that discussion.

stevecheckoway:
At the moment, Gumbo isn't tied to ruby in any way, I think it makes sense to keep it that way, absent a compelling reason.

Gumbo originally had a way of supporting other allocators, but it made the code worse and was removed. It makes sense to me to keep Gumbo's "public" API returning memory from the standard allocator. I don't think any of the utility functions in `util.h` should be part of Gumbo's public API.

I'd be in favor of removing `gumbo_free()` entirely and just replacing it with calls to `free()`.

flavorjones:
Ah, OK, @stevecheckoway I just saw your update and will defer to your decision if you feel strongly. Go ahead and close #2490 if you wish.

flavorjones:
I've rebased after merging #2480 and so this changeset should be more readable now.

flavorjones:
I also kicked off the truffle CI jobs on this branch because I think this PR may break some of the patches being applied in https://github.com/oracle/truffleruby/blob/master/lib/truffle/truffle/patches/nokogiri_patches.rb

cc @eregon @aardvark179

eregon:
@flavorjones Those runs failed, any idea of the issue?
For the first 2 I can't find the error, maybe `Expected '| ' but got `?

For the 3rd job it's:
```
 /__w/nokogiri/nokogiri/ext/nokogiri/xml_xpath_context.c:301:63: warning: function 'xpath_generic_exception_handler' declared 'noreturn' should not return [-Winvalid-noreturn]
va_list args; rb_raise(rb_eRuntimeError, "%s", "Exception:"); return;
                                                              ^
/__w/nokogiri/nokogiri/ext/nokogiri/xml_xpath_context.c:344:27: error: use of undeclared identifier 'errors'
xmlSetStructuredErrorFunc(errors, Nokogiri_error_array_pusher);
                          ^
1 warning and 1 error generated.
```

It fails the same way on master (https://github.com/sparklemotion/nokogiri/runs/5800516006?check_suite_focus=true)

flavorjones:
@eregon some of the TR patches are no longer applying, let's discuss at https://github.com/sparklemotion/nokogiri/issues/2491

PR #2531, opened 2022-05-07
ext: declare gnu_printf format attribute for some functions

flavorjones:

Commit 3831a00 replaced `vasprintf` calls with `rb_vsprintf`, however that caused gcc to start emitting warnings because of the `-Wsuggest-attribute=format` flag.

This PR introduces a `PRINTFLIKE_DECL` macro to declare the gnu_printf format attribute for some functions to quash those warnings.

It also replaces the `NORETURN` macro (which will come from ruby.h`) with `NORETURN_DECL` that doesn't take an argument, because that syntax is easier to combine with `PRINTFLIKE_DECL` IMHO.

No tests necessary.

No behavior change.

### DIFF
--- a/ext/nokogiri/nokogiri.h
+++ b/ext/nokogiri/nokogiri.h
@@ -83,11 +83,19 @@ xmlNodePtr xmlLastElementChild(xmlNodePtr parent);
 #  define NOKOGIRI_DEBUG_END(p)
 #endif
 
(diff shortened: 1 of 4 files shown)


amir2livepc (commented):
Let me see thx

amir2livepc:
This is me

PR #2532, opened 2022-05-07
test/helper.rb: ignore GC compaction on unsupported platforms

terceiro:
For example, ruby 3.0 on Debian ppc64el architecture does not support GC
compaction. When running the tests, every 20th test crashes like this:

> ```
> Error:
> Minitest::Result#test_parsing_attribute_namespace:
> NotImplementedError: Compaction isn't available on this platform
>     <internal:gc>:213:in `compact'
>     /home/terceiro/ruby-nokogiri-1.13.5+dfsg/test/helper.rb:123:in `teardown'
> ```

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -154,6 +154,16 @@ module Nokogiri
       else
         defined?(GC.compact) ? "compact" : "major"
       end
+      if ["compact", "verify"].include?(@@gc_level)
+        # the only way of detecting an unsupported platform is actually
+        # trying GC compaction
+        begin
+          GC.compact
+        rescue NotImplementedError
+          @@gc_level = "normal"
+          warn("#{__FILE__}:#{__LINE__}: GC compaction not suppport by platform")
+        end
+      end
       warn("#{__FILE__}:#{__LINE__}: NOKOGIRI_TEST_GC_LEVEL: #{@@gc_level}")
     end
 


flavorjones:
If you look at the method `initialize_nokogiri_test_gc_level` in `test/helper.rb` you can see that it checks if `GC.compact?` and `GC.verify_compaction_references` are `defined?` before deciding to set the compaction verification. Can you help me understand why that's not working?

flavorjones:
I'd prefer a PR that did the appropriate checks in `initialize_nokogiri_test_gc_level` than wrapped these calls in rescue blocks, if that's possible.

stevecheckoway:
> For that, you would need to actuall call GC.compact at the beginning. I just sent an updated patch, let me know what you think.

This seems like a reasonable approach.

flavorjones:
I agree, this looks good! I've just kicked off CI.

I'm wondering if it would be worth running CI on ppc64le (using qemu like we do for native gem installations) ...

stevecheckoway:
Little endian PowerPC seems like a pretty esoteric platform. Maybe something that runs periodically but not on every commit?

flavorjones:
I'm curious if there's a way I can easily reproduce what you're seeing? I'm trying the `platform=linux/ppc64le` docker container for `ruby:3.0` (and `multiarch/qemu-user-static` to emulate that hardware), but it seems to be working OK (using current `main` at 199434c):

```
warnings: []
nokogiri:
  version: 1.14.0.dev
  cppflags:
  - "-I/root/nokogiri/ext/nokogiri"
  ldflags: []
ruby:
  version: 3.0.4
  platform: powerpc64le-linux
  gem_platform: powerpc64le-linux
  description: ruby 3.0.4p208 (2022-04-12 revision 3fa771dded) [powerpc64le-linux]
  engine: ruby
libxml:
  source: system
  memory_management: ruby
  iconv_enabled: true
  compiled: 2.9.10
  loaded: 2.9.10
libxslt:
  source: system
  datetime_enabled: true
  compiled: 1.1.34
  loaded: 1.1.34
other_libraries:
  libgumbo: 1.0.0-nokogiri

# Running tests with run options --seed 63270:

/root/nokogiri/test/helper.rb:157: NOKOGIRI_TEST_GC_LEVEL: compact
........................................................................................................................................................................................................................................S....................................S........................................S....................................................................................................................................................................................S.......................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................S.................................................SSS.................................................................................................................................S..........................................................S.......S.....S....S................................S.....................S......S.........S........S.......S.............................................................................................................................................................................................................................................................................................................^Z
```

flavorjones:
Ah, never mind, I compiled a custom version of ruby with GC_COMPACTION_SUPPORTED set to 0 and I'm testing it that way.

flavorjones:
@terceiro Thanks for this! Merging.

flavorjones:
Note for posterity, this was reported upstream in https://bugs.ruby-lang.org/issues/18779

Issue #2551, opened 2022-05-17
[bug] HTML5 fragment has extraneous head/body tags


dan42:
```ruby
require "nokogiri" #1.13.6

puts Nokogiri.HTML("x").fragment("hello")
#=> hello

puts Nokogiri.HTML5("x").fragment("hello")
#=> <head></head><body>hello</body>
```
I'm fairly sure that a "fragment" is not supposed to include the `<head>` and `<body>` tags.

stevecheckoway:
Hi @dan42,

Indeed it should not! I'll take a look.

stevecheckoway:
So actually, I misread that initially as `Nokogiri::HTML5.fragment("hello")`.

The [HTML fragment parsing algorithm](https://html.spec.whatwg.org/multipage/parsing.html#parsing-html-fragments) takes an HTML element as input. When we call `Element#fragment()`, the element we're calling it on is the input to the algorithm.

In your example, you're calling it on a Document. I'm not sure what the expected output of that should be. One option would be to treat it as if we'd called it on the body element. This matches what you see using the HTML4 parser.
```ruby
> Nokogiri::HTML5("x").root.children[1].fragment("hello")
=> #(DocumentFragment:0x70b48 { name = "#document-fragment", children = [ #(Text "hello")] })
```

For both the HTML4 and HTML5 `Document#fragment`, we use the root `html` element as the context element.

After giving it some thought, the HTML5 output makes sense to me. If we parse just the string `hello` with the context element set to an `html` element, then on the first character, the parser will create an empty `head` element and a `body` element and then `hello` becomes a text node child of the `body` element.

And indeed, if we ask a browser to create a fragment with the `html` element as the context element, it produces the same output:

```javascript
> document.children[0].innerHTML = "hello"
"hello"
> document.children[0].innerHTML
"<head></head><body>hello</body>" 
```

I don't know why the HTML4 fragment parser doesn't do the same thing, but it's probably a limitation of libxml2.

So the question is what behavior do we want when `#fragment` is called on the document? My inclination is that the current behavior for HTML5 is confusing and should be changed to be as if we called it on a `body` element instead. This is a one line change.

```
diff --git a/lib/nokogiri/html5/document.rb b/lib/nokogiri/html5/document.rb
index d3b69431..e1bb446e 100644
--- a/lib/nokogiri/html5/document.rb
+++ b/lib/nokogiri/html5/document.rb
@@ -70,7 +70,7 @@ module Nokogiri
[...]

PR #2553, opened 2022-05-20
Use `body` as the context element for `Document#fragment`

stevecheckoway:

Fixes: #2551

Yes.

It changes the C HTML5 visible behavior (but not implementation) to match the C HTML4 implementation. I don't know what the Java implementation does.

### DIFF
--- a/lib/nokogiri/html5/document.rb
+++ b/lib/nokogiri/html5/document.rb
@@ -112,9 +112,20 @@ module Nokogiri
         @url = nil
       end
 
-      # Parse a HTML5 document fragment from +tags+, returning a HTML5::DocumentFragment.
-      def fragment(tags = nil)
-        DocumentFragment.new(self, tags, nil)
+      # :call-seq:
+      #   fragment() → Nokogiri::HTML5::DocumentFragment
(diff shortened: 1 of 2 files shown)


flavorjones:
I rebased and added a bit of documentation to `HTML5::Document#fragment` while I had context (no pun intended).

Should go green, I'll merge after that.

lib/nokogiri/html5/document.rb
```diff
@@ -112,9 +112,20 @@ def initialize(*args) # :nodoc:
...
+      # Parse a HTML5 document fragment from +markup+, returning a Nokogiri::HTML5::DocumentFragment.
+      #
+      # [Properties]
+      # - +markup+ (String) The HTML5 markup fragment to be parsed
+      #
+      # [Returns]
+      #   Nokogiri::HTML5::DocumentFragment. This object's children will be empty if `markup` is not passed, is empty, or is `nil`.
+      #
+      def fragment(markup = nil)
+        DocumentFragment.new(self, markup)
```
flavorjones, line 128: Note that I changed this line to not pass in the third argument, rather than pass in an explicit `nil`.

flavorjones (approved):

PR #2556, opened 2022-05-25
test: improve html5lib test format handling

### LINKED ISSUE #2555: Change html5lib parser to accommodate out-of-order test metadata
Regarding https://github.com/html5lib/html5lib-tests/pull/148, one possibility is we could change our parser for these tests to either give a warning when the test format is wrong or just handle out of order sections (or both).

As I recall, the code I wrote to parse these is…not great. I think I just cobbled a parser together that worked and then never revisited it.

flavorjones:

The parser for html5lib tests is conservative in what it accepts, and chooses to abort when formatting expectations aren't met. This PR updates that code to skip the test and emit a warning message.

Fixes #2555 

N/A

N/A


stevecheckoway:
This looks better all around! Thank you for cleaning that code up.

flavorjones:
Also running this against upstream html5lib-tests at https://github.com/sparklemotion/nokogiri/runs/6598806979?check_suite_focus=true

PR #2566, opened 2022-06-05
support HTML5 `search` element

flavorjones:

Upstream the `<search>` element has been adopted (https://github.com/whatwg/html/pull/7320) and the tests were added to `html5lib-tests` (https://github.com/html5lib/html5lib-tests/pull/142).

The upstream tests have been failing on this for a while, so I'm attempting to add support in this PR.

Notable changes:
- I've restored our ability to generate source files from the `.gperf` files
- I've added a new `search` tag
- I've modified the parser to include `search` in the set of tags that close `p` and that should have implicit end tags

I've updated the `test/html5libs-test` ref to the latest that is still green.

The HTML5 parser is only supported in the C extension.


stevecheckoway (approved):
This all looks good to me.

stevecheckoway:
The new search element doesn't yet appear on the web page, but from reading the changes, it looks like it only impacts the parser in the ways your commits handle.

flavorjones:
Thanks for the review! Merging.

PR #2579, opened 2022-06-17
Fix compaction on nodes

### LINKED ISSUE #2578: [bug] Document / Nodes aren't compaction friendly
**Please describe the bug**

`xmlNodePtr` instances keep a reference to their Ruby wrapper inside the `_private` member.  You can see the `_private` member being set to the wrapping Ruby instance [here](https://github.com/sparklemotion/nokogiri/blob/0e75392d49ca3758dcdfc6610c9c22ccdc23001c/ext/nokogiri/xml_node.c#L2048-L2049).

After it's been wrapped, the instance is optionally stored in an array that is kept on the associated document.  You can see the array push [here](https://github.com/sparklemotion/nokogiri/blob/0e75392d49ca3758dcdfc6610c9c22ccdc23001c/ext/nokogiri/xml_node.c#L2053-L2054).

Ruby arrays are compaction friendly.  When the compactor runs, elements inside the array are allowed to move because the array knows how to update its own references.  However, Nokogiri's XML nodes are not compaction friendly.  They don't know to update their references.  Normally this isn't a problem because objects are required to call `rb_gc_mark` on their references, and `rb_gc_mark` will pin the references.  The `xmlNodePtr` instance references the Ruby wrapper, but since the Ruby wrapper is allowed to move, the `_private` pointer can go bad when compaction runs.

```ruby
require "nokogiri"

big_doc = "<root>#{("a".."zz").map { |x| "<#{x}>#{x}</#{x}>" }.join}</root>"
doc = Nokogiri.XML(big_doc)
doc.root.children.each do |child|
  child
end

GC.verify_compaction_references(double_heap: true, toward: :empty)

doc.root.children.each do |child|
  p child
end
```

This script will SEGV because the node objects move, and then we try to reference them.

**Expected behavior**

It shouldn't segv.

**Possible Fixes**

I'm not 100% sure how we should go about fixing this.  This patch fixes the above code:

```diff
diff --git a/ext/nokogiri/xml_document.c b/ext/nokogiri/xml_document.c
index 134931de..31ddc24d 100644
--- a/ext/nokogiri/xml_document.c
+++ b/ext/nokogiri/xml_document.c
@@ -57,6 +57,9 @@ mark(xmlDocPtr doc)
   if (tuple) {
     rb_gc_mark(tuple->doc);
     rb_gc_mark(tuple->node_cache);
+    for (long i = 0; i < RARRAY_LEN(tuple->node_cache); i++) {
+        rb_gc_mark(rb_ary_entry(tuple->node_cache, i));
+    }
   }
 }
 
```

But I don't like this patch.  It means we iterate the array too many times.

I think a better solution is probably to teach Node instances about compaction so that they can update themselves, but that might be a more involved patch.

tenderlove:
This PR is an attempt to fix compaction issues from #2578.  We've split the PR in to 3 commits.

The first commit consistently sets a mark function on the xml node.  We need this because we need to provide a static struct to `TypedData_Wrap_Struct`.

The second commit introduces a helper function for unwrapping xml nodes.  Unfortunately, xml nodes use a union type, and subclasses want to unwrap the type in to different things, so we had to extract a function that would unwrap the nodes correctly.

The third commit actually fixes compaction by adding a callback for updating references.

Fixes #2578

### DIFF
--- a/ext/nokogiri/xml_node.c
+++ b/ext/nokogiri/xml_node.c
@@ -7,7 +7,6 @@ static ID id_decorate, id_decorate_bang;
 
 typedef xmlNodePtr(*pivot_reparentee_func)(xmlNodePtr, xmlNodePtr);
 
-
 #ifdef DEBUG
 static void
 _xml_node_dealloc(xmlNodePtr x)
@@ -36,6 +35,14 @@ _xml_node_mark(xmlNodePtr node)
   }
 }
 
+static const rb_data_type_t nokogiri_node_type = {
+    "Nokogiri/XMLNode",
+    {_xml_node_mark, _xml_node_dealloc, 0,},
+    0, 0,
+#ifdef RUBY_TYPED_FREE_IMMEDIATELY
+    RUBY_TYPED_FREE_IMMEDIATELY,
+#endif
+};
 
 static void
 relink_namespace(xmlNodePtr reparented)
@@ -201,8 +208,8 @@ reparent_node_with(VALUE pivot_obj, VALUE reparentee_obj, pivot_reparentee_func 
     rb_raise(rb_eArgError, "node must be a Nokogiri::XML::Node");
   }
 
-  Data_Get_Struct(reparentee_obj, xmlNode, reparentee);
-  Data_Get_Struct(pivot_obj, xmlNode, pivot);
+  Noko_Node_Get_Struct(reparentee_obj, xmlNode, reparentee);
+  Noko_Node_Get_Struct(pivot_obj, xmlNode, pivot);
 
   /*
    * Check if nodes given are appropriate to have a parent-child
@@ -442,7 +449,7 @@ rb_xml_node_add_namespace_definition(VALUE rb_node, VALUE rb_prefix, VALUE rb_hr
   xmlNsPtr c_namespace;
   const xmlChar *c_prefix = (const xmlChar *)(NIL_P(rb_prefix) ? NULL : StringValueCStr(rb_prefix));
 
-  Data_Get_Struct(rb_node, xmlNode, c_node);
+  Noko_Node_Get_Struct(rb_node, xmlNode, c_node);
   element = c_node ;
 
   c_namespace = xmlSearchNs(c_node->doc, c_node, c_prefix);
@@ -509,7 +516,7 @@ rb_xml_node_attribute(VALUE self, VALUE name)
 {
   xmlNodePtr node;
   xmlAttrPtr prop;
-  Data_Get_Struct(self, xmlNode, node);
+  Noko_Node_Get_Struct(self, xmlNode, node);
   prop = xmlHasProp(node, (xmlChar *)StringValueCStr(name));
 
   if (! prop) { return Qnil; }
@@ -560,7 +567,7 @@ rb_xml_node_attribute_nodes(VALUE rb_node)
 {
   xmlNodePtr c_node;
 
-  Data_Get_Struct(rb_node, xmlNode, c_node);
+  Noko_Node_Get_Struct(rb_node, xmlNode, c_node);
 
   return noko_xml_node_attrs(c_node);
 }
@@ -612,7 +619,7 @@ rb_xml_node_attribute_with_ns(VALUE self, VALUE name, VALUE namespace)
 {
   xmlNodePtr node;
   xmlAttrPtr prop;
-  Data_Get_Struct(self, xmlNode, node);
+  Noko_Node_Get_Struct(self, xmlNode, node);
   prop = xmlHasNsProp(node, (xmlChar *)StringValueCStr(name),
                       NIL_P(namespace) ? NULL : (xmlChar *)StringValueCStr(namespace));
 
@@ -639,7 +646,7 @@ static VALUE
 rb_xml_node_blank_eh(VALUE self)
 {
   xmlNodePtr node;
-  Data_Get_Struct(self, xmlNode, node);
+  Noko_Node_Get_Struct(self, xmlNode, node);
   return (1 == xmlIsBlankNode(node)) ? Qtrue : Qfalse ;
 }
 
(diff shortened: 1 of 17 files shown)


flavorjones:
Note: this should probably be backported to v1.13.x in a patch release. I'll do that in a separate PR once this is on `main`.

ext/nokogiri/xml_node.c
```diff
@@ -33,6 +43,30 @@ _xml_node_mark(xmlNodePtr node)
...
+}
+#endif
+
+typedef void (*gc_callback_t)(void *);
+
+static const rb_data_type_t nokogiri_node_type = {
+    "Nokogiri/XMLNode",
+    {(gc_callback_t)_xml_node_mark, (gc_callback_t)_xml_node_dealloc, 0,
+#ifdef HAVE_RB_GC_LOCATION
+        (gc_callback_t)_xml_node_update_references
```
stevecheckoway, line 62: I think this code has undefined behavior. Although you can cast between incompatible function pointers, you can't legally call functions through them. C99, Annex J's list of undefined behavior includes
> A pointer is used to call a function whose type is not compatible with the pointed-to type (6.3.2.3).

And indeed §6.3.2.3 paragraph 8 says
> A pointer to a function of one type may be converted to a pointer to a function of another type and back again; the result shall compare equal to the original pointer. If a converted pointer is used to call a function whose type is not compatible with the pointed-to type, the behavior is undefined.

I believe all of `_xml_node_mark`, `_xml_node_dealloc`, and `_xml_node_update_references` need to take a `void *` argument and then assign the argument to an `xmlNodePtr` local.

ext/nokogiri/xml_text.c
```diff
@@ -20,7 +20,7 @@ new (int argc, VALUE *argv, VALUE klass)
 
   rb_scan_args(argc, argv, "2*", &string, &document, &rest);
 
-  Data_Get_Struct(document, xmlDoc, doc);
+  doc = (xmlDocPtr)noko_xml_node_unwrap(document);
```
stevecheckoway, line 23: `(xmlDocPtr)noko_xml_node_unwrap(document)` happens often enough that I think we should have a `noko_xml_document_unwrap()` although I don't feel super strongly about it.

stevecheckoway (changes requested):

ext/nokogiri/xml_text.c
flavorjones: Yeah, that jumped out at me, too. I may try some variations on this this weekend. I may also try using a macro with the same parameters as `Data_Get_Struct`.

ext/nokogiri/xml_node.c
tenderlove: I think the function types are compatible though?  They all take a single pointer as the parameter (IOW the width of a `void *` is the same width as an `xmlNodePtr` so there shouldn't be any problem wrt C calling convention).  I don't have any strong feelings wrt taking a `void *` and casting in the function body, but I don't think this behavior is undefined.

stevecheckoway: The types are not compatible according to C. It's undefined in exactly the same way [this is](https://stackoverflow.com/questions/2561697/warning-when-using-qsort-in-c). The problem is that [function pointer types have to have compatible arguments](https://en.cppreference.com/w/c/language/type#Compatible_types) and `void *` is not compatible with `xmlNodePtr`.

In fact, C doesn't mandate that `void *` and `xmlNodePtr` be the same size; although they probably are on any platform on which Ruby runs.

stevecheckoway: You shouldn't need to cast the `void *` to assign it to another pointer though.

tenderlove: Should be fixed in 0fd5321931c80250743d322381eb571a90a7f926!

flavorjones: Thanks!

flavorjones:
Sorry for being slow to merge this. I will work on it (and the backport?) this week.

stevecheckoway (approved):
This looks good. You should even be able to get rid of the casts, but they're not doing any harm.

**Edit:** I meant the approval to mean the updated callback signatures looked good. I did notice one of the tests shows a memory leak reported by Valgrind.

flavorjones:
I rebased and made some minor changes.

flavorjones:
I've added a commit 1c74ba1, which I would like to squash, which uses a macro `Noko_Node_Get_Struct` to replace the previous `Data_Get_Struct` macro and `noko_xml_node_unwrap()` function.

This results in a smaller diff, since `Noko_Node_Get_Struct` has the same parameters as `Data_Get_Struct`. It should apply cleanly to the v1.13.x branch when I backport.

The resulting diff is here: https://github.com/sparklemotion/nokogiri/compare/b370fbd290343e93f20e6bf12522c5ba3f2f5caf..1c74ba18fccc4e809d62c62f2f8ee536aae3ee94

@tenderlove @eightbitraptor would love a quick review when you have a moment.

flavorjones:
Added one more commit onto the end to remove `RBIMPL_CAST` which blew up on windows.

flavorjones:
@tenderlove I've got one last question for you, which is "why are we using `DATA_PTR` instead of `TypedData_Get_Struct`?" 

I _think_ I know that answer (but would like confirmation): the alternative of using `TypedData_Get_Struct` would require creating a new `rb_data_type_t` for each of the "subclasses" of `struct xmlNode`, and that's a lot of code for little gain; and `DATA_PTR` has the advantage of being much less code, and the lack of type safety is OK because libxml2 seems to take care of this reasonably well.

Does that sound about right?

flavorjones:
The memcheck failure is because `ruby_memcheck` needs to be told to ignore memory allocated by `rb_ary_push`, will take care of that tomorrow.

tenderlove:
> Does that sound about right?

Yes, about.

The problem I found is that since we have so many objects that inherit from `Node`.  In order to unwrap the object in a method on node, we'd have to check the type of `self` on every call in order to understand which struct we need to pass to `TypedData_Get_Struct`.

I'd prefer to use the "typed" unwrappers everywhere, but I couldn't figure out a "proper" way to do that with C objects that use inheritance (without adding tons of checks).

flavorjones:
@tenderlove Thanks. This is going to go green so I'm going to clean up commit history (squash some of my changes) and let it go through CI one more time.

PR #2584, opened 2022-06-23
fix: HTML5::Document#url returns the URL passed to the parser

### LINKED ISSUE #2583: [bug] HTML5 parser loses reference to `url` parameter value
**Please describe the bug**

Invocations of `Nokogiri::HTML4()` and `Nokogiri::HTML5()` both support optional second parameters representing the parsed document's URL:

- https://github.com/sparklemotion/nokogiri/blob/main/lib/nokogiri/html4.rb#L6-L12
- https://github.com/sparklemotion/nokogiri/blob/main/lib/nokogiri/html5.rb#L25-L32

The resulting `Nokogiri::HTMLx::Document` instance responds to a `url` method that _should_ return the string value passed to `Nokogiri::HTML4`/`Nokogiri::HTML5`. This works with the HTML4 parser ([alternatively invoked](https://github.com/sparklemotion/nokogiri/blob/main/lib/nokogiri/html.rb#L7-L8) as `Nokogiri::HTML()`) but the HTML5 parser consistently returns `nil`.

I did my best to trace the code path from `Nokogiri::HTML5()` downward, but I got lost in `Nokogiri::HTML5::Document`'s private `do_parse` method ([source](https://github.com/sparklemotion/nokogiri/blob/main/lib/nokogiri/html5/document.rb#L61-L69)):

```ruby
def do_parse(string_or_io, url, encoding, options)
  string = HTML5.read_and_encode(string_or_io, encoding)
  max_attributes = options[:max_attributes] || Nokogiri::Gumbo::DEFAULT_MAX_ATTRIBUTES
  max_errors = options[:max_errors] || options[:max_parse_errors] || Nokogiri::Gumbo::DEFAULT_MAX_ERRORS
  max_depth = options[:max_tree_depth] || Nokogiri::Gumbo::DEFAULT_MAX_TREE_DEPTH
  doc = Nokogiri::Gumbo.parse(string, url, max_attributes, max_errors, max_depth, self)
  doc.encoding = "UTF-8"
  doc
end
```

Specifically, I couldn't locate source for the `Nokogiri::Gumbo.parse` call on [line 66](https://github.com/sparklemotion/nokogiri/blob/main/lib/nokogiri/html5/document.rb#L66). I suspect this code may be generated from C code which is beyond my level of understanding. Is that the case? Did I miss something?

Save the following script as `test.rb` (or similar) and run via `ruby test.rb`.

```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'

  gem 'nokogiri', '~> 1.13'
end

url = 'https://hello.example'
markup = '<body>Hello, world!</body>'

puts 'Testing Nokogiri::HTML...'
puts Nokogiri::HTML(markup, url).url
# => https://hello.example

puts 'Testing Nokogiri::HTML4...'
puts Nokogiri::HTML4(markup, url).url
# => https://hello.example

puts 'Testing Nokogiri::HTML5...'
puts Nokogiri::HTML5(markup, url).url
# => nil
```

**Expected behavior**

I expected that a call to `Nokogiri::HTML5(markup, url).url` would return the value passed as the second parameter, consistent with the HTML4 parser.

**Environment**

```
# Nokogiri (1.13.6)
    ---
    warnings: []
    nokogiri:
      version: 1.13.6
      cppflags:
      - "-I/Users/jason/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/nokogiri-1.13.6-arm64-darwin/ext/nokogiri"
      - "-I/Users/jason/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/nokogiri-1.13.6-arm64-darwin/ext/nokogiri/include"
      - "-I/Users/jason/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/nokogiri-1.13.6-arm64-darwin/ext/nokogiri/include/libxml2"
      ldflags: []
    ruby:
      version: 2.7.6
      platform: arm64-darwin21
      gem_platform: arm64-darwin-21
      description: ruby 2.7.6p219 (2022-04-12 revision c9c2245c0a) [arm64-darwin21]
      engine: ruby
    libxml:
      source: packaged
      precompiled: true
      patches:
      - 0001-Remove-script-macro-support.patch
      - 0002-Update-entities-to-remove-handling-of-ssi.patch
      - 0003-libxml2.la-is-in-top_builddir.patch
      - 0004-use-glibc-strlen.patch
      - 0005-avoid-isnan-isinf.patch
      - 0006-update-automake-files-for-arm64.patch
      - '0008-htmlParseComment-handle-abruptly-closed-comments.patch'
      - '0009-allow-wildcard-namespaces.patch'
      libxml2_path: "/Users/jason/.rbenv/versions/2.7.6/lib/ruby/gems/2.7.0/gems/nokogiri-1.13.6-arm64-darwin/ext/nokogiri"
      memory_management: ruby
      iconv_enabled: true
      compiled: 2.9.14
      loaded: 2.9.14
    libxslt:
      source: packaged
      precompiled: true
      patches:
      - 0001-update-automake-files-for-arm64.patch
      datetime_enabled: true
      compiled: 1.1.35
      loaded: 1.1.35
    other_libraries:
      zlib: 1.2.12
      libiconv: '1.16'
      libgumbo: 1.0.0-nokogiri
```

**Additional context**

I'm currently using the HTML4 parser in some code and relying on the `url` parameter to resolve relative URLs in parsed markup. I discovered the above documented inconsistency while updating that code to use the newer HTML5 parser.

Thanks for reading and I appreciate your help!

flavorjones:

Closes #2583
    
HTML5::Document#url now matches the behavior of HTML4::Document and XML::Document. Previously this method always returned `nil`.

I've also improved the documentation for HTML5::Document.

Yes!

This affects the behavior of HTML5::Document which only exists in the C implementation today.

### DIFF
--- a/ext/nokogiri/gumbo.c
+++ b/ext/nokogiri/gumbo.c
@@ -360,6 +360,7 @@ parse_continue(VALUE parse_args)
   args->doc = doc; // Make sure doc gets cleaned up if an error is thrown.
   build_tree(doc, (xmlNodePtr)doc, output->document);
   VALUE rdoc = noko_xml_document_wrap(args->klass, doc);
+  rb_iv_set(rdoc, "@url", args->url_or_frag);
   args->doc = NULL; // The Ruby runtime now owns doc so don't delete it.
   add_errors(output, rdoc, args->input, args->url_or_frag);
   return rdoc;
--- a/lib/nokogiri/html5/document.rb
+++ b/lib/nokogiri/html5/document.rb
@@ -25,9 +25,39 @@ module Nokogiri
     #
     # 💡 HTML5 functionality is not available when running JRuby.
     class Document < Nokogiri::HTML4::Document
+      # Get the url name for this document, as passed into Document.parse, Document.read_io, or
+      # Document.read_memory
       attr_reader :url
 
       class << self
+        # :call-seq:
+        #   parse(input)
+        #   parse(input, url=nil, encoding=nil, **options)
+        #   parse(input, url=nil, encoding=nil) { |options| ... }
+        #
+        # Parse HTML5 input.
+        #
+        # [Parameters]
+        # - +input+ may be a String, or any object that responds to _read_ and _close_ such as an
+        #   IO, or StringIO.
+        #
+        # - +url+ (optional) is a String indicating the canonical URI where this document is located.
+        #
+        # - +encoding+ (optional) is the encoding that should be used when processing
+        #   the document.
+        #
+        # - +options+ (optional) is a configuration Hash (or keyword arguments) to set options
+        #   during parsing. The three currently supported options are +:max_errors+,
+        #   +:max_tree_depth+ and +:max_attributes+, described at Nokogiri::HTML5.
+        #
+        #   ⚠ Note that these options are different than those made available by
+        #   Nokogiri::XML::Document and Nokogiri::HTML4::Document.
+        #
+        # - +block+ (optional) is passed a configuration Hash on which parse options may be set. See
+        #   Nokogiri::HTML5 for more information and usage.
+        #
+        # [Returns] Nokogiri::HTML5::Document
+        #
         def parse(string_or_io, url = nil, encoding = nil, **options, &block)
           yield options if block
           string_or_io = "" unless string_or_io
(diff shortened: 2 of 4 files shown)


jgarber623 (approved):
Tested locally with the same script outlined in #2583 (but with the gem source pointing at this PR's branch) and everything works as expected!

```sh
$ ruby test.rb
Testing Nokogiri::HTML...
https://hello.example
Testing Nokogiri::HTML4...
https://hello.example
Testing Nokogiri::HTML5...
https://hello.example
```

Thanks for the quick fix! 🎉

PR #2589, opened 2022-07-10
ext: rename the extension file to `nokogiri_ext`

flavorjones:

I keep running into problems with some Ruby 3.1 environments
where the extension is loaded when I `require "nokogiri"` because of
the file name collision.

I imagine this is a path problem of some kind, and so I'm hesitant to fix it here, but I did at least want to have this ready.

Yes, existing tests.

No behavioral changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,11 +7,11 @@
 /ext/nokogiri/**/nokogiri.dll
 /ext/nokogiri/include
 /gems/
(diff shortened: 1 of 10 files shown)


flavorjones:
I can't reproduce the original problem in Ruby 3.1.2 (though I can in my Ruby 3.1.0 environment).

Issue #2598, opened 2022-07-20
bug: v1.13.7 XML::Reader may segfault


flavorjones:
Originally reported downstream in https://github.com/pythonicrubyist/creek/issues/105

Special thanks to @bf4 who helped reliably reproduce this.

More details coming shortly.

flavorjones:
## Summary and Diagnosis

This comment is an explanation of the problem. I'll follow up with a few suggestions for approaches to fix it.


### The Symptoms

Users of `Nokogiri::XML::Reader` experiencing segfaults in Nokogiri v1.13.7 while traversing a document.

A reproduction (provided by @bf4) captured in valgrind showed the following memory error:

``` text
==721789== Invalid read of size 8
==721789==    at 0xBAF3D44: _xml_node_mark (xml_node.c:24)
==721789==    by 0x49474AE: gc_mark_children (gc.c:6344)
==721789==    by 0x494871C: gc_mark_stacked_objects (gc.c:6448)
==721789==    by 0x494871C: gc_mark_stacked_objects_all (gc.c:6488)
==721789==    by 0x494871C: gc_marks_rest (gc.c:7429)
==721789==    by 0x4949007: gc_marks (gc.c:7485)
==721789==    by 0x4949007: gc_start (gc.c:8334)
...
```

Nokogiri v1.13.7 introduced some changes in how Nodes behave during garbage collection, so this stack isn't _terribly_ surprising. But the changes were relatively straightforward and this _is_ rather unexpected.

It also happens only rarely, so it's challenging to reproduce and didn't show up in any of the valgrind jobs in our CI pipeline.


### How Reader's memory lifecycle works

Nodes associated with an `XML::Reader` don't have a fully-realized document. They have an associated `xmlDoc` struct (pointed to in the `xmlNode->doc` struct member), but the libxml Reader API doesn't expose it, and Nokogiri doesn't expose it, and so it never gets wrapped in a Nokogiri::XML::Document ruby object.

To demonstrate, if we add some debugging output to `xml_node.c:noko_xml_node_wrap()` as follows:

``` patch
--- a/ext/nokogiri/xml_node.c
+++ b/ext/nokogiri/xml_node.c
@@ -2069,6 +2069,18 @@ noko_xml_node_wrap(VALUE rb_class, xmlNodePtr c_node)
   rb_node = TypedData_Wrap_Struct(rb_class, &nokogiri_node_type, c_node) ;
   c_node->_private = (void *)rb_node;
 
[...]

PR #2599, opened 2022-07-21
fix: XML::Reader XML::Attr garbage collection

flavorjones:

This is a proposed fix for #2598, see that issue for an extended explanation of the problem.

This PR implements "option 2" from that issue's proposed solutions:

- introduce a new `Reader#attribute_hash` that will return a `Hash<String ⇒ String>` (instead of an `Array<XML::Attr>`)
- deprecate `Reader#attribute_nodes` with a plan to remove it entirely in a future release
- re-implement `Reader#attributes` to use `#attribute_hash` (instead of `#attribute_nodes`)

After this change, only applications calling `Reader#attribute_nodes` directly will be running the unsafe code. These users will see a deprecation warning and may use `#attribute_hash` as a replacement.

I think it's very possible that `Reader#attribute_hash` won't meet the needs of people who are working with namespaced attributes and are using `#attribute_nodes` for this purpose. However, I'm intentionally deferring any attempt to solve that DX problem until someone who needs this functionality asks for it.

I tried and failed to add test coverage to the suite that would reproduce the underlying GC bug.

However, existing test coverage of `Reader#attributes` is sufficient for now.

This PR modifies both the C and Java implementations to behave the same.

Notably, the Java implementation contains a small bugfix which is that `Reader#namespaces` now returns an empty hash when there are no namespaces (it previously returned `nil`).

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -42,6 +42,9 @@ jobs:
           - url: https://github.com/rails/rails
             name: xmlmini
             command: "cd activesupport && bundle exec rake test TESTOPTS=-n/XmlMini/"
+          - url: https://github.com/pythonicrubyist/creek
+            name: creek
+            command: "bundle exec rake spec"
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/sparklemotion/nokogiri-test:mri-3.1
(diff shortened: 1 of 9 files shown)


ccutrer:
> However, I'm intentionally deferring any attempt to solve that DX problem until someone who needs this functionality asks for it.

:sigh: here I am. https://github.com/instructure/ratom/blob/v0.10.11/lib/atom/xml/parser.rb#L103. I'm not the original author of this code, so I'm not intimately familiar with it. But it sure seems like proper namespace handling is imperative for this code. It needs the attribute name in all forms: with the namespace prefix, and with the namespace href. It doesn't hang on to the attribute node though, so copying out just those values and not having a full xmlAttrPtr is just fine.

flavorjones:
@ccutrer Thanks for the note. I created a new issue, let's continue the conversation at https://github.com/sparklemotion/nokogiri/issues/3102!

Issue #2621, opened 2022-08-08
Investigate upstream libxml2 CI failures related to improperly-opened comment parsing


flavorjones:
This commit upstream:

- https://gitlab.gnome.org/GNOME/libxml2/-/commit/e986d09cf531e77a9ab46af2d1a219072f310190

is likely causing "upstream" CI failures:

- https://github.com/sparklemotion/nokogiri/runs/7647412154?check_suite_focus=true#step:8:4198

PR #2624, opened 2022-08-15
ci: upstream libxml2 fixed conditional html comments

flavorjones:

Upstream libxml2 changed how it handles conditional HTML comments, see https://gitlab.gnome.org/GNOME/libxml2/-/issues/380

Closes https://github.com/sparklemotion/nokogiri/issues/2621

Update the test suite to handle this and explain what's going on a bit. Especially since this behavior changed in 2.9.14 and is essentially changing back.

Yes

No behavioral changes other than handling the changes in upstream libxml2.

### DIFF
--- a/test/html4/test_comments.rb
+++ b/test/html4/test_comments.rb
@@ -173,18 +173,8 @@ module Nokogiri
         let(:body) { doc.at_css("body") }
         let(:subject) { doc.at_css("div#under-test") }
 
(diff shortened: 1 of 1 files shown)


flavorjones:
Also manually kicked off upstream CI at https://github.com/sparklemotion/nokogiri/actions/runs/2861667512

PR #2631, opened 2022-08-23
ci: add asciidoctor to downstream CI

flavorjones:

See https://github.com/asciidoctor/asciidoctor/pull/4339 for the issue that caused me to add this.


flavorjones:
Seems like this is only a dev dependency of asciidoctor, so I'm not going to add to CI

Issue #1829, opened 2018-12-03
Explore replacements for Xalan-J


flavorjones:
See #1760 for more context.

The JRuby implementation of Nokogiri uses [Xalan-J](https://xml.apache.org/xalan-j/) for XSLT support, but [this thread](https://mail-archives.apache.org/mod_mbox/xalan-dev/201810.mbox/browser) reveals that development (and likely support) has stalled on that project.

A potential replacement, noted in the thread linked above, is [Saxon](http://saxon.sourceforge.net/). There may be other potential replacements as well.

The outcome intended for this issue is to do some exploration of replacing Xalan-J:

* is there a replacement that will functionally support what Xalan-J supports today?
* how difficult will a transition be?

Tagging @jvshahid and @kares for visibility

headius:
I don't believe there's been much forward momentum on this, but another motivator to make a move came up recently in the form of a CVE for Xalan:

https://github.com/sparklemotion/nokogiri/security/advisories/GHSA-qwq9-89rg-ww72

The comment here asks whether we might be able to transition (at least) to the shipping Xalan in OpenJDK, to which the answer is "probably": https://github.com/sparklemotion/nokogiri/security/advisories/GHSA-qwq9-89rg-ww72#advisory-comment-74545

In any case this should probably be addressed sooner than later since Xalan-J has been EOL for many years now, and is only being maintained in JDK as part of JDK. Saxon may be the best path forward, but using the JDK Xalan may be a short-term way to avoid the super-stale external version.

flavorjones:
@headius Thanks for commenting. I may need help with the mechanics of switching to the OpenJDK version of Xalan. I've got a branch where I've updated the package paths, but there are a few places where the API has changed that I need to work through. I'll push it shortly.

flavorjones:
I've created a draft PR migrating to the JDK Xalan that needs help to get unstuck: https://github.com/sparklemotion/nokogiri/pull/2632

kares:
> Saxon may be the best path forward, but using the JDK Xalan may be a short-term way to avoid the super-stale external version.

@headius so the only downside is we will need to add-opens the internal `com.sun.org.apache` package?
not sure if this is smt for JDK 17+, whether Nokogiri would be able to rely on those in the long-term ...

flavorjones:
#2632 is a PR that @kares and I collaborated on to try to move Nokogiri away from upstream Xalan to using the JDK internal packages.

[...]

PR #2632, opened 2022-08-23
jruby: Try to use JDK version of Xalan

flavorjones:

See https://github.com/sparklemotion/nokogiri/security/advisories/GHSA-qwq9-89rg-ww72 for documentation on a vulnerability in Xalan, and a note that upstream Xalan is no longer maintained.

See https://github.com/sparklemotion/nokogiri/issues/1829 for a conversation about alternatives to Xalan

This PR attempts to use the JDK's vendored version of Xalan, with only limited success ...

**The Good**

On a positive note, the majority of the changes are simply updating the package paths.

**The Bad**

However, I've discovered that OpenJDK pulled in Xalan 2.6.0 (see https://github.com/openjdk/jdk/commit/74f4779) which **predates** the version of Xalan used to implement Nokogiri and so now I'm a little worried about whether JDKs will all have the same API? Over time this detail has emerged in issues like https://github.com/sparklemotion/nokogiri/issues/1114 and commits like c5ad35a and 6cf2b6e.

I worked around this assuming OpenJDK's API, but I'm worried about portability.

**The Ugly**

Finally, it looks like the Xalan imported into OpenJDK is incomplete and does not contain the class equivalent for Xalan's `./src/org/apache/xalan/transformer/TransformerImpl.java`. In this PR I've used

```java
import com.sun.org.apache.xalan.internal.xsltc.trax.TransformerImpl;
```

which is incorrect and results in this compiler error:

```
ext/java/nokogiri/XsltStylesheet.java:188: error: cannot find symbol
    SerializationHandler serializationHandler = transformer.createSerializationHandler(streamResult);
                                                           ^
  symbol:   method createSerializationHandler(StreamResult)
  location: variable transformer of type TransformerImpl
```

`com.sun.org.apache.xalan.internal.xsltc.trax.TransformerImpl` is the analog for Xalan's `./src/org/apache/xalan/xsltc/trax/TransformerImpl.java` which seems to be a very different class despite similar names. I'm not familiar enough with this library to completely understand how these classes are related, and I need help resolving this issue.

Tagging @headius for assistance!

Existing test coverage should prevent any regressions.

Ideally there will be no behavioral changes, the test suite should help us stay accountable.


kares:
took a quick look and here's smt that might unblock this (the **ugly part**): https://github.com/sparklemotion/nokogiri/pull/2644

kares:
for the record, after merging with #2644 I am seeing: **4 failures, 3 errors** (using OpenJDK 11)
... the errors are related to the part changed as part of #2644 (which is likely still be fixable)
```
Error:
Nokogiri::XSLT::Stylesheet#test_transform:
Java::ComSunOrgApacheXmlInternalSerializerUtils::WrappedRuntimeException: The Properties object passed to the SerializerFactory does not have a 'method' property.
    java.xml/com.sun.org.apache.xml.internal.serializer.SerializerFactory.getSerializer(SerializerFactory.java:172)
    nokogiri.XsltStylesheet.serialize(XsltStylesheet.java:188)
    nokogiri.XsltStylesheet$INVOKER$i$1$0$serialize_DBG.call(XsltStylesheet$INVOKER$i$1$0$serialize_DBG.gen)

```

however, on Java 17 the whole thing bumps into accessibility issues due having to open the internals we're trying to use from the JDK:
```
Java::JavaLang::IllegalAccessError: superclass access check failed: class nokogiri.internals.XalanDTMManagerPatch (in unnamed module @0x41d20f06) cannot access class com.sun.org.apache.xml.internal.dtm.ref.DTMManagerDefault (in module java.xml) because module java.xml does not export com.sun.org.apache.xml.internal.dtm.ref to unnamed module @0x41d20f06
    java.base/java.lang.ClassLoader.defineClass1(Native Method)
    java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1012)
    java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150)
    java.base/java.net.URLClassLoader.defineClass(URLClassLoader.java:524)
    java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:427)
    java.base/java.net.URLClassLoader$1.run(URLClassLoader.java:421)
    java.base/java.security.AccessController.doPrivileged(AccessController.java:712)
    java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:420)
    java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:587)
    java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520)
    nokogiri.XmlXpathContext.<clinit>(XmlXpathContext.java:51)
    nokogiri.XmlXpathContext$INVOKER$s$1$0$rbNew_DBG.call(XmlXpathContext$INVOKER$s$1$0$rbNew_DBG.gen)

```
all in all the change to the internal xalan would need more work but at the same time I am not sure it's worth the effort, given the reliance on internals, which the JDK does not want us doing (also makes the change a breaking change to existing users - they would need to adjust JRUBY_OPTS flags to get it working).

flavorjones:
I just pushed the merged version (rebased onto #2644 which I just merged) but I hear and understand what you're saying about using the internals. We can take this knowledge back to #1829 and continue the conversation about alternatives to Xalan.

flavorjones:
Just a reminder: don't delete this branch, it might be useful to refer back to in the future.

PR #2640, opened 2022-08-29
dep: update libxml2 to v2.10.2 and libxslt to v1.1.37

flavorjones:

See https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.10.2 and https://gitlab.gnome.org/GNOME/libxslt/-/releases/v1.1.37

Note the commit updating libxml2 to automake 1.16.3 which allows us to
finally remove the 0006 patch for arm64 builds:

> https://gitlab.gnome.org/GNOME/libxml2/-/commit/ed80e8c9d8fbeb509f7bbbd3f58deb3e09be32f2

Also note that the workaround applied in 1ab64d2 to work around symbol exports in libxslt 1.1.36 was upstreamed, so that workaround has been reverted.

Current tests are fine for regression-testing.

Upstream caveats always apply.


flavorjones:
Kicked CI

PR #2644, opened 2022-09-01
[jruby] avoid relying on transformer internals

kares:

Should unblock compilation issue with older Xalan versions (https://github.com/sparklemotion/nokogiri/pull/2632).

Using the `TransformerImpl` (internal) class from Xalan is avoided.

Existing tests cover the serialization feature.

JRuby only and should be considered a simple refactoring change - wout side-effects.


flavorjones:
Nice! Thank you, I'll take a look

flavorjones:
I want to apologize for not getting to this in the past week -- I'm very thankful that you did this work. Thanks for your patience.

Issue #1582, opened 2017-01-11
libxml 2.9.{4,5,...} is vulnerable to CVE-2016-9318


flavorjones:
Credit for reporting this vulnerability to Nokogiri Core goes to Danny Grander <danny@snyk.io>, @grnd. Thanks, Danny!

This issue is being opened for nokogiri core to triage this vulnerability within the context of Nokogiri.

---

**Triage summary**:

Applications using Nokogiri 1.5.4 (circa June 2012) and later are **not vulnerable** to this CVE unless opting into the `DTDLOAD` option *and* opting out of the `NONET` option.

Note that by default, Nokogiri turns off both DTD loading and network access, treating all docs as untrusted docs. This behavior is independent of the version of libxml2 used.

Also note that an upstream fix is not available from libxml2 upstream as of 2017-11-13.

flavorjones:
# CVE-2016-9318 resources

For the libxml2 vulnerability:

* Mitre: https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-9318
* NVD:
  * https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2016-9318
  * CVSS v3 base score: [7.8 High](https://nvd.nist.gov/cvss/v3-calculator?name=CVE-2016-9318&vector=AV:L/AC:L/PR:N/UI:R/S:U/C:H/I:H/A:H)
* Security Focus: http://www.securityfocus.com/bid/94347
* RedHat:
  * https://access.redhat.com/security/cve/cve-2016-9318
  * CVSS v3 base score: [7.1](https://www.first.org/cvss/calculator/3.0#CVSS:3.0/AV:N/AC:L/PR:N/UI:R/S:C/C:L/I:L/A:L)
  * CWE-611 [Improper Restriction of XML External Entity Reference ('XXE')](https://cwe.mitre.org/data/definitions/611.html)
  * bugzilla: https://bugzilla.redhat.com/show_bug.cgi?id=1395609
* Upstream
  * libxml2: https://bugzilla.gnome.org/show_bug.cgi?id=772726
  * xmlsec: https://github.com/lsh123/xmlsec/issues/43

flavorjones:
## attempt 1 to exploit: default Nokogiri parse options


In a console window, start up netcat listening on port 8080:

```bash
$ nc -l 0.0.0.0 8080
```

Then, in a second console windows, run this Ruby script:

```ruby
#! /usr/bin/env ruby

[...]

PR #2096, opened 2020-10-12
Clean up error handling in the C extension.

flavorjones:
This PR is for a set of accumulated changes in the C extension around libxml2 error handling.

I'd like to, whenever possible, avoid raising an exception from a native C callback. This is primarily to do with preventing memory leaks as detailed in #1610, but also to allow TruffleRuby to unwind the stack properly as detailed in #1882. I'm creating it as a draft so that I can get incremental feedback from CI as I go.

Issue #1610, opened 2017-03-01
Undefined namespace prefix error results in memory leak


pcapcanari:
Memory leak is observed while searching in a document with undefined namespace prefix. 
```text
# Nokogiri (1.7.0.1)
    ---
    warnings: []
    nokogiri: 1.7.0.1
    ruby:
      version: 2.3.1
      platform: x86_64-linux
      description: ruby 2.3.1p112 (2016-04-26 revision 54768) [x86_64-linux]
      engine: ruby
    libxml:
      binding: extension
      source: packaged
      libxml2_path: "/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/nokogiri-1.7.0.1/ports/x86_64-pc-linux-gnu/libxml2/2.9.4"
      libxslt_path: "/opt/rbenv/versions/2.3.1/lib/ruby/gems/2.3.0/gems/nokogiri-1.7.0.1/ports/x86_64-pc-linux-gnu/libxslt/1.1.29"
      libxml2_patches: []
      libxslt_patches: []
      compiled: 2.9.4
      loaded: 2.9.4
```
Reproduced with:
```ruby
require 'nokogiri'
while true do 
  begin
    #parse simple xml. no namespace definition
    doc = Nokogiri::XML('<ns1:Root></ns1:Root>')
    # below code raises exception and leaks memory  
    body = doc.search('//ns1:Root').first 
    puts body.to_xml 
  rescue => e
   puts e
  end 
end 
```

flavorjones:
Thanks for opening this issue! I'll take a look.

flavorjones:
I've reproduced this behavior. Digging in.

flavorjones:
OK, this leak appears to come from using `Nokogiri_error_raise` to raise an exception during evaluation of the xpath query, which under the hood does a C `longjmp` and prevents libxml2 from cleaning up memory that had been allocated before the error was encountered.

`Nokogiri_error_raise` is used in a handful of places, and so now I'm worried that we may be leaking memory under other circumstances as well. I think I'd like to take a look at overhauling our current error-handling pattern to avoid raising exceptions while libxml2 is in the middle of something.

flavorjones:
I've audited the C code for non-trivial raises, use of handlers, and general patterns of error handling:

## places where `Nokogiri_error_raise` is used

### ext/nokogiri/xml_sax_push_parser.c

```C
[...]

PR #2654, opened 2022-09-28
wip: use xmlsec's noxxe entity loader

flavorjones:
Draft PR to try to use xmlsec's noxxe entity loader.

See #1582 for context.

This branch segfaults with seed 12441 and I have not yet been able to track it down


flavorjones:
I've pushed a commit that resets the external entity loader before the call that was segfaulting (in the HTML4 SAX parser context) and the segfault stops.

This tells me that we're not resetting the loader state properly after every call, this may be another instance of problems caused by raising errors - see #1610 and #2096.

PR #2658, opened 2022-10-06
fix: namespace nodes behave with compaction

flavorjones:

Under certain circumstances, GC compaction will cause a segfault when accessing wrapped `XML::Namespace` objects.

See previous similar work at #2579 

This fix, if merged, should be backported to v1.13.x and released as a bugfix.

Yes, repro test is included thanks to @eightbitraptor and @peterzhu2118

No functional behavior changes.

### DIFF
--- a/ext/nokogiri/xml_namespace.c
+++ b/ext/nokogiri/xml_namespace.c
@@ -25,12 +25,14 @@
 VALUE cNokogiriXmlNamespace ;
 
 static void
-dealloc_namespace(xmlNsPtr ns)
+_xml_namespace_dealloc(void *ptr)
 {
   /*
    * this deallocator is only used for namespace nodes that are part of an xpath
    * node set. see noko_xml_namespace_wrap().
    */
+  xmlNsPtr ns = ptr;
+
   if (ns->href) {
     xmlFree(DISCARD_CONST_QUAL_XMLCHAR(ns->href));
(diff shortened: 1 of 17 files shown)


ext/nokogiri/xml_namespace.c
```diff
@@ -25,7 +25,13 @@
 VALUE cNokogiriXmlNamespace ;
 
 static void
-dealloc_namespace(xmlNsPtr ns)
+_xml_namespace_mark(xmlNsPtr ns)
+{
+  rb_gc_mark((VALUE)ns->_private);
```
peterzhu2118, line 30: Since you support compaction, I think it shouldn't be pinned.

```suggestion
  rb_gc_mark_movable((VALUE)ns->_private);
```

flavorjones: The problem is actually that we shouldn't be marking here at all! My mistake.

flavorjones:
Interesting! Valgrind caught a problem: https://github.com/sparklemotion/nokogiri/actions/runs/3256748720/jobs/5347591384#step:7:3223

flavorjones:
:exploding_head: So fixing compaction for namespaces has caught a long-standing use-after-free bug in in the `Document#remove_namespaces!` method. Just pushed a fix for that.

Issue #2167, opened 2021-01-04
[bug] v1.11.0 native Windows gem causes nokogumbo installation to fail


flavorjones:
**Please describe the bug**

Nokogumbo cannot build against the Nokogiri v1.11.0 native Windows gem.

It is able to build against
- the v1.10.10 native Windows gem
- the v1.11.0 native Linux and Darwin gems.

**Help us reproduce what you're seeing**

See the Nokogumbo test suite's [failing windows jobs when built against v1.11.0](https://github.com/rubys/nokogumbo/actions/runs/461489890):

```
MAKE='make -j2 V=1' bundle exec rake -- --with-libxml2
mkdir -p tmp/x64-mingw32/nokogumbo/2.7.2
cd tmp/x64-mingw32/nokogumbo/2.7.2
C:/hostedtoolcache/windows/Ruby/2.7.2/x64/bin/ruby.exe -I. ../../../../ext/nokogumbo/extconf.rb -- --with-libxml2
checking for whether -ID:/a/nokogumbo/nokogumbo/vendor/bundle/ruby/2.7.0/gems/nokogiri-1.11.0-x64-mingw32/ext/nokogiri is accepted as CFLAGS... yes
checking for whether -ID:/a/nokogumbo/nokogumbo/vendor/bundle/ruby/2.7.0/gems/nokogiri-1.11.0-x64-mingw32/ext/nokogiri/include is accepted as CFLAGS... yes
checking for whether -ID:/a/nokogumbo/nokogumbo/vendor/bundle/ruby/2.7.0/gems/nokogiri-1.11.0-x64-mingw32/ext/nokogiri/include/libxml2 is accepted as CFLAGS... yes
checking for libxml/tree.h... yes
checking for nokogiri.h... yes
creating Makefile
cd -
cd tmp/x64-mingw32/nokogumbo/2.7.2
make -j2 V=1
(echo EXPORTS && echo Init_nokogumbo) > nokogumbo-x64-mingw32.def
...
[...]

PR #2659, opened 2022-10-07
ci: run bash scripts on windows runners

flavorjones:

Run the full test suite against the installed gems on windows systems.


flavorjones:
OK, these 4 remaining failures are real failures when building from source on Windows, I think. Grr.

flavorjones:
I think I've fixed the failures, which were compiling-from-source-with-system-libraries.

flavorjones:
Still struggling to get the nokogumbo integration test to not fail during linktime. Relevant prior work is at https://github.com/sparklemotion/nokogiri/issues/2167

flavorjones:
Ugh, it's because Nokogumbo 2.0.4 doesn't have the fixes for the linked flags! :facepalm: OK, should be green now.

Issue #2491, opened 2022-04-02
upstream remaining truffleruby patches


flavorjones:
See #1882 and #2095 for previous threads.

Conversation also happening in slack at https://graalvm.slack.com/archives/CMY63522F/p1648917137877149

---

Nokogiri patches live in https://github.com/oracle/truffleruby/blob/master/lib/truffle/truffle/patches/nokogiri_patches.rb

This approach is brittle in the face of trivial changes in Nokogiri's C extension. For example, [this patch](https://github.com/oracle/truffleruby/blob/aeb95dc5f03dc4397b8bc83acfc36cd53eeec188/lib/truffle/truffle/patches/nokogiri_patches.rb#L78-L81) has been broken since [this commit](https://github.com/sparklemotion/nokogiri/commit/9c5eed5) renamed `thing` to `retval`  in Nokogiri v1.13.0. This is coming up again in https://github.com/sparklemotion/nokogiri/pull/2489 where the code using variadic args is being modified and breaking more patches

One option is to upstream all of those changes, which I think would be preferable because it's easier to see and reason about.

flavorjones:
Note to self re: truffle dev env: comment out rbenv init in .bashrc and launch a new shell

aardvark179:
I think I've now got an approach that works for completely disabling the patches when not building with system libraries, and allowing them to still be used by older version (https://github.com/oracle/truffleruby/issues/2639#issuecomment-1095302734).

flavorjones:
@aardvark179 That's great! Can you tell me which of these patches is still needed so I can upstream them? (I imagine the varargs workarounds are still needed?)

flavorjones:
OK, closing this for now. I'll take a look at the truffle CI job which is failing in interesting ways.

PR #2663, opened 2022-10-12
Upstream truffleruby patches needed when using system libraries

eregon:
* Adapt tests for that configuration.

https://github.com/sparklemotion/nokogiri/issues/2491

Yes.

Only C, no need to change Java code.


flavorjones:
Normally we don't run the suite against TR for every commit (because it takes so long we opt for a cron job), so I've just pushed a commit that tries to make sure CI will run TR when branch names have "truffle" in them.

eregon:
> Normally we don't run the suite against TR for every commit (because it takes so long we opt for a cron job), so I've just pushed a commit that tries to make sure CI will run TR when branch names have "truffle" in them.

Right. Since there is workflow_dispatch I should be able to schedule it on my fork manually.

BTW for test speed, the main cost is the default NOKOGIRI_TEST_GC_LEVEL=major. Without that tests execute in just 1 minute for me locally. So maybe we should set NOKOGIRI_TEST_GC_LEVEL=normal for TruffleRuby, or GC less often and then we could run it in the gate/more often?

eregon:
It passed in https://github.com/eregon/nokogiri/actions/runs/3237310347/jobs/5304216696

eregon:
Given these test times over 1 hour yeah I think we should definitely force GC less often 😅 
We can do that separately though, I can try it on a branch tomorrow with `NOKOGIRI_TEST_GC_LEVEL=normal` for comparison.

ext/nokogiri/xml_sax_parser.c
```diff
@@ -203,10 +203,16 @@ warning_func(void *ctx, const char *msg, ...)
   VALUE doc = rb_iv_get(self, "@document");
   VALUE rb_message;
 
+#ifdef TRUFFLERUBY_NOKOGIRI_SYSTEM_LIBRARIES
+  /* It is not currently possible to pass var args from native
+     functions to sulong, so we work around the issue here. */
```
eregon, line 208: I've asked the Sulong team about that (@rschatz), and it sounds like a limitation of libffi (can't read varargs from a libffi callback), so probably nothing we can do about it in the short term.

eregon: Links I've found about that from a quick search:
* https://github.com/libffi/libffi/issues/388
* https://github.com/frida/frida-gum/issues/364#issuecomment-547761075

eregon:
> We can do that separately though, I can try it on a branch tomorrow with `NOKOGIRI_TEST_GC_LEVEL=normal` for comparison.

It runs in ~3min (with `--enable-system-libraries`) and ~8-9min (without): https://github.com/eregon/nokogiri/actions/runs/3241791089/jobs/5314201500
I'll push that on this PR too, but feel free to remove it if you want to keep it for these jobs.

I'm unsure how useful it is to GC between tests on TruffleRuby. I would say not worth it given the huge overhead.
Marking actually happens when leaving a C function in TruffleRuby, not during the JVM GC, and to detect a problem we need to execute code using e.g. a freed handle. The GC between tests makes it rather unlikely to reuse such a handle, so for TruffleRuby I'd think GCs happening during the tests (which already occur naturally) are more useful to detect this kind of bugs.

flavorjones:
Yes, the GC settings are there primarily to exercise CRuby and I wasn't even aware that TruffleRuby supported those settings. Please do feel free to add a commit that restores normal GC behavior for TR.

I have a terribly busy day today, but I will circle back to this by the weekend. I'm glad we're green and look forward to staying green!

flavorjones:
Thank you, @eregon!

eregon:
Thank you for merging!

I plan to remove the patches in TruffleRuby after the next nokogiri release, because they shouldn't be needed anymore, they are currently only needed when using system libraries, and with this PR won't be necessary for future nokogiri versions (when using system libraries).
When using packaged libraries (the default) there is already no patching.

flavorjones:
OK. If it's helpful, for managing upstream version checks, these changes will show up in Nokogiri 1.14.0.

Issue #2665, opened 2022-10-14
update vendored libxml2 to v2.10.3


flavorjones:
See https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.10.3

This addresses two CVEs, so we should also backport to v1.13.x and cut a security release.

- [x] main: https://github.com/sparklemotion/nokogiri/pull/2666
- [x] v1.13.x: https://github.com/sparklemotion/nokogiri/pull/2667

PR #2666, opened 2022-10-14
dep: update libxml2 to v2.10.3

flavorjones:

update libxml2 to v2.10.3 from v2.10.2

See https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.10.3

Related to #2665 

Existing coverage is fine.

Note the new libxml2 behavior around CDATA nodes longer than 10MB.


flavorjones:
I'll kick these jobs off again once the tarball is mirrored.

PR #2667, opened 2022-10-14
dep: update libxml2 to v2.10.3 (backport to v1.13.x)

flavorjones:

Backport of #2666 to the v1.13.x branch.

Also see parent issue #2665.


EduardoGHdez:
Hey @flavorjones Would it be okay if I put a PR for backporting this to 1.12.x?
Curious if there is any strong reason for not doing so 😄

flavorjones:
@EduardoGHdez Thanks for asking this question.

The v1.12.x branch is not maintained and we are not planning on making any more releases on that branch. Can you say more about what you're hoping to achieve with a backport? Why aren't you instead upgrading to v1.13.x?

If your primary interest is security, then I'm obligated to point out there are published vulnerabilities in the v1.12.x code that will remain unaddressed (see [here](https://github.com/sparklemotion/nokogiri/security/advisories/GHSA-crjr-9rc5-ghw8) and [here](https://github.com/sparklemotion/nokogiri/security/advisories/GHSA-xh29-r2w5-wx8m)).

EduardoGHdez:
> Why aren't you instead upgrading to v1.13.x?

It's in our plans to upgrade to 1.13.x, but the app I'm working on hasn't been upgraded to ruby 2.6.x (Nokogiri 1.13 requires using it)

I was just looking for a temp solution meanwhile we work on the ruby upgrade.

Perhaps this is not but one more reason to prioritize that upgrade work 😅 

Thanks for answering and pointing to the other unaddressed vulnerabilities 🙌🏽 , this is very helpful 🙇🏽

PR #2672, opened 2022-10-18
Modernize rcd_image_version retrieval

gjtorikian:
GitHub is [deprecating the use of `set-output`](https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/), so this PR removes the warning by updating the same logic to the new sequence.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -375,7 +375,7 @@ jobs:
           bundler-cache: true
           bundler: latest
       - id: rcd_image_version
-        run: bundle exec ruby -e 'require "rake_compiler_dock"; puts "::set-output name=rcd_image_version::#{RakeCompilerDock::IMAGE_VERSION}"'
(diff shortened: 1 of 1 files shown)


flavorjones:
Hi, I had tried this over the weekend and I believe it won't work because you can't use `env` in the `container.image` portion of the job config. I'll click "approve and run" and we'll see?

flavorjones:
From the [environment files docs](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#environment-files)

> You can make an environment variable available to **any subsequent steps in a workflow job** by defining or updating the environment variable and writing this to the GITHUB_ENV environment file.

gjtorikian:
Oh, strange. I'm not sure what you tried but I ported these exact changes into [my own builds](https://github.com/gjtorikian/commonmarker/actions/runs/3274848618/jobs/5388925554#step:2:18) and it works.

flavorjones:
OK, committed a fixed commit and it seem to be running! :crossed_fingers:

flavorjones:
All green! Going to merge. Thank you for showing me how to do this!!

gjtorikian:
Thank you for showing me how to test and build native gems! 😅  My wish was to pay that forward.

PR #2676, opened 2022-10-20
Remove GC.start required by ruby_memcheck

peterzhu2118:

As of version 1.2.0, ruby_memcheck no longer needs the call to GC.start at exit. It will do it automatically.

N/A.

No.


flavorjones:
All green! Merging, thanks for doing this.

Issue #2689, opened 2022-11-09
extconf not adding an appropriate -I for gumbo-parser on macOS


stevecheckoway:
**Please describe the bug**

When compiling the C extension on macOS, compiling `gumbo.c` fails because it cannot find `nokogiri_gumbo.h`. The real problem is that the `-Ipath/to/gumbo-parser/include` is not getting passed to the compiler.

Here's the relevant line in the `mkmf.log` file
```
block in append_cppflags: checking for whether -I/Users/steve/programming/nokogiri/tmp/x86_64-darwin21/nokogiri/3.1.2/ports/x86_64-darwin21/libgumbo/1.0.0-nokogiri/include is accepted as CPPFLAGS... -------------------- no
```

The underlying problem seems to be that the configuration test is using `-Werror` along with the unsupported warning `-Wmisleading-indentation`.

Here's the relevant portion of the log. Note that checking for the flag fails but that the last line shows checking for `gumbo_parse_with_options()` succeeds.

```
block in append_cppflags: checking for whether -I/Users/steve/programming/nokogiri/tmp/x86_64-darwin21/nokogiri/3.1.2/ports/x86_64-darwin21/libgumbo/1.0.0-nokogiri/include is accepted as CPPFLAGS... -------------------- no

[...]

PR #2690, opened 2022-11-11
ext: tell clang to ignore unknown options

flavorjones:

Related to #2689.

No, not sure how to test this in an automated way without adding dispropotionate complexity.

@stevecheckoway any chance you can try this out on your system?

N/A


stevecheckoway (approved):
LGTM

This fixes the build when Ruby is compiled with a newer version of Clang than the system Clang.

PR #2691, opened 2022-11-11
Fix loss of precision warnings

stevecheckoway:

There are multiple integer loss of precision warnings. 

There are three categories of warnings fixed.
1. Most are spurious warnings where we cast an `int` to a `long` and then use `INT2NUM` rather than just avoiding the cast);
2. With the HTML5 parser and a file containing an error on line number 2^31 or higher (or column number 2^31 or higher), the returned errors will have the wrong line (resp. column) number.
3. Node pointers are cast to a `int`s and used as ids for node comparison. Two different nodes could have the same least significant 32 bits in their pointers which would cause them to compare equal.

I added no test coverage. The warnings in the first category are harmless. The ones in the second category require producing input with at least 2^31 bytes. This doesn't seem worth it. Testing the third category seems extremely difficult.

Yes! It fixes the behavior of two different nodes possibly comparing equal in the C implementation. It also fixes incorrect line/column number for errors in pathological cases.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -627,6 +627,13 @@ append_cflags("-Winline")
 # good to have no matter what Ruby was compiled with
 append_cflags("-Wmissing-noreturn")
 
+# check integer loss of precision
+if darwin?
+  append_cflags("-Wshorten-64-to-32")
+else
+  append_cflags("-Wconversion -Wno-sign-conversion")
+end
+
 # handle clang variations, see #1101
 if darwin?
   append_cflags("-Wno-error=unused-command-line-argument-hard-error-in-future")
--- a/ext/nokogiri/gumbo.c
+++ b/ext/nokogiri/gumbo.c
@@ -281,12 +281,12 @@ add_errors(const GumboOutput *output, VALUE rdoc, VALUE input, VALUE url)
       rb_iv_set(syntax_error, "@code", INT2NUM(1));   // XML_ERR_INTERNAL_ERROR
       rb_iv_set(syntax_error, "@level", INT2NUM(2));  // XML_ERR_ERROR
       rb_iv_set(syntax_error, "@file", url);
-      rb_iv_set(syntax_error, "@line", INT2NUM(position.line));
+      rb_iv_set(syntax_error, "@line", SIZET2NUM(position.line));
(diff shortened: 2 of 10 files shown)


flavorjones:
I prepended a commit that updates extconf.rb to set gcc or clang flags for integer precision warnings.

The changes here also break the truffleruby build, something that shouldn't be a blocker but that I want to understand before merging.

flavorjones:
Undoing that last commit for now. Will investigate the truffleruby breakage and then circle back on #2692

stevecheckoway:
> I also noticed that the test suite only exercises `html5lib-tests` fragment cases by using the tag name (passing context as a string) and not by passing context as an instantiated node. Would it be worthwhile to do both, since the majority of real-world cases are going to be using a context node (e.g., `Node#fragment`)?

Yes. We should definitely be testing with a context node. The specific context node _mostly_ doesn't matter (beyond the type), but it certainly does in this case.

flavorjones:
> Yes. We should definitely be testing with a context node

Ok, I'll open a new issue so we don't forget to do that.

flavorjones:
@eregon In this fix, we're using `rb_unit2big` to convert a pointer into a `VALUE` that we can use for equality comparisons. That function doesn't exist in truffleruby. Do you have any suggestions on portable ways to do this?

eregon:
`rb_uint2inum` is implemented on TruffleRuby and has the same signature.

My understanding is `rb_uint2big` is not recommend since that assumes it creates a Bignum but the value might be in Fixnum range, and what the docs say: https://github.com/ruby/ruby/blob/6dd1a5f532711a28ffe815455503f5a2b1f70d23/include/ruby/internal/arithmetic/intptr_t.h#L60-L61. TruffleRuby does not allow creating Bignums in Fixnum range because that's wasteful and inefficient, so cannot really implement that C API function.

stevecheckoway:
It's probably the case on every platform that Nokogiri (and probably Ruby) works on that the size of a `size_t` is the same as the size of an `uintptr_t`. It maybe sufficient to use `SIZET2NUM((size_t)(uintptr_t)node)`.

stevecheckoway:
Oh I missed @eregon's comment about `rb_uint2inum()`. We should just try that instead.

flavorjones:
OK, changed it to `rb_uint2inum`, rebased, and force-pushed.

PR #2692, opened 2022-11-12
Fix crash when calling fragment on annotation-xml elements

stevecheckoway:

Calling `fragment()` on an `annotation-xml` element crashes.

```ruby
require 'nokogiri'

doc = Nokogiri::HTML5.parse('<!DOCTYPE html><math><annotation-xml encoding="text/html" /></math>')
a_xml = doc.xpath('//math:annotation-xml')[0]
frag = a_xml.fragment('<p>')
```

Yes. This functionality was not previously exercised by tests. It is now.

It fixes a crash that's only present in the C implementation.

### DIFF
--- a/ext/nokogiri/gumbo.c
+++ b/ext/nokogiri/gumbo.c
@@ -501,7 +501,8 @@ error:
     }
 
     // Encoding.
-    if (RSTRING_LEN(tag_name) == 14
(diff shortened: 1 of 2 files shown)


flavorjones:
I'll sort it today.

flavorjones:
LGTM! Merging. Thanks!

Issue #2513, opened 2022-04-11
explore: removing or overhauling the EncodingReader


flavorjones:
The Nokogiri::HTML4::EncodingReader class is used to try to detect encoding of HTML4 documents when they have ambiguous encoding.

Recently, [a REDOS vulnerability was found in this code](https://github.com/sparklemotion/nokogiri/security/advisories/GHSA-crjr-9rc5-ghw8). There are other regular expressions which should be vetted; and we should explore replacing some of those regexes with simpler calls like `String#include?`.

This class was written during a time (Ruby 1.9) when Ruby strings were encoded as ASCII-8BIT by default. This hasn't been true since (I think) Ruby 2.0, and so this complexity may only be for an edge case that we no longer need to support; and so maybe we can remove the entire class thereby simplifying both CRuby and JRuby implementations.

flavorjones:
Perhaps more specifically: let's consider unifying the encoding detection algorithm from the HTML5 parser and the HTML4 parser.

flavorjones:
@stevecheckoway notes that the HTML5 encoding detection is incomplete with respect to https://html.spec.whatwg.org/multipage/parsing.html#prescan-a-byte-stream-to-determine-its-encoding

PR #2693, opened 2022-11-14
fix: html5 encoding detection case insensitive re: meta tag

flavorjones:

I found an interesting test failure while playing with upgrading Mastodon to use the HTML5 parser that was caused by case-sensitivity in the HTML5 encoding detection.

Yes.

N/A


stevecheckoway (approved):
This change seems good.

The implementation of the [prescan algorithm](https://html.spec.whatwg.org/multipage/parsing.html#prescan-a-byte-stream-to-determine-its-encoding) is incomplete though.

flavorjones:
> The implementation of the [prescan algorithm](https://html.spec.whatwg.org/multipage/parsing.html#prescan-a-byte-stream-to-determine-its-encoding) is incomplete

I added a note to #2513 so this won't get forgotten.

PR #2698, opened 2022-11-14
fix: include ruby/defines.h early to avoid _GNU_SOURCE confusion

### LINKED ISSUE #2696: [bug] nokogiri.h includes ruby.h after other headers
**Please describe the bug**

See https://github.com/sparklemotion/nokogiri/blob/cd2700ab2a8b981d555e40c9f355ed22cbf5f161/ext/nokogiri/nokogiri.h#L68
So ruby.h is included after many other headers.

One issue I noticed with that is that for instance `_GNU_SOURCE` is defined by Ruby headers but not by nokogiri or the headers before, and so as a result e.g. `_DEFAULT_SOURCE` (which would be set if `_GNU_SOURCE` was defined at the time, in `/usr/include/features.h`) and as a result `__USE_MISC` are not defined, which means the first include of `sys/select.h` does not define `fd_mask` and `NFDBITS`, which means `rb_fdset_t` is the one from [`intern/select/posix.h`](https://github.com/ruby/ruby/blob/master/include/ruby/internal/intern/select/posix.h) and not the expected one from https://github.com/ruby/ruby/blob/master/include/ruby/internal/intern/select/largesize.h.
This means if rb_thread_fd_select() would be used by nokogiri it would most likely segfault or behave incorrectly.

A solution seems to include `ruby/defines.h` or `ruby.h` before other headers, then everything should be consistent.

I noticed this because I added a `#error` in TruffleRuby if `intern/select/posix.h` gets included as that would be a bug if rb_thread_fd_select() gets used with an inconsistent definition of `rb_fdset_t`.

This seems an unfortunate pitfall of C, because the first include of `features.h` decides features for all future-included headers.

Adding `#error "this file should not be indluded"` manually in ruby's `intern/select/posix.h` should reproduce this when installing nokogiri from source.

**Environment**

Fedora 35

flavorjones:

Fixes #2696

Unfortunately there's no failing test, but #2696 makes the argument that this is an improvement, and existing native gem CI pipelines should make sure nothing breaks.

No.

### DIFF
--- a/ext/nokogiri/nokogiri.h
+++ b/ext/nokogiri/nokogiri.h
@@ -1,6 +1,8 @@
 #ifndef NOKOGIRI_NATIVE
 #define NOKOGIRI_NATIVE
 
+#include <ruby/defines.h> // https://github.com/sparklemotion/nokogiri/issues/2696
+
 #ifdef _MSC_VER
 #  ifndef WIN32_LEAN_AND_MEAN
 #    define WIN32_LEAN_AND_MEAN
@@ -22,7 +24,6 @@
 #  define NOKOPUBFUN
 #  define NOKOPUBVAR extern
 #endif
-
 
 #include <stdlib.h>
 #include <string.h>


flavorjones:
cc @eregon

ext/nokogiri/nokogiri.h
```diff
@@ -23,6 +23,7 @@
 #  define NOKOPUBVAR extern
 #endif
 
+#include <ruby/defines.h> // https://github.com/sparklemotion/nokogiri/issues/2696
```
eregon, line 26: Maybe it should be even above `#ifdef _MSC_VER`? Since there are some includes in that block on Windows.

eregon (commented):
Thank you!
Indeed, there is no obvious test for this.
One could check that `NFDBITS` is defined e.g. on Linux, but that might not be the case for all unix'es.

ext/nokogiri/nokogiri.h
flavorjones: Sure, let's give it a whirl and see what CI says.

PR #2709, opened 2022-11-30
test: exercise fragment construction with a context Element

### LINKED ISSUE #2697: Run html5lib-tests fragment construction tests with an instantiated context node
As noted in #2691, the `html5lib-tests` fragment construction tests are being run using a string (tagname and optional namespace prefix) rather than an instantiated node. This is a hole in the test coverage that resulted in an undiscovered big fixed in #2692.

The desired outcome is to run the fragment tests twice: once with the tag name string, and once with a fully-instantiated node as the context.

flavorjones:

exercise fragment construction with a context Element, in addition to the existing tests that use the context element name

Closes #2697

Before:
`4191 tests, 58094 assertions, 0 failures, 0 errors, 17 skips`

After:
`4385 tests, 60489 assertions, 0 failures, 0 errors, 17 skips`

### DIFF
--- a/test/html5/test_tree_construction.rb
+++ b/test/html5/test_tree_construction.rb
@@ -60,9 +60,27 @@ class TestHtml5TreeConstructionBase < Nokogiri::TestCase
 
   def run_test
     if @test[:context]
-      ctx = @test[:context].join(":")
-      doc = Nokogiri::HTML5::Document.new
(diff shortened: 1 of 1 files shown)


stevecheckoway (approved):
Good idea to test using a context node. This LGTM.

PR #2735, opened 2022-12-20
perf: update gumbo utf8 decode

flavorjones:

Related to https://github.com/sparklemotion/nokogiri/issues/2722

update gumbo utf8 decode() from https://bjoern.hoehrmann.de/utf-8/decoder/dfa/ which apparently saves a shift instruction for every byte. benchmarking doesn't find a discernible performance improvement, though.

N/A

N/A

### DIFF
--- a/gumbo-parser/src/utf8.c
+++ b/gumbo-parser/src/utf8.c
@@ -50,38 +50,34 @@
 // all copies or substantial portions of the Software.
 
 #define UTF8_ACCEPT 0
-#define UTF8_REJECT 12
+#define UTF8_REJECT 1
 
 static const uint8_t utf8d[] = {
-  // The first part of the table maps bytes to character classes that
-  // to reduce the size of the transition table and create bitmasks.
-   0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,  0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,0,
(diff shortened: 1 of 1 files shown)


gumbo-parser/src/utf8.c
```diff
@@ -50,38 +50,34 @@
...
-    (*state != UTF8_ACCEPT)
-      ? (byte & 0x3fu) | (*codep << 6)
-      : (0xff >> type) & (byte);
+  *codep = (*state != UTF8_ACCEPT) ?
+    (byte & 0x3fu) | (*codep << 6) :
+    (0xff >> type) & (byte);
 
-  *state = utf8d[256 + *state + type];
+  *state = utf8d[256 + *state*16 + type];
   return *state;
```
stevecheckoway, line 81: I believe this is the older, slower code. The `*16` is the shift that Rich Felker eliminated by pre-multiplying
> On 24th June 2010 Rich Felker pointed out that the state values in the transition table can be pre-multiplied with 16 which would save a shift instruction for every byte. D'oh! We actually just need 12 and can throw away the filler values previously in the table making the table 36 bytes shorter and save the shift in the code.

stevecheckoway (commented):
I don't think this should be merged. This moves from the slightly more optimized version to the slightly less optimized version.

gumbo-parser/src/utf8.c
flavorjones: Ah, I see, that's confusing. OK, will close.

PR #2736, opened 2022-12-20
fix: fragment context parsing

### LINKED ISSUE #2646: [bug] parsing a fragment in-context may yield a different DOM than parsing the same document due to quirks mode handling
I'm not sure if the title above is correct, but please consider these two ways of parsing a `<table>` within a `<p>`:

```ruby
require "nokogiri"
tags = "<p>A <table><tr><td>B</table> C"
doc = Nokogiri::HTML5.parse("<div>"+tags)
puts doc
doc.at("div").inner_html = tags
puts doc
```

output:
```
<html><head></head><body><div><p>A <table><tbody><tr><td>B</td></tr></tbody></table> C</p></div></body></html>
<html><head></head><body><div><p>A </p><table><tbody><tr><td>B</td></tr></tbody></table> C</div></body></html>
```

I'm fairly certain the output should not be different. In both cases we're parsing the same tags, and the parent element is a `<div>`.   Note that `Nokogiri::HTML5.fragment(tags)` produces the same structure as `Nokogiri::HTML5.parse(tags)`. So this bug/behavior looks specific to `coerce` somehow. But in the code sample above if we replace `HTML5.parse` by `HTML5.fragment`, the two `puts doc` produce the same output. So honestly I'm not sure what's going on.

flavorjones:

Fixes #2646.

This PR introduces:

- `HTML5::QuirksMode`, an enum
- `HTML5::Document#quirks_mode`
- `HTML5::DocumentFragment#quirks_mode`

It also changes the behavior of the fragment parser to use the parent document's quirks mode, and default to `NO_QUIRKS` in edge cases.

Yes!

The HTML5 parser is only supported in CRuby.


stevecheckoway (approved):
LGTM

PR #2738, opened 2022-12-21
ci: cache ports for zlib and iconv on windows

flavorjones:

CI windows builds have been failing recently while attempting to download the zlib tarball.

This change should cache the tarball for subsequent runs.


flavorjones:
This isn't doing what I expected it to do, because the gem installation uses a ruby system directory and not `ports/`.

flavorjones:
see #2739 for an alternative solution

PR #2757, opened 2023-01-11
Get rid of anonymous eval calls

casperisfine:
Things declared in anonymous eval are always annoying to locate. (e.g. profling etc).

cc @flavorjones

### DIFF
--- a/test/xml/node/test_subclass.rb
+++ b/test/xml/node/test_subclass.rb
@@ -16,16 +16,16 @@ module Nokogiri
           Nokogiri::XML::Node => '"foo", doc',
           Nokogiri::XML::Text => '"foo", doc',
         }.each do |klass, constructor|
-          class_eval %{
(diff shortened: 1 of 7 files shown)


test/xml/node/test_subclass.rb
```diff
@@ -16,16 +16,16 @@ class TestSubclass < Nokogiri::TestCase
...
               doc = Nokogiri::XML::Document.new
               klass = Class.new(#{klass.name})
               node = klass.new(#{constructor})
               assert_instance_of klass, node
             end
-          }
+          RUBY
 
-          class_eval <<-eocode, __FILE__, __LINE__ + 1
+          class_eval <<~RUBY, __FILE__, __LINE__ + 1
```
casperisfine, line 28: This one was OK, but I changed it simply for consistency.

flavorjones:
Thanks!

PR #2768, opened 2023-01-17
Fix Clang warnings about casting void* to an enum

stevecheckoway:

Clang warns about casting `void *` to an enum. This cast is fine in this case so first cast to `intptr_t`.

This isn't a functional change so no tests were added. (I don't think we test for the absence of compiler warnings.)

No.


flavorjones:
Looks good! Thanks!

Issue #2769, opened 2023-01-17
Updating the psych requirements to 5.0 breaks macOS build


stevecheckoway:
Commit c7a780298614c81f3dc5925e2b0c23b836e1c6e0 updates the psych requirements from 4.0 to 5.0 and this is broken on macOS. There's more context [here](https://github.com/ruby/setup-ruby/issues/409).

The basic problem is that psych no longer bundles libyaml and macOS doesn't come with libyaml nor does it have a system package manager that can be used to install it. It seems the expected situation is to use homebrew to install the library system wide.

It's possible to install the psych gem by passing an option pointing to the libyaml source directory, but this is broken on M1 macs. After I filed a [bug report](https://github.com/ruby/psych/discussions/615), a pre-release version of psych was created and pushed to rubygems. I _can_ successfully install this version of psych, but that doesn't help with building Nokogiri any longer as `bundle install` still tries to install the non-prerelease version of the gem.

<details><summary><code>bundle install</code> output</summary>

```
$ bundle install
Fetching gem metadata from https://rubygems.org/........
Resolving dependencies...
Using rake 13.0.6
Using ast 2.4.2
Using builder 3.2.4
Using bundler 2.3.7
Using docile 1.4.0
Using hoe-markdown 1.4.0
Using mini_portile2 2.8.1
Using minitest 5.17.0
Using ruby-progressbar 1.11.0
Using racc 1.6.2
Using rexml 3.2.5
Using parallel 1.22.1
Using rake-compiler 1.2.1
Using rake-compiler-dock 1.3.0
Using regexp_parser 2.6.1
Using nokogiri 1.14.1.dev from source at `.` and installing its executables
Using parser 3.2.0.0
Using rainbow 3.1.1
Using rubocop-ast 1.24.1
Using json 2.6.3
Using stringio 3.0.4
Using simplecov-html 0.12.3
Using simplecov_json_formatter 0.1.4
Fetching psych 5.0.1
Using ansi 1.5.0
Using rexical 1.0.7
Using unicode-display_width 2.4.2
Using minitest-reporters 1.5.0
Using simplecov 0.21.2
Using rubocop 1.43.0
Using rubocop-rake 0.6.0
Using rubocop-performance 1.15.2
Using rubocop-shopify 2.10.1
Using rubocop-minitest 0.26.0
Using ruby_memcheck 1.2.0
Installing psych 5.0.1 with native extensions
[...]

PR #2770, opened 2023-01-18
Move rdoc and psych into their own bundler group

stevecheckoway:
Psych 5 no longer bundles libyaml which causes `bundle install` to fail on macOS which doesn't have libyaml and doesn't have an easy, standard way to install it.

One can use `bundle config set --local without rdoc` to disable installing the `psych` and `rdoc` (which depends on `psych`) gems.

Hopefully, this is only temporary and the upstream Psych team will produce a better way to handle this.

#2769

As this is a (hopefully temporary) workaround, rather than a supported configuration, I have not.

No.

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -40,5 +40,4 @@ end
 # Then re-run `bundle install`.
 group :rdoc do
   gem "rdoc", "6.5.0"
-  gem "psych", "~> 5.0" # psych 5 isn't building in places yet https://github.com/ruby/setup-ruby/issues/409
 end
--- a/Gemfile
+++ b/Gemfile
@@ -15,8 +15,6 @@ group :development do
 
   # documentation
   gem "hoe-markdown", "= 1.4.0"
-  gem "rdoc", "6.5.0"
-  gem "psych", "~> 5.0" # psych 5 isn't building in places yet https://github.com/ruby/setup-ruby/issues/409
 
   # parser generator
   gem "rexical", "= 1.0.7"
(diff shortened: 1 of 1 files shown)


Gemfile
```diff
@@ -36,3 +34,11 @@ group :development do
     gem "rubocop-shopify", "2.10.1"
   end
 end
+
+# If Psych doesn't build, you can disable this group locally by running
+# `bundle config set --local without rdoc`
+# Then re-run `bundle install`.
+group :rdoc do
+  gem "rdoc", "6.5.0"
+  gem "psych", "~> 5.0" # psych 5 isn't building in places yet https://github.com/ruby/setup-ruby/issues/409
```
flavorjones, line 43: I think we should remove the `psych` line now. It was originally added in 6a3254cd to pin to `~> 4.0` but that was later updated by dependabot to `~> 5.0` in c7a78029. We actually don't care about Psych as long as it stays out of our way!

flavorjones (changes requested):

stevecheckoway:
I don't recall if you want me to squash and force push to the branch or if you use the "squash and merge" button normally.

flavorjones:
I generally use the "Merge pull request" to create a merge commit, but I don't feel very strongly about it. I'll hit that button now!

flavorjones:
Oh, but if you were asking if I usually force-push to the branch to have a clean history before merging, the answer is "yes"!

stevecheckoway:
Great. I'll do that in the future.

PR #2798, opened 2023-02-27
fix: support Encoding class for xml write_to

### LINKED ISSUE #2774: [feature] `Node#write_to` and friends should accept `Encoding` objects for `:encoding` kwargs
Currently these methods accept an `:encoding` arg, whose value is the name of the encoding:

```ruby
node.write_to(io, encoding: "UTF-8")
```

When passing in an `Encoding` object, though, we get an error:

```ruby
node.write_to(io, encoding: Encoding::UTF_8)
# TypeError: no implicit conversion of Encoding into String
#    .../lib/nokogiri/xml/node.rb:1368:in `native_write_to'
#    .../lib/nokogiri/xml/node.rb:1368:in `write_to'
```

We should handle `Encoding` objects wherever we accept encoding name strings.

ellaklara:

Fixes #2774

The PR adds two tests that cover both setting encoding with a string and the `Encoding` class.

**Question**

I discovered that the html5 node `write_to` (`lib/nokogiri/html5/node.rb`) also doesn't support the `Encoding` class, is this something you'd like me to add as well? Where would be a good place to test this? I couldn't find any encoding tests for this class.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 ### Fixed
 
 * [JRuby] Serializing an HTML4 document with `#write_to` and specifying no save options will properly emit an HTML document anyway, like libxml2 does. Previously JRuby emitted XML in this situation.
+* [JRuby] Serializing with `#write_to` will fall back to the document encoding when no encoding is specified, like libxml2 does. Previously JRuby emitted UTF-8 in this situation.
 
 
 ### Improved
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,9 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 
 ### Added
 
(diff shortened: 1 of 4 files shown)


flavorjones:
@ellaklara Thank you for contributing this! I've kicked off CI. I think we should support this in the HTML5  version as well. If there aren't tests we should definitely add that coverage. I'll take a deeper look in the next day or so.

flavorjones:
OK, bug report filed at https://github.com/sparklemotion/nokogiri/issues/2801

flavorjones:
@ellaklara OK! I've added a commit on this PR (before yours) that introduces a new file, `test/test_serialization_encoding.rb` that has some coverage for encoded output for XML, HTML4, and HTML5 serialization methods.

Again, only if you're up for it, it would be great to add `Encoding` tests there, and make sure they work for all those classes and methods. Let me know either way! :heart:

flavorjones:
I'm working on the failing JRuby tests, so don't worry about those!

flavorjones:
@ellaklara It's been about a week, so I'm going to go ahead and finish the work you so awesomely started! Will be sure to credit you in the CHANGELOG!

PR #2806, opened 2023-03-02
Migrate Nokogiri::XSLT::Stylesheet to the TypedData API

casperisfine:
The old `Data` API is soft deprecated since a long time. The `TypedData` API is a bit more strict and opens the door for compaction, write barriers, memsize, etc (they have to be opted individually)

Here I just do a very simple translation of a single type, so it's easy to review.

### DIFF
--- a/ext/nokogiri/xslt_stylesheet.c
+++ b/ext/nokogiri/xslt_stylesheet.c
@@ -18,6 +18,15 @@ dealloc(void *data)
   ruby_xfree(wrapper);
 }
 
+static const rb_data_type_t xslt_stylesheet_type = {
+  .wrap_struct_name = "Nokogiri::XSLT::Stylesheet",
+  .function = {
+    .dmark = mark,
+    .dfree = dealloc,
+  },
+  .flags = RUBY_TYPED_FREE_IMMEDIATELY
+};
+
 PRINTFLIKE_DECL(2, 3)
 static void
 xslt_generic_error_handler(void *ctx, const char *msg, ...)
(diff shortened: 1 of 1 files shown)


flavorjones:
@casperisfine Thank you for starting this work! The C codebase definitely needs to be modernized and I've been putting this off for too long.

Do you think it's worth making all the TypedData changes on one branch to benchmark anything? If not I'm happy to merge this and we can just work our way through the classes one at a time.

tenderlove:
I don't think using TypedData would cause any performance regressions.  Might be interesting to benchmark later, but IMO it's not worth our time to do it now.

btw, this PR LGTM

flavorjones:
> I don't think using TypedData would cause any performance regressions

Sorry, to be clear, I meant it to measure some real-world benchmarks for performance _improvements_, not regressions -- just in case it was academically interesting to anybody.

casperisfine:
So TypedData by itself won't make any difference. (accessing a typeddata is one or two extra comparisons, not much else).

What would give performance eventually, would be to implement write barriers, and the performance impact would be on minor GC based on the amount of these objects are long lived.

But If Shopify's monolith's is anything to go by, Nokogiri objects aren't really long lived, when inspecting a production heap all I could find was  `12` `"DATA(Nokogiri/XMLNode)"`. Some different apps may have long lived Nokogiri objects perhaps, I don't know.

So I'm really not expecting much if any measurable performance improvements here, this PR (and the followups I was planning) are mostly about using the more strict and modern API, and allowing to implement the `_memsize` functions.

flavorjones:
:+1: Will merge. Added a commit with some style changes (I'm trying to update the C files as I touch them). Will merge when it goes green.

Issue #2808, opened 2023-03-03
epic: update C extension to use the `TypedData` API


flavorjones:
The old `Data` API has been soft deprecated for a long time. The `TypedData` API is a bit more strict and opens the door for compaction, write barriers, memsize, etc.

Here are the classes that need to be updated:

- [x] `XSLT::Stylesheet` https://github.com/sparklemotion/nokogiri/pull/2806
- [x] `XML::XPathContext` https://github.com/sparklemotion/nokogiri/pull/2809
- [x] `XML::Document` https://github.com/sparklemotion/nokogiri/pull/2807
- [x] `XML::Reader` https://github.com/sparklemotion/nokogiri/pull/2810
- [x] `Nokogiri::EncodingHandler` https://github.com/sparklemotion/nokogiri/pull/2811
- [x] `XML::ElementContent` https://github.com/sparklemotion/nokogiri/pull/2814
- [x] `XML::Schema` https://github.com/sparklemotion/nokogiri/pull/2815
- [x] `XML::RelaxNG` https://github.com/sparklemotion/nokogiri/pull/2816
- [x] `HTML4::ElementDescription` https://github.com/sparklemotion/nokogiri/pull/2821
- [x] `XML::SAX::Parser` https://github.com/sparklemotion/nokogiri/pull/2823
- [x] `XML::NodeSet` https://github.com/sparklemotion/nokogiri/pull/2824
- [x] `XML::SAX::ParserContext` and `HTML4::SAX::ParserContext` https://github.com/sparklemotion/nokogiri/pull/2826
- [x] `XML::SAX::PushParser` and `HTML4::SAX::PushParser` https://github.com/sparklemotion/nokogiri/pull/2826
- [x] use `WB_PROTECTED` on all of the classes that don't declare mark functions https://github.com/sparklemotion/nokogiri/pull/2825
- [x] consistently use a method to unwrap rather than extern the data type struct https://github.com/sparklemotion/nokogiri/pull/2825
- [x] memleak warning https://github.com/sparklemotion/nokogiri/pull/2827

cc @casperisfine

flavorjones:
If anyone wants to lend a hand, look at the changes in #2806 for guidance on what to do.

casperisfine:
cc @etiennebarrie ^

flavorjones:
I'm grabbing XPathContext

flavorjones:
I'm grabbing NodeSet

casperisfine:
[...]

PR #2809, opened 2023-03-03
migrate XML::XPathContext to the TypedData API

flavorjones:

Part of the TypedData epic described at #2808.


flavorjones:
cc @byroot for :eyes:

byroot (commented):
I only reviewed the second commit, but it looks great to me.

PR #2811, opened 2023-03-06
ext: migrate Nokogiri::EncodingHandler to TypedData API

flavorjones:

See #2808


casperisfine (commented):
LGTM

PR #2816, opened 2023-03-06
Migrate XML::RelaxNG to the TypedData API

etiennebarrie:
Ref: #2808


flavorjones:
This looks good! I've rebased and refactored some of the existing C code (while we're touching it).

flavorjones (approved):

PR #2817, opened 2023-03-06
CI: Update Appveyor to latest image and ruby version

larskanis:

There are no relevant changes in the new image. It's just to avoid using an outdated one eventually.

Not changed.

Only the C impl.


flavorjones:
Thank you!

PR #2818, opened 2023-03-06
Fix deprecated double_heap option to GC.verify_compaction_references under 3.2

etiennebarrie:
When I'm running the tests with Ruby 3.2 I can see a deprecation warning, this fixes it.


flavorjones (approved):
Awesome, thank you!

PR #2821, opened 2023-03-07
Migrate HTM4::ElementDescription to TypedData

casperisfine:
Ref: https://github.com/sparklemotion/nokogiri/issues/2808

Since it has no mark function, we can enable write barrier without changing any code.

cc @flavorjones @etiennebarrie


flavorjones:
Looks good! Thanks.

(Side note: this morning I briefly looked into avoiding a T_DATA struct entirely, and just calling `htmlLookupTag` every time we need to access it, but that function does a binary search using string comparisons ... I didn't benchmark but it feels less efficient than just dereferencing a pointer.)

PR #2823, opened 2023-03-07
Migrate XML::SAX::Parser to the TypedData API

etiennebarrie:
Ref: #2808

We now rely on `TypedData_Make_Struct` to do the allocation instead of using `ruby_xcalloc` directly, and we use RUBY_TYPED_DEFAULT_FREE to let Ruby free it as well.

There's no mark function so we can enable RUBY_TYPED_WB_PROTECTED.

### DIFF
--- a/ext/nokogiri/html4_sax_parser_context.c
+++ b/ext/nokogiri/html4_sax_parser_context.c
@@ -83,7 +83,7 @@ parse_with(VALUE self, VALUE sax_handler)
   }
 
   Data_Get_Struct(self, htmlParserCtxt, ctxt);
-  Data_Get_Struct(sax_handler, htmlSAXHandler, sax);
+  TypedData_Get_Struct(sax_handler, htmlSAXHandler, &noko_sax_handler_type, sax);
 
   /* Free the sax handler since we'll assign our own */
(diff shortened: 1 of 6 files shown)


ext/nokogiri/html4_sax_parser_context.c
```diff
@@ -83,7 +83,7 @@ parse_with(VALUE self, VALUE sax_handler)
   }
 
   Data_Get_Struct(self, htmlParserCtxt, ctxt);
-  Data_Get_Struct(sax_handler, htmlSAXHandler, sax);
+  TypedData_Get_Struct(sax_handler, htmlSAXHandler, &noko_sax_handler_type, sax);
```
etiennebarrie, line 86: This one was surprising to us, it's cast as a `htmlSAXHandler`, but it's the same type actually. This is why we named the type `noko_sax_handler_type` without specializing either HTML or XML.

flavorjones: :+1:

flavorjones:
The memcheck failure is, I think, because we're no longer doing majors during CI runs. I'm going to ignore it for now and I'll circle back in a separate commit to do an explicit major GC `at_exit`.

PR #2831, opened 2023-03-13
fix build error for nokogiri in loongarch64

zhangwenlong8911:
update config.guess and config.sub for libxml2 and libxslt form https://git.savannah.gnu.org/gitweb/?p=config.git;a=tree to fix build error for libxml2 and libxslt


zhangwenlong8911:
> 



> @zhangwenlong8911 Thank you for submitting this PR!
> 
> Can you help me understand what the problem is that you're trying to solve? I don't think you need to update `rakelib/extensions.rake` and those changes should probably be removed. That file is only used to precompile/cross-compile.
> 
> If you _do_ want to precompile or cross-compile for loongarch64, let's have that conversation in a new issue, and let this PR be just for compiling from source.
> 
> For compiIing from source, I believe fixing the libxml2/libxslt config files is sufficient. However, I would very much prefer if that change was made _upstream_ in libxml2 and libxslt. Have you reported this problem in an upstream issue?
> 
> If it's not possible to fix upstream, I'm OK with patching those files here. However -- we are already patching libxslt with `patches/libxslt/0001-update-automake-files-for-arm64.patch`! Can you please just make sure we only have one patch?
> 
> Finally: do you have any ideas on how we can test that this platform works properly with these changes?

Thanks for the suggestion, I have removed the modifications to rakelib/extensions.rake and updated config.guess and config.sub in libxslt with one patch

At present, there is no better way to test. This submission is directly obtained from the upstream config project, and no other modifications have been made. At the same time, I have compiled aarch64 and loongarch64 locally.

flavorjones:
For future readers, I've verified that post-patch `config.guess` and `config.sub` match the current versions available at `https://git.savannah.gnu.org/gitweb/?p=config.git;a=tree`:

```
$ find . -name config.guess -or -name config.sub | xargs sha256sum | sort
b45ba96fa578cfca60ed16e27e689f10812c3f946535e779229afe7a840763e6  ./config.sub
b45ba96fa578cfca60ed16e27e689f10812c3f946535e779229afe7a840763e6  ./tmp/x86_64-linux-musl/nokogiri/3.0.2/tmp/x86_64-pc-linux-musl/ports/libxml2/2.10.3/libxml2-2.10.3/config.sub
b45ba96fa578cfca60ed16e27e689f10812c3f946535e779229afe7a840763e6  ./tmp/x86_64-linux-musl/nokogiri/3.0.2/tmp/x86_64-pc-linux-musl/ports/libxslt/1.1.37/libxslt-1.1.37/config.sub
b45ba96fa578cfca60ed16e27e689f10812c3f946535e779229afe7a840763e6  ./tmp/x86_64-linux/nokogiri/3.2.1/tmp/x86_64-pc-linux/ports/libxml2/2.10.3/libxml2-2.10.3/config.sub
bb14470dba3adf469b6e683307b783172b571abca13e7f5a77a4c94ea07b3811  ./config.guess
bb14470dba3adf469b6e683307b783172b571abca13e7f5a77a4c94ea07b3811  ./tmp/x86_64-linux-musl/nokogiri/3.0.2/tmp/x86_64-pc-linux-musl/ports/libxml2/2.10.3/libxml2-2.10.3/config.guess
bb14470dba3adf469b6e683307b783172b571abca13e7f5a77a4c94ea07b3811  ./tmp/x86_64-linux-musl/nokogiri/3.0.2/tmp/x86_64-pc-linux-musl/ports/libxslt/1.1.37/libxslt-1.1.37/config.guess
bb14470dba3adf469b6e683307b783172b571abca13e7f5a77a4c94ea07b3811  ./tmp/x86_64-linux/nokogiri/3.2.1/tmp/x86_64-pc-linux/ports/libxml2/2.10.3/libxml2-2.10.3/config.guess
```

(the files in the root directory are downloaded directly from gnu.org).

Kicking off CI!

stevecheckoway:
Did this get reported upstream? That seems preferable to maintaining a patch file.

flavorjones:
@stevecheckoway I agree, upstream is the long-term fix. I asked for it to be reported upstream, not sure if it has (probably not). If memory serves (I wasn't able to find an example quickly, though), complaints about using old autoconf files have been deprioritized because you can "just run autoreconf", and at times a commitment was made to do better in the next release. The config.guess and config.sub files are added by the human packager of the tarball and are not under source control, which means that the versions used have varied quite a bit over time.

The reality is that we've had to maintain a patch for libxslt for arm64 support since April 2021 so this isn't the worst thing in the world. Part of me actually wants to update mini_portile2 to inject the latest autoconf files if it can, but I'm not sure if that would break some projects.

zhangwenlong8911:
@stevecheckoway @flavorjones 
https://gitlab.gnome.org/GNOME/libxml2/-/releases
https://gitlab.gnome.org/GNOME/libxslt/-/releases
I checked the latest upstream code of libxml2 and libslt. The code does not contain config.guess and config.sub. The upstream uses autogen.sh to automatically obtain config.guess and config.sub from automake. At the same time, the latest upstream automake in config.guess and config.sub already include arm64 and loongarch64 support[https://savannah.gnu.org/git/?group=automake](url)

flavorjones:
Merged! Thank you, @zhangwenlong8911. This will be in the next release.

Issue #2841, opened 2023-04-03
[upstream] html5lib-tests head has failing tests


flavorjones:
Updates to upstream https://github.com/html5lib/html5lib-tests are failing in Nokogiri's "upstream" CI:

- https://github.com/sparklemotion/nokogiri/actions/runs/4594272018/jobs/8113098993

These failures are from one of these two commits:

- https://github.com/html5lib/html5lib-tests/commit/4f45c0211cf1d1f1af319470f77851f60f29914c
- https://github.com/html5lib/html5lib-tests/commit/1314b094ed2c4e418404576c24b88aae2ea4e0c1

Here's the output for the seven failing tests:

```
Failure:
TestHtml5TreeConstructionWebkit02#test_11 [/__w/nokogiri/nokogiri/test/html5/test_tree_construction.rb:120]
Minitest::Assertion: Wrong number of errors for <b><em><dcell><postfield><postfield><postfield><postfield><missing_glyph><missing_glyph><missing_glyph><missing_glyph><hkern><aside></b></em>.
Expected: 0
  Actual: 3

Failure:
TestHtml5TreeConstructionWebkit02#test_11__with_node [/__w/nokogiri/nokogiri/test/html5/test_tree_construction.rb:120]
Minitest::Assertion: Wrong number of errors for <b><em><dcell><postfield><postfield><postfield><postfield><missing_glyph><missing_glyph><missing_glyph><missing_glyph><hkern><aside></b></em>.
Expected: 0
  Actual: 3

Failure:
TestHtml5TreeConstructionWebkit01#test_26 [/__w/nokogiri/nokogiri/test/html5/test_tree_construction.rb:120]
Minitest::Assertion: Wrong number of errors for <html><body></body>
   <!-- Hi there --></html>.
Expected: 0
  Actual: 1

Failure:
TestHtml5TreeConstructionWebkit01#test_27 [/__w/nokogiri/nokogiri/test/html5/test_tree_construction.rb:120]
Minitest::Assertion: Wrong number of errors for <html><body></body></html>
   <!-- Hi there -->.
Expected: 0
  Actual: 1

Failure:
TestHtml5TreeConstructionTemplate#test_107 [/__w/nokogiri/nokogiri/test/html5/test_tree_construction.rb:120]
Minitest::Assertion: Wrong number of errors for <body></body><template>.
Expected: 0
  Actual: 3

Failure:
[...]

PR #2842, opened 2023-04-07
gumbo: update adjusted foreign attributes

flavorjones:

Partial fix for #2841

A test recently added upstream showed that the list of foreign attributes gumbo is adjusting needed to be updated. Specifically:

- add `xlink:arcrole`
- remove `xml:base`

(see https://html.spec.whatwg.org/multipage/parsing.html#adjust-foreign-attributes)

I've also pinned (for now) to my branch of html5lib-tests, and improved some of the error messages from the tree construction test wrapper.

This PR is all about the tests! :tada: 

The HTML5/gumbo functionality changed, but this does not exist in the Java implementation.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,8 @@
+# [submodule "test/html5lib-tests"]
+# 	path = test/html5lib-tests
+# 	url = https://github.com/html5lib/html5lib-tests.git
+# 	branch = master
 [submodule "test/html5lib-tests"]
 	path = test/html5lib-tests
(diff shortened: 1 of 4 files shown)


stevecheckoway (approved):
LGTM.

The changes to the test infrastructure are very welcome. Debugging why a particular test is failing was always a pain point.

stevecheckoway:
It might be worth thinking about how we can stay on top of changes to the spec other than following the [whatwg/html](https://github.com/whatwg/html) repo.

Issue #2565, opened 2022-06-05
test against upstream htmlunit-neko


flavorjones:
I recently shipped a PR upstream: https://github.com/HtmlUnit/htmlunit-neko/pull/13

If that gets accepted, then the pending tests from 7cec00e and 277db2e should start passing, and that's a good time to implement upstream testing against https://github.com/HtmlUnit/htmlunit-neko/

flavorjones:
Well, this is a bummer: htmlunit-neko dropped compatibility with Xerces in 2.68.0 (see https://github.com/HtmlUnit/htmlunit-neko/issues/17 for some context), which means 2.67.0 is the latest version we can compile against without some work.

I've spiked a bit today on what's involved, and it's mostly refactoring the XML and HTML4 parser context java classes to not inherit from one another. I think I can do it with a little bit more work.

rbri:
Yust released version 3.1.0 of Neko. Feel free to contact me if i you need some support with updating.

flavorjones:
@rbri Thank you for the offer! I saw the release, congrats!

I likely will need help (see my previous note), but it's not at the top of my list just yet. I've pushed my WIP branch to #2856, if you want to take a look and let me know if I'm on the right track or not that would be helpful.

chadlwilson:
Would be fantastic to see some progress here, as nokogiri-java reports a [vulnerability](https://web.nvd.nist.gov/view/vuln/detail?vulnId=CVE-2023-26119) via `gems/jruby/3.1.0/gems/nokogiri-1.14.3-java/lib/nokogiri/jruby/net/sourceforge/htmlunit/neko-htmlunit/2.63.0/neko-htmlunit-2.63.0.jar` right now (not sure if actually exploitable via nokigiri, but a bit of an inconvenience all the same).

flavorjones:
@chadlwilson If you've got time and the motivation, please take a look at the WIP branch at #2856. I need help!

chadlwilson:
For what it's worth (and for anyone else who arrives here concerned about CVEs being reported against nokogiri java variant) the current CVEs reported against neko-htmlunit `2.63.0` jar I believe to be false positives. They instead only affect the wider `htmlunit` which is versioned and released alongside `neko-htmlunit` and unfortunately shares a component grouping within the NIST NVD - but is not used by nokogiri's Java variant.

[...]

PR #2856, opened 2023-04-12
Update to latest htmlunit neko

flavorjones:

This is a WIP for #2565. Feedback and help welcome.


chadlwilson:
I had a look at this but it's probably a bit too involved for my xerces and nokogiri/htmlunit knowledge.

Was your thinking to essentially have two xerces versions eventually packaged? (Retain the mainline org.apache one as well as the shaded, refactored/simplified one inside htmlunit)?

It looks like nokogiri seems to try and reuse the same xerces xonfiguration for its own direct stuff as well as for use by htmlunit so seemed to require a change to the configuration approach.

PR #2857, opened 2023-04-12
Update libxml2 to 2.10.4 for 1.13.x

wwahammy:


flavorjones:
@wwahammy I'm curious. If you don't mind sharing, what are the reasons for not upgrading to Nokogiri v1.14.x? Did something break for you on this version branch?

wwahammy:
Sure, I'd be happy to share!

I have an app that runs Ruby 2.6 so I can't update to v1.14.x.

PR #2863, opened 2023-04-27
Improve gumbo error messages during tree construction

flavorjones:

Gumbo tree constructions errors could be improved. In this PR we're adding punctuation, cleaning up whitespace, and naming the input tag.

If the message was:

> 1:13: ERROR: This tag isn't allowed here  Currently open tags: html, body, div, table.

it is now:

> 1:13: ERROR: Start tag 'svg' isn't allowed here. Currently open tags: html, body, div, table.

No tests exist for error message contents. This is a potential area we could improve in the future.

No behavioral changes other than the contents of the error message.

### DIFF
--- a/gumbo-parser/src/error.c
+++ b/gumbo-parser/src/error.c
@@ -91,7 +91,7 @@ static void print_tag_stack (
   const GumboParserError* error,
   GumboStringBuffer* output
 ) {
-  print_message(output, "  Currently open tags: ");
+  print_message(output, " Currently open tags: ");
   for (unsigned int i = 0; i < error->tag_stack.length; ++i) {
     if (i) {
       print_message(output, ", ");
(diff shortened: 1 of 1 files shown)


gumbo-parser/src/error.c
```diff
@@ -347,13 +347,14 @@ static void handle_parser_error (
...
-        print_message(output, "Premature end of file");
+        print_message(output, "Premature end of file.");
         print_tag_stack(error, output);
       }
       return;
     case GUMBO_TOKEN_START_TAG:
     case GUMBO_TOKEN_END_TAG:
-      print_message(output, "That tag isn't allowed here");
+      print_message(output, "'%s' tag isn't allowed here.",
+                    gumbo_normalized_tagname(error->input_tag));
```
stevecheckoway, line 357: Would it be better to have the error message make clear if it's a start or end tag? Maybe `<%s> isn't allowed here.` or `</%s> isn't allowed here.`? Or maybe that'd be confusing if the opening tag is really something like `<foo attr1="value1" ... attrn="valuen">` and the error message says `<foo>`

Maybe `Start tag '%s' isn't allowed here.` and `End tag '%s' isn't allowed here.` would be better?

stevecheckoway (approved):
This is obviously an improvement over what we have. LGTM.

gumbo-parser/src/error.c
flavorjones: Oooh, I like that. :+1:

flavorjones: Updated the PR.

Issue #2865, opened 2023-04-28
update to libxml2 2.11.0 or later


flavorjones:
See https://discourse.gnome.org/t/libxml2-2-11-0-released/15123 for release notes.

It doesn't look like anything in here affects nokogiri security concerns, so I'm planning to include this in a 1.15.0 release, but not a 1.14.x patch release. Opinions welcome.

Lots of yak shaving to try to make this release work ...

### Summary of blockers

- [x] work around windows atexit crashes: https://github.com/sparklemotion/nokogiri/pull/2871
- [x] waiting for a bugfix release of libxml2 (v2.11.2) with this fix: https://gitlab.gnome.org/GNOME/libxml2/-/issues/530
- [x] waiting for a bugfix release of libxml2 (v2.11.2) with this fix: https://gitlab.gnome.org/GNOME/libxml2/-/commit/d6882f6454258197916499a6546472b69baa63df
- [x] waiting for upstream release with this iconv fix: https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/216

### Summary of nonblocking items

- [x] waiting for a ruby_memcheck release with this fix: https://github.com/Shopify/ruby_memcheck/pull/16
  - https://github.com/Shopify/ruby_memcheck/releases/tag/1.3.2
- [x] test upstream libxml2 on windows: https://github.com/sparklemotion/nokogiri/pull/2868
- [x] test upstream libxml2 with memcheck: https://github.com/sparklemotion/nokogiri/pull/2868
- [x] report bug in new upstream schema analysis (not yet v2.11): https://gitlab.gnome.org/GNOME/libxml2/-/issues/531
- [x] fix compiler warnings in Nokogiri

### Details

- ☹ [ruby_memcheck is failing to parse valgrind XML files because libxml2 changed its buffering code](https://github.com/sparklemotion/nokogiri/actions/runs/4831357226/jobs/8614155369#step:7:4476)
  - 🤔 patch ruby_memcheck to use the HUGE parse option
    - ☑ https://github.com/Shopify/ruby_memcheck/pull/16
    - ☹ waiting for a release
      - 🤔 specify a git ref in Gemfile
        - ☹ not all CI images have git
          - 🤔 update CI docker images
            - ☑ https://github.com/sparklemotion/nokogiri/pull/2869

- ☹ ruby_memcheck is now flagging additional errors
  - 🤔 be proactive: integration-test upstream libxml2 with ruby_memcheck to detect regressions
[...]

PR #2866, opened 2023-04-28
dep: update libxml2 to v2.11.3, libxslt to v1.1.38

flavorjones:

#2865

Note that this PR is a work in progress, see #2865 for status of the overall work.

One test was modified to accommodate quote-escape behavior that changed.

No Nokogiri behaviors have changed, though obviously libxml2 bugfixes have changed some implementation-specific behaviors.


flavorjones:
OK, looks like there are two issues:

- symbol resolution on truffleruby
- segfault at exit on windows (again)

Digging in.

PR #2868, opened 2023-04-28
ci: upstream libxml2 testing: windows and ruby_memcheck coverage

flavorjones:

[libxml2 2.11.0 is failing windows builds](https://github.com/sparklemotion/nokogiri/pull/2866), and I think I probably could have caught the problem earlier if I was running the upstream builds on windows. This PR adds that integration testing.

libxml2 is also leaking memory, so this PR also adds `ruby_memcheck` testing on the upstream builds to try to detect these problems earlier.

### DIFF
--- a/.github/workflows/upstream.yml
+++ b/.github/workflows/upstream.yml
@@ -61,7 +61,6 @@ jobs:
       - run: "bundle exec rake test"
 
   xmlsoft-head-valgrind:
-    needs: ["xmlsoft-head"]
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/sparklemotion/nokogiri-test:mri-3.1
@@ -90,6 +89,7 @@ jobs:
       - name: "Compile against libxml2 and libxslt source directories"
         run: "bundle exec rake compile -- --with-xml2-source-dir=${GITHUB_WORKSPACE}/libxml2 --with-xslt-source-dir=${GITHUB_WORKSPACE}/libxslt"
       - run: "bundle exec rake test:valgrind"
+      - run: "bundle exec rake test:memcheck"
 
   ruby-head:
     strategy:
@@ -123,7 +123,6 @@ jobs:
       - run: bundle exec rake test
 
   ruby-head-valgrind:
-    needs: ["ruby-head"]
     strategy:
       fail-fast: false
       matrix:
@@ -146,6 +145,7 @@ jobs:
           key: ports-ubuntu-${{hashFiles('dependencies.yml', 'patches/**/*.patch')}}
       - run: bundle exec rake compile -- --${{matrix.sys}}-system-libraries
       - run: bundle exec rake test:valgrind
+      - run: bundle exec rake test:memcheck
 
   jruby-head:
     runs-on: ubuntu-latest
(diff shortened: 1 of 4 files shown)


MSP-Greg:
@flavorjones

Sorry for the delay.  I believe config.guess and config.sub need to be downloaded, I think ruby/ruby does so.  Not sure about the other items.  Build systems are not something I have much knowledge of.

Both files are 'small', so they could be added to the Windows 'build tool packages' that are installed by setup-ruby.  Not sure where to put them.

I checked MSY2 and MSFT vcpkg, and neither have updated to libxml2 2.11.0, so looking at their build scripts probably won't be very helpful...

flavorjones:
@MSP-Greg Thanks for responding, I appreciate your time. I have to apologize for not explaining better what I am having trouble with -- mostly because I'm not sure what's going wrong, so thanks for your patience.

I've tried explicitly installing [libtool](https://packages.msys2.org/package/libtool) which contains the files being logged as missing. And they're [still not being found](https://github.com/sparklemotion/nokogiri/actions/runs/4841862616/jobs/8628433301#step:7:576).

Do you know: is there some sort of `$PATH` shenanigans that I need to be setting so that these scripts and files get found?

My next step is going to be digging into where pacman installs the files and what the paths are in the environment, but the feedback loops are slow here so I was hoping you'd spot something silly that I'm doing or not doing.

flavorjones:
@MSP-Greg Apologies again, after some more digging I don't think this is related to the mingw packages, I think the problem is likely something in the Nokogiri build system that's causing the `--with-xml2-source-dir` flag to not work properly on windows. Sorry to have tagged you needlessly.

MSP-Greg:
@flavorjones

I think `'libtool` is already installed.

For the info on what packages are installed with setup-ruby, see

https://github.com/ruby/setup-msys2-gcc/actions/workflows/windows-build-tools.yml

The most recent workflow run shows jobs for MSYS2 and the various build archives that are available.  The 'msys2 tools' jobs show that `libtool` is already [installed](https://github.com/ruby/setup-msys2-gcc/actions/runs/4841600553/jobs/8627973865#step:5:174).

As to the other jobs, they're pairs of MINGW64 & UCRT64 gcc tools and ruby build dependencies, with the `-3.0` jobs containing current OpenSSL (now 3.1.0) and the other jobs containing the most recent OpenSSL 1.1.1 build.

Back to this issue, `pacman -Qo config.guess` should show which package 'owns' the file. Locally it reports `error: No package owns config.guess`, but as your link to the `libtool` info shows, they are installed locally on my system in the following folder:

C:/msys64/usr/share/libtool/build-aux

> Do you know: is there some sort of $PATH shenanigans that I need to be setting so that these scripts and files get found?

The ruby/ruby build scripts copy some of the files into the src folder, in a `tools` directory.  Not sure.

JFYI, I use nokogiri daily when updating https://msp-greg.github.io/.  I also always use ruby master/head/trunk, so I never use the pre-compiled gem.  This code has been in place for years.  I normally had the `--use-system-libraries` option set, but I just yanked it and installed/compiled nokogiri with the vendored code in the gem.  Installed fine, haven't run it yet.  Not sure if you're doing something similar in CI here...

flavorjones:
@MSP-Greg Just closing the loop with you -- see https://github.com/flavorjones/mini_portile/pull/126 for the underlying fix. It was a matter of converting a windows path to a posix path so the `configure` script can correctly detect the aux files (like `config.guess`).

PR #2873, opened 2023-05-07
dep: update xalan to 2.7.3

chadlwilson:

Upgrades xalan-J to 2.7.3, predominantly to resolve CVE-2022-34169 noise.

Changelog: https://xalan.apache.org/xalan-j/readme.html#notes_latest

This would be in the meantime, pending https://github.com/sparklemotion/nokogiri/issues/1829

Assume existing regression covers the library usage,

Should not change behaviour based on based reading of changelog. It does change the minimum Java version from 1.3 to Java 8 (!!!) but that shouldn't be an issue for nokogiri to my knowledge, as Java 8 is presumably the earliest supported version anyway.


flavorjones:
@chadlwilson Thanks for your attention and starting this work! If you can, please take a look at CONTRIBUTING.md which has a section on bumping the java dependencies. The jar files need to be checked into source control.

flavorjones:
@chadlwilson Never mind, I ran the additional steps and appended a commit. CI is running now.

flavorjones:
Can you look at the JRuby failures? Looks like an API changed out from under us.

chadlwilson:
Apologies @flavorjones - got lazy with contribution guide and was hoping for the build to do the talking 😅 will take a look.

chadlwilson:
After a quick initial look I can't quite see what has changed here that might have caused these changes in behaviour.

There are 2 failures and 3 errors, cataloguing here for now.
- 3x `Nokogiri::XSLT::Stylesheet#test_class_methods` / `Nokogiri::XSLT::Stylesheet#test_transform` / `Nokogiri::XSLT::Stylesheet#test_transform2:` > `Java::OrgApacheXmlSerializerUtils::WrappedRuntimeException: The Properties object passed to the SerializerFactory does not have a 'method' property`
  - https://github.com/sparklemotion/nokogiri/blob/805405910b0115b5d48ed68130d31e4f60fb4578/test/test_xslt_transforms.rb#L163
- ` Nokogiri::XSLT::Stylesheet#test_0001_should not crash when given XPath 2.0 features [/home/runner/work/nokogiri/nokogiri/test/test_xslt_transforms.rb:353] Minitest::Assertion: RuntimeError expected but nothing was raised.`
  - https://github.com/sparklemotion/nokogiri/blob/805405910b0115b5d48ed68130d31e4f60fb4578/test/test_xslt_transforms.rb#L312
  - `xs:decimal(..)` XPath 2 function no longer throws `java.lang.NoSuchMethodException.*decimal` as it used to?
- Exception handling for multiple roots
  ```
   Nokogiri::XSLT::TestExceptionHandling#test_java_exception_handling [/home/runner/work/nokogiri/nokogiri/test/xslt/test_exception_handling.rb:33]
  Minitest::Assertion: The exception message does not contain the expected information.
  Expected /Can't have more than one root/ to match "org.w3c.dom.DOMException: HIERARCHY_REQUEST_ERR: An attempt was made to insert a node where it is not permitted.".
  ```
  - https://github.com/sparklemotion/nokogiri/blob/071618a3915d557c71a577d3dfa8360e1eededf7/test/xslt/test_exception_handling.rb#L11-L27

Generally speaking:
* Xalan 2.7.3 does seem to include some patches/fixes that have been un-deployed/unreleased for over 10 years. Bit scary :-(
* There is some patching on nokogiri side in places like https://github.com/sparklemotion/nokogiri/blob/9d0bb8c4b95dac7b3ee1c8d97790dca7a38c9726/ext/java/nokogiri/internals/XalanDTMManagerPatch.java which may or may not require review/reassessment on newer Xalan

I'm not sure if I will have the expertise to look into these, so understand if you want to just kill this for now. Won't have much time for the next week or so most likely.

flavorjones:
I pushed some changes to the tests to address the behavioral changes. Need to look into the remaining three method-related errors.

flavorjones:
I just pushed another commit that sets the default output property "method" value if it's not set. I'm willing to bet this is a regression in Xalan, but it doesn't seem like the bug tracker is being maintained anymore, so working around it seems like a reasonable fallback. WDYT?

flavorjones:
Because this would be a security update, I've backported to v1.14.x at https://github.com/sparklemotion/nokogiri/pull/2875 and we'll see if that's green on CI.

flavorjones:
v1.14.4 has been released, with vendored xalan-j updated to 2.7.3. Thanks for your help!

chadlwilson:
Thanks for this @flavorjones ! Still not sure how you were able to apply your expertise to determine those failures were nothing to worry about (other than the output property), but I'll assume your knowledge is sufficient here 🙂

Yeah, your guess is as good as mine regarding the issue tracking. I hunted earlier and found a few stale things marked against 2.7.3 and stuff like https://issues.apache.org/jira/browse/XALANJ-2649 so they are using JIRA but there seems to be something screwy with the Jira issue filters making it a pain to navigate around.issues in the normal JIRA way.

The latest release also has broken Maven POM metadata (which doesn't affect nokogiri due to the vendoring) per the above issue which will limit uptake.

I believe the diff is https://github.com/apache/xalan-java/compare/xalan-j_2_7_2...xalan-j_2_7_3-rc10 given the vote was to release 2.7.3 off RC10 on the mailing list.

PR #2880, opened 2023-05-17
Move back to release verion of ruby_memcheck

peterzhu2118:
Shopify/ruby_memcheck#16 has been merged and released, so we can move back to released version of ruby_memcheck.


flavorjones:
I got to this before you! #2879.  ♥ ♥ ♥

I'm going to prefer that one because I don't want to revert the CI image changes.

PR #2885, opened 2023-05-20
ci: add mechanize to the downstream suite

flavorjones:

Mechanize just started failing its test suite on encoding errors. I suspect this has to do with the libxml2 update to 2.11.

Add mechanize to the downstream suite.


flavorjones:
Failing tests are a behavior change in libxml2, filed upstream here: https://gitlab.gnome.org/GNOME/libxml2/-/issues/543

flavorjones:
This is blocked for now on resolution of https://github.com/sparklemotion/mechanize/issues/613

PR #2888, opened 2023-05-23
Integrate nokogiri-xmlsec into nokogiri

maths22:

https://github.com/sparklemotion/nokogiri/discussions/2746#discussioncomment-5561385

Tests have been ported over from the separate library, and some that were skipped
there have been fixed to be passing.

This adds new functionality to the C implementation.  In the draft version it doesn't
touch Java. We will either need Java to throw useful notimplemented exceptions or
add the features of this library to the Java side, presumably using a third-party library.

** Additional Note **
Definitely needs some work on how the bundled xmlsec build works, and I think some #ifdefs to handle a range of xmlsec versions.  Also seems to occasionally segfault running the test suite, so that definitely needs resolving.


flavorjones:
Repeating what I said in #2547: @maths22 I appreciate you making the attempt to get xmlsec to work in JRuby, but I think the best first change we can make regarding xmlsec is simply to import the C code and tests. I'd prefer to defer tackling the introduction of JRuby support.

Issue #2887, opened 2023-05-22
[bug] java.lang.UnsupportedClassVersionError: org/cyberneko/dtd/DTDConfiguration


malcolm-pro:
We've been seeing this error under Java 8 starting with v1.14:

```
java.lang.UnsupportedClassVersionError: org/cyberneko/dtd/DTDConfiguration has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
```

That's a pretty common error when trying to run libraries compiled for Java 11 under Java 8. It looks like this is due to the newer [nekodtd jar](https://github.com/sparklemotion/nokogiri/blob/main/lib/nokogiri/jruby/org/nokogiri/nekodtd/0.1.11.noko1/nekodtd-0.1.11.noko1.jar). I might have missed something, but I don't see any minimum Java version requirements listed, so I've been assuming Nokogiri is still compatible with Java 8. If this is the case, maybe this jar needs to be recompiled under Java 8? If this is _not_ the case, and Nokogiri 1.14+ needs Java 11:

1) could that be listed in the main Readme?
2) is there a way to recompile just this library locally for Java 8 and use that in the gem?
3) if not, will `force_ruby_platform: true` in the Gemfile/gemspec avoid loading this JAR?

FWIW, JRuby did mention in the 9.3.0.0 release notes that they're considering dropping Java 8 support, and Java 8 is quite long in the tooth, but afaik Java 8 is still supported (and as of last year [was still quite widely used](https://newrelic.com/resources/report/2022-state-of-java-ecosystem)).

flavorjones:
@malcolm-pro Sorry you're having problems!

I will admit to a very low level of knowledge around Java version support in JRuby and in the vendored Java libraries.

That said, the vendored `nekodtd.jar` that's is built by the Nokogiri maintainers from https://github.com/sparklemotion/nekodtd, though -- see https://github.com/sparklemotion/nokogiri/issues/2437 and https://github.com/sparklemotion/nekodtd/issues/1 for context -- so I may be able to build it in such a way that it still supports Java 8.

cc @headius in case you're able to point me in the right direction.

flavorjones:
OK, so I rebuild nekodtd.jar with Java 8 and ran into more issues:

```
ext/java/nokogiri/internals/NokogiriXPathFunction.java:33: error: NokogiriXPathFunction is not abstract and does not override abstract method evaluate(List) in XPathFunction
[...]

PR #2889, opened 2023-05-23
java 8 support

flavorjones:

#2887 let us know that we lost Java 8 support with the new nekodtd jar in v1.14+.

This is an attempt to add CI coverage for Java 8, and then check if I can easily restore support by rebuilding nekodtd.

- [x] get the expected failure with JRE8 - https://github.com/sparklemotion/nokogiri/actions/runs/5060549469/jobs/9083793812
- [x] compile nekodtd with Java 8 and copy it into this PR branch
- [x] if that works, push the Java 8 nekodtd to maven central - https://github.com/sparklemotion/nekodtd/issues/2
- [ ] and then update this branch with that version


flavorjones:
Got the right failure here: https://github.com/sparklemotion/nokogiri/actions/runs/5060549469/jobs/9083793812

Issue #1430, opened 2016-02-17
JRuby nokogiri incorrectly include xml declaration for html transformation


jvshahid:
Using the following code:

```ruby
input_xml = <<-EOS
<?xml version="1.0" encoding="utf-8"?>
<report>
  <title>My Report</title>
</report>
EOS

input_xsl = <<-EOS
<?xml version="1.0" encoding="utf-8"?>
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
  <xsl:template match="/">
    <html>
      <head>
        <title><xsl:value-of select="report/title"/></title>
      </head>
      <body>
        <h1><xsl:value-of select="report/title"/></h1>
      </body>
    </html>
  </xsl:template>
</xsl:stylesheet>
EOS

require 'nokogiri'

xml = ::Nokogiri::XML(input_xml)
xsl = ::Nokogiri::XSLT(input_xsl)

puts xsl.apply_to(xml)
```

expected behavior:

```
<html>
<head>
<meta http-equiv="Content-Type" content="text/html; charset=UTF-8">
<title>My Report</title>
</head>
<body><h1>My Report</h1></body>
</html>
```

actual behavior:

```
<?xml version="1.0" encoding="UTF-8"?><html><head><title>My Report</title></head><body><h1>My Report</h1></body></html>
```

cbasguti:
Hey everyone! Just wanted to give you a heads up that I'm actively working on this issue and putting in my best effort to find a solution.

stevecheckoway:
I'm not an expert on xslt, but shouldn't the method be set by an `xsl:output` element?

stevecheckoway:
To follow up on this, I _think_ our jruby XSLT processor is incorrectly determining the default method. Here's what the [standard](https://www.w3.org/TR/1999/REC-xslt-19991116)  has to say:
> The default for the method attribute is chosen as follows. If
>
>    - the root node of the result tree has an element child,
>
>    - the expanded-name of the first element child of the root node (i.e. the document element) of the result tree has local part html (in any combination of upper and lower case) and a null namespace URI, and
>
>    - any text nodes preceding the first element child of the root node of the result tree contain only whitespace characters,
>
> then the default output method is html; otherwise, the default output method is xml. The default output method should be used if there are no xsl:output elements or if none of the xsl:output elements specifies a value for the method attribute.

[...]

PR #2903, opened 2023-06-10
Avoid incorrectly including xml header in xslt output on JRuby

cbasguti:

- In the updated function, the line props.setProperty(OutputKeys.METHOD, org.apache.xml.serializer.Method.UNKNOWN); was changed to props.setProperty(OutputKeys.METHOD, "html");.
- This change ensures that the METHOD property is set to "html" instead of the unknown method. This modification aims to address issue #1430 by explicitly specifying the method as "html" during serialization.

A new test file named test_xml_header.rb was added .

The modification made to the Java implementation was to exclude the XML declaration tag (<?xml ... ?>) from the output. This was achieved by adjusting the code in the serialize function to remove the inclusion of the XML declaration. The purpose was to align the output with the expected format, where the XML declaration should not be present.

### DIFF
--- a/ext/java/nokogiri/XsltStylesheet.java
+++ b/ext/java/nokogiri/XsltStylesheet.java
@@ -40,6 +40,7 @@ import org.jruby.runtime.Helpers;
 import org.jruby.runtime.ThreadContext;
 import org.jruby.runtime.builtin.IRubyObject;
 import org.w3c.dom.Document;
+import org.w3c.dom.Node;
 
 import nokogiri.internals.NokogiriXsltErrorListener;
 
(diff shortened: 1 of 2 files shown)


stevecheckoway:
Html is not obviously the correct method when none is specified. If this is the correct default, I think you should add a comment explaining why it's correct.

stevecheckoway:
E.g., does this make the jruby implementation match the C implementation? (I'm on my phone so I cannot easily check.)

stevecheckoway:
After digging into this a bit, I don't think this is the correct fix. The XSLT standard specifies how the default method should be determined—either XML or HTML—when none is specified by the stylesheet.

For some reason, this is not being correctly detected by Xalan. (The code that's supposed to do that is [here](https://github.com/apache/xalan-j/blob/660e12dca29e28df9f3f9ebb7c19435bb16362f3/src/org/apache/xml/serializer/ToUnknownStream.java#L1112), I think.)

If I modify the code in #1430 slightly to call `#transform` and print the root element's name and namespace, it says its name is `html` and it has no namespace. I'm not sure why Xalan is using the XML output method rather than the HTML output method

Here's the code.
```ruby
input_xml = <<-EOS
<?xml version="1.0" encoding="utf-8"?>
<report>
  <title>My Report</title>
</report>
EOS

input_xsl = <<-EOS
<?xml version="1.0" encoding="utf-8"?>
<xsl:stylesheet version="1.0" xmlns:xsl="http://www.w3.org/1999/XSL/Transform">
  <xsl:template match="/">
    <html>
      <head>
        <title><xsl:value-of select="report/title"/></title>
      </head>
      <body>
        <h1><xsl:value-of select="report/title"/></h1>
      </body>
    </html>
  </xsl:template>
</xsl:stylesheet>
EOS

require 'nokogiri'

xml = Nokogiri::XML(input_xml)
xsl = Nokogiri::XSLT(input_xsl)
html = xsl.transform(xml)

puts(html.root.name)
puts(html.root.namespace || "no namespace")
```

cbasguti:
Thank you for your feedback and for providing additional insights into the issue! Based on your comments, it appears that determining the correct default method (XML or HTML) when it is not explicitly specified can be a complex matter. As you said, it seems that Xalan is not correctly detecting the default method according to the XSLT standard.

Considering this, I propose handling the default method determination within the serialize method itself. By doing so, we can ensure that the output format aligns with the expected behavior, especially when the method is not explicitly specified. Like this:

flavorjones:
I'm going to close this, and we can continue the conversation at #1430

flavorjones:
Oh, apologies, I see you pushed additional changes that are trying to implement what Steve is describing ... reopening

flavorjones:
@cbasguti I appreciate the time and effort you've put into trying to fix this issue. However, I'm not confident that we understand what's going on enough to start working around it by writing our own Java code.

If it's a bug in Xalan, we should report it upstream and/or submit a patch there.

If it's not a bug in Xalan, then it may be that we're using the library incorrectly and so the fix to Nokogiri should then be a usage fix to how we're calling the libraries.

I'll leave this open for now, but let's continue the conversation in #1430.

Issue #2923, opened 2023-07-06
[bug] Crash while acquiring the memsize of nodes


peterzhu2118:
**Please describe the bug**

When acquiring the memsize of nodes, nokogiri crashes with the following stack trace:

```
/opt/rubies/ruby-3.2.2/lib/ruby/3.2.0/objspace.rb:102: [BUG] Segmentation fault at 0x0000000000000041
ruby 3.2.2 (2023-03-30 revision e51014f9c0) [x86_64-linux]

-- Control frame information -----------------------------------------------
c:0004 p:---- s:0025 e:000024 CFUNC  :_dump_all
c:0003 p:0124 s:0017 e:000016 METHOD /opt/rubies/ruby-3.2.2/lib/ruby/3.2.0/objspace.rb:102
c:0002 p:0017 s:0006 e:000005 BLOCK  /home/spin/src/github.com/Shopify/nokogiri/test/helper.rb:21 [FINISH]
c:0001 p:0000 s:0003 E:000250 DUMMY  [FINISH]

-- Ruby level backtrace information ----------------------------------------
/home/spin/src/github.com/Shopify/nokogiri/test/helper.rb:21:in `block in <top (required)>'
/opt/rubies/ruby-3.2.2/lib/ruby/3.2.0/objspace.rb:102:in `dump_all'
/opt/rubies/ruby-3.2.2/lib/ruby/3.2.0/objspace.rb:102:in `_dump_all'

-- Machine register context ------------------------------------------------
 RIP: 0x00007fa029564bc7 RBP: 0x0000000000000031 RSP: 0x00007fff3cf76a10
 RAX: 0x0000000000000001 RBX: 0x0000000000000031 RCX: 0x000000000000000d
 RDX: 0x00007fa02efd2150 RDI: 0x0000000000000031 RSI: 0x000000000000000c
  R8: 0x0000562d2d49a780  R9: 0x0000000000000015 R10: 0x0000000000000000
 R11: 0x000000000000000c R12: 0x0000000000000001 R13: 0x00007fa022d8e978
 R14: 0x0000562d24ec0228 R15: 0x00007fff3cf76da0 EFL: 0x0000000000010202

-- C level backtrace information -------------------------------------------
/opt/rubies/ruby-3.2.2/lib/libruby.so.3.2(rb_print_backtrace+0xd) [0x7fa02eea4c3f] /tmp/ruby-3.2.2/vm_dump.c:785
[...]

PR #2924, opened 2023-07-06
fix memsize_node when called on xmlAttrs

stevecheckoway:
The `properties` field of an `xmlNode` element points to an `xmlAttr`. The first few fields of `xmlAttr` are in common with `xmlNode`, but not the `properties` field which doesn't exist in an `xmlAttr`.

The `memsize_node` function was passing an `xmlAttr` to a recursive call and then trying to do the same with the properties of that.

This led to type confusion and subsequent crashes.

Fixes: #2923

#2923

No. I don't know how to test this behavior. I'd appreciate help with this.

It fixes a crash in the C implementation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 
 ## next / unreleased
 
-...
+* [CRuby] `ObjectSpace.memsize_of` is now safe to call on `Document`s with complex DTDs. In previous versions, this debugging method could result in a segfault. [[#2923](https://github.com/sparklemotion/nokogiri/issues/2923), [#2924](https://github.com/sparklemotion/nokogiri/issues/2924)]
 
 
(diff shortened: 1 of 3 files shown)


flavorjones:
cc @peterzhu2118 Can you kick the tires on this when you get a moment?

peterzhu2118:
Thank you for finding the fix so quickly! I can confirm that it no longer crashes with this patch.

flavorjones:
I'd really like to have a narrow test case that fails (even if it's just in valgrind). Going to play with this a bit before merging.

flavorjones:
OK, inspecting the crash in the debugger shows it's because either there's an `xmlDTD` node, or an `xmlEntity` node within the DTD. `xmlDTD` has `attributes` where `xmlNode` has `properties; and `xmlEntity` has `length`.

Anyway. I added a test that creates a DTD with an attribute just to demonstrate the bug, and prevent the unlikely regression.

PR #2926, opened 2023-07-10
internal: remove mark function from XML::Document

flavorjones:

remove mark function from XML::Document. originally introduced in dd07ce86, I don't think we need it anymore since cb1557d0 introduced TypedData conventions to support compaction.

cc @tenderlove and @byroot for a gut check

existing coverage should be sufficient to detect if there's an issue with compaction.

No behavior changes.

### DIFF
--- a/ext/nokogiri/xml_document.c
+++ b/ext/nokogiri/xml_document.c
@@ -51,17 +51,6 @@ remove_private(xmlNodePtr node)
 }
 
 static void
-mark(void *data)
-{
-  xmlDocPtr doc = (xmlDocPtr)data;
-  nokogiriTuplePtr tuple = (nokogiriTuplePtr)doc->_private;
-  if (tuple) {
-    rb_gc_mark(tuple->doc);
-    rb_gc_mark(tuple->node_cache);
-  }
-}
-
-static void
 dealloc(void *data)
 {
   xmlDocPtr doc = (xmlDocPtr)data;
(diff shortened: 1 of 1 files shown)


ext/nokogiri/xml_document.c
```diff
@@ -50,17 +50,6 @@ remove_private(xmlNodePtr node)
...
 }
 
-static void
-mark(void *data)
-{
-  xmlDocPtr doc = (xmlDocPtr)data;
-  nokogiriTuplePtr tuple = (nokogiriTuplePtr)doc->_private;
-  if (tuple) {
-    rb_gc_mark(tuple->doc);
-    rb_gc_mark(tuple->node_cache);
```
byroot, line 60: So if I understand correctly, these marks aren't necessary to prevent the objects being GCed, because one is a reference to self, and another is held some other way.

However they were added so the objects would be pinned.

Up until this point it checks out, but AFAIK we didn't add any `dcompact` functions, did we? At least not in the commit you linked. So we either need to add a compaction callback to update the references, or we need to keep them pined.

flavorjones: @byroot That's correct, they were being marked so that they wouldn't get relocated.

:thinking: Ahh, you're right, we do need to implement dcompact, my bad.

flavorjones: Although: the reason this doesn't segfault now is, I think, because a `Node` calls `rb_gc_mark` on its `Document` owner, essentially keeping it pinned anyway.

Issue #2928, opened 2023-07-14
ci: ruby_memcheck reporting leaks on ruby-head


flavorjones:
e.g., https://github.com/sparklemotion/nokogiri/actions/runs/5551865569/jobs/10138487549

I'm creating this issue primarily to communicate that this is a known problem.

Leaks are reported due to the problem with inline caches described at https://github.com/Shopify/ruby_memcheck/issues/23. This is the root cause of the failures recorded at https://github.com/sparklemotion/nokogiri/issues/2838.

Recently, the removal of the transient heap in ruby head is resulting in these leaks being reported much more frequently.

I've been in touch with Peter, the maintainer of ruby_memcheck, who's got some ideas for a Ruby feature that will free up all memory at shutdown, including the inline caches.

In the meantime, I may just add suppressions for Ruby 3.3.

flavorjones:
Suppressions added in #2929

flavorjones:
Context: https://github.com/Shopify/ruby_memcheck/issues/23#issuecomment-1640748018

PR #2929, opened 2023-07-16
Add suppressions for ruby-3.3 inline cache references

flavorjones:

See #2928 and https://github.com/Shopify/ruby_memcheck/issues/23


flavorjones:
Watching upstream CI: https://github.com/sparklemotion/nokogiri/actions/runs/5569507609

PR #2943, opened 2023-08-04
Bump ruby_memcheck to 2.0.0

peterzhu2118:

Bumps ruby_memcheck to 2.0.0 which includes several breaking changes, one of which being that `binary_name` is no longer used in `RubyMemcheck.config`.

The new ruby_memcheck does not work well with SimpleCov, so it's disabled when ruby_memcheck is running. 

N/A.

No.


flavorjones:
Thank you!

Issue #2838, opened 2023-04-02
[journal] tracking (likely spurious) `ruby_memcheck` errors in CI


flavorjones:
On occasion, the `ruby_memcheck` job in CI raises a failure that isn't obviously fixable within Nokogiri.

My memory is soft, so I'm creating this issue to track these when they show up to see if there's a pattern that we can either

1. identify as an issue that is fixable within Nokogiri
2. if it's something we can ignore (e.g., a memory pool within Ruby) then we can updated `ruby_memcheck` to skip that function / those functions.

I really feel like I need to collect more data before I go back to Peter with a PR for `ruby_memcheck`, so this is where we'll keep that history for now.

flavorjones:
https://github.com/sparklemotion/nokogiri/actions/runs/4580734520/jobs/8089726674#step:7:4483

```
160 bytes in 1 blocks are definitely lost in loss record 29,850 of 40,713
  malloc (vg_replace_malloc.c:307)
  objspace_xmalloc0 (gc.c:11462)
  ary_heap_alloc (array.c:329)
  ary_heap_alloc (array.c:320)
  ary_resize_capa (array.c:453)
  ary_double_capa (array.c:507)
  ary_ensure_room_for_push (array.c:654)
  rb_ary_push (array.c:1311)
 *__xmlRaiseError (error.c:604)
 *htmlParseErr (HTMLparser.c:101)
 *htmlParseEndTag (HTMLparser.c:4201)
 *htmlParseContentInternal (HTMLparser.c:4742)
 *htmlParseDocument (HTMLparser.c:4998)
 *htmlDoRead (HTMLparser.c:6921)
 *rb_html_document_s_read_memory (html4_document.c:111)
  vm_call_cfunc_with_frame (vm_insnhelper.c:3037)
  vm_sendish (vm_insnhelper.c:4751)
```

My memory is soft, but I think the past warnings have also been this. It seems like we're appending to the `errors` array while parsing a file, and ruby increases the capacity of the array to do do.

This isn't obviously something that Nokogiri can do anything about, and my suspicion is that we should consider adding `rb_double_capa` (or a related function) to `ruby_memcheck`'s `DEFAULT_SKIPPED_RUBY_FUNCTIONS`.

flavorjones:
Got another one in https://github.com/sparklemotion/nokogiri/actions/runs/5003753830/jobs/8965429361?pr=2879

```
[...]

PR #2944, opened 2023-08-04
Remove suppressions for Ruby 3.3

peterzhu2118:

The issues were fixed in ruby/ruby@8120.

N/A.

No.


flavorjones:
Nice! Thank you, I totally missed that upstream commit.

flavorjones:
For posterity: #2928 introduced these suppressions, and more historical background is available at #2838.

PR #2950, opened 2023-08-10
WIP: use ragel options to minimize states and use gotos

jtarchie:

No problem.

Experimenting with the features of `ragel` for faster parsing tables for `gumbo`.
I'd like to see if the tests passed through CI.

This shouldn't affect the functionality.
This relies on `ragel` and the C compiler for better optimizations in parsing.

As far as I know, just the C implementation.


stevecheckoway:
It's nice to see someone looking at this!

I played with the various options a few years ago and couldn't get better output. Looks like this choice of options ends up with 50k more lines of code to compile.

I'm curious how this choice impacts the size of the binary, compilation time, and runtime parsing performance.

flavorjones:
Thanks, @jtarchie! I kicked off CI and have an HTML5 benchmark on my dev machine that I'll run a comparison with.

jtarchie:
There are file differences:
- `char_ref.c` goes from 692K -> 1.4M in size
- the object file goes from 298K -> 870K

I imagine the object file shows the size difference.
This may not give the speed increase needed, based on the changelog ragel originally made this faster. This might have already hit the speed sweet spot. It might not be enough of bottleneck to make more things faster.

jtarchie:
=Is this something I can help explore still?

flavorjones:
@jtarchie Thanks for asking. I definitely want to understand the tradeoffs between size, compilation speed (less of an issue these days), and runtime performance. It's on my list of things to explore next month when I have some free time to work on OSS.

flavorjones:
Rebased.

flavorjones:
@stevecheckoway Do you know why Craig used gperf for foreign attributes but ragel for these character references? They seem very similar in purpose.

stevecheckoway:
@flavorjones No, I'm afraid I don't know for certain.

If I had to guess, I'd say it was likely because attributes and tag names are completely known to the parser by the time the lookup occurs whereas that isn't the case for character references. For example, given the input `<feImage href="mdn_logo_only_color.png" />` (taken from [here](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/feImage). By the time the parser calls `gumbo_normalize_svg_tagname()`, we know the tag name is `feimage`. Gperf lets us look that up in the hash table with exactly one probe.

In contrast, `match_named_char_ref()` is called by the tokenizer and needs to match the strings character by character. So rather than a hash look up, it's a state machine.

We can't do something like find everything of the form `&blah;` because some entities don't need semicolons.

flavorjones:
@stevecheckoway See https://github.com/sparklemotion/nokogiri/pull/3402 for removal of `normalize_svg_tagname`

jtarchie:
This does not make things faster. It is such a specific use case of ragel, after discussion with @flavorjones, should either be replaced with hash-map lookup or completely rewritten. Looks like it is old code no one ever touched again. :-)

stevecheckoway:
That's an interesting idea. You'd have to benchmark it to see how well it performs.

From a quick scroll through the list of named character references, some are as long as 25 characters. An algorithm like when we see an &, perform hash look ups on the next 3 through 25 characters seems like it's unlikely to be a performance win over the state machine approach. The state machine only has to perform table lookups for each character so long as it's a prefix of a valid named character reference.

But that's just a guess and anyway it doesn't mean a more clever algorithm couldn't do better.

PR #2977, opened 2023-09-13
[draft] use mini_portile 2.9.0.rc1 to simplify extconf

flavorjones:

Simplifying the extconf by using the new `mkmf_config` feature in mini_portile 2.9.0

Existing test coverage is sufficient to ensure nothing's broken

N/A


flavorjones:
Welp, the BSD build is actually a legit failure -- looks like it's dynamically linking the system shared libxml2 library rather than statically linking the packaged libxml2 library. Need to investigate what's going on there.

flavorjones:
Update: I've reproduced the FreeBSD failure using system ruby on freebsd 13.2, but it doesn't reproduce if I use a ruby installed via rbenv/ruby-build.

flavorjones:
OK, reproduced with a system ruby on an Ubuntu image: https://github.com/sparklemotion/nokogiri/actions/runs/6209371364/job/16856521387?pr=2977

PR #2991, opened 2023-09-19
dep: ruby_memcheck to 2.2.0

flavorjones:

Update ruby_memcheck to 2.2.0 and set binary_name which should help with false positives


flavorjones:
Also running `upstream` ci: https://github.com/sparklemotion/nokogiri/actions/runs/6236229101

PR #3008, opened 2023-10-17
Update psych version to fix JRuby load issues

stefannibrasil:
Psych 5.1.1 fails to load on JRuby 9.x issue has been solved on the latest Psych release:

https://github.com/ruby/psych/releases/tag/v5.1.1.1

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -44,4 +44,4 @@ group :rdoc do
   gem "rdoc", "6.5.0"
 end
 
-gem "psych", "!= 5.1.1", "!= 5.1.0" # https://github.com/ruby/psych/issues/655
+gem "psych", "5.1.1.1"


stefannibrasil:
@flavorjones I saw you added a commit locking psych version. There's been a release fixing the issue. This PR changes the version so we get CI passing here:

https://github.com/flavorjones/loofah/pull/275

flavorjones:
Hi, @stefannibrasil. Thank you for opening this!

Commit 59e4981a didn't lock Psych to a particular version, it instead avoided the two known-bad versions. If there's a new release that's fixed, then I want to just remove it from the Gemfile since it's an implicit dependency.

flavorjones:
See https://github.com/sparklemotion/nokogiri/pull/3011

stefannibrasil:
ahhh thank you for sharing that! I appreciate you taking the time to explain it. I will rebase the loofah's PR with main. Thank you!

PR #3014, opened 2023-10-29
dev: handle upcoming libxml2's use of const for xmlError pointers

flavorjones:

upcoming libxml 2.12.0 will use `const xmlError *` as the data type for many function signatures where it's `xmlErrorPtr` today.

let's update our code to use the new convention.

see https://gitlab.gnome.org/GNOME/libxml2/-/commit/45470611b047db78106dcb2fdbd4164163c15ab7 and other upstream commits


flavorjones:
cherry-picking this patch into #3026

PR #3018, opened 2023-11-10
fix: strip spaces from the operator when translating nth css to xpath

fusion2004:

Prevent errors when nokogiri is translating nth css to xpath and there are spaces in it. (E.g. `2n + 1` instead of `2n+1`)

Background: This came up because prettier started rewriting my email css from `li:nth-child(2n+1)` to `li:nth-child(2n + 1)` (which is perfectly valid in browsers) and this caused a `expected an+b node to have either + or - as the operator, but is " + " (ArgumentError)` error from https://github.com/Mange/roadie using nokogiri as part of inlining the css.

I believe so!

Since I only modified ruby code, I'm assuming not?


flavorjones:
@fusion2004 Thank you for this! Looks good, merging.

PR #3022, opened 2023-11-14
fix bug of http redirection limit

xiaohui-zhangxh:

Fix the bug when http rediretion limit equal to 1, it's not stopped.

No.

No.


flavorjones:
@xiaohui-zhangxh Thank you for submitting this fix!

Unfortunately this method has been deprecated since v1.12.0 and will be removed in the next version of Nokogiri. If you're using `HTML5.get` please migrate to another HTTP client library.

PR #3025, opened 2023-11-17
dep: update libxml to 2.11.5 and libxslt to 1.1.39 (v1.15.x)

flavorjones:
**What problem is this PR intended to solve?**

update libxml to 2.11.5 and libxslt to 1.1.39

- https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.11.6
- https://gitlab.gnome.org/GNOME/libxslt/-/releases/v1.1.39

PR #3026, opened 2023-11-17
dep: update libxml to 2.12.0 (with explicit `-lpthread` linkline)

### LINKED ISSUE #3031: Update to libxml 2.12.0 or later
Release notes are at https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.12.0

I don't see any security-related changes, so I'm planning this for a Nokogiri v1.16.0 relesae that will coincide with Ruby 3.3 support.

I'm seeing some challenges with a naive upgrade, so I'm creating this issue to track the conversation. I'll keep this description up to date.

---

### the situation

libxml 2.12 makes some changes related to thread-local storage, primarily in [this upstream commit](https://gitlab.gnome.org/GNOME/libxml2/-/commit/c19771c1f13de9196f98260d142d8c8672eb5733). These changes are similar in nature to the changes in 2.11 made that update challenging (see #2865 for context), having to do with weak references to pthread symbols combined with the fact that [libpthread was merged into glibc in glibc v2.34](https://developers.redhat.com/articles/2021/12/17/why-glibc-234-removed-libpthread).

### the naive first attempt

Simply updating libxml2 will result in errors while building libxml2, where the linker tries to resolve pthread symbols while linking `xmllint`.

### resolving pthread

Trying to explicitly resolve libpthread:

- https://github.com/sparklemotion/nokogiri/pull/3026

But we find another problem when installing precompiled native gems on musl, which is that `nokogiri.so` references `ld-linux-x86_64.so.2` as a "needed" dependency.

Is this problem related? Probably, but it's not clear what's going on.

I could continue this approach by finishing the work to support musl in rake-compiler dock started at https://github.com/rake-compiler/rake-compiler-dock/issues/75, but 

1. I'm not sure that's the right thing to do
2. that approach couples two significant changes together, each of which has their own risk factors, which feels doubly risky.

### upstream fix?

Is this a problem with upstream libxml2? I actually think it is ...

I've submitted upstream fixes:

- [fix: pthread weak references in globals.c (!229) · Merge requests · GNOME / libxml2 · GitLab](https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/229)
- [fix: more pthread weak references in globals.c (!230) · Merge requests · GNOME / libxml2 · GitLab](https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/230)

Applying those patches locally:

- https://github.com/sparklemotion/nokogiri/pull/3032

But now we're once again still pulling in [`libwinpthread` on windows](https://github.com/sparklemotion/nokogiri/actions/runs/6916120729/job/18815839251) and [`ld-linux` on linux](https://github.com/sparklemotion/nokogiri/actions/runs/6916120729/job/18815839735#step:4:1192) ... there's something else going on here.

### thread local storage

Looking at the compiled object files in `libxml2.a`, I can see that `__tls_get_addr` is a new "unresolved" dependency which, on glibc systems in `ld-linux.so` (which doesn't exist on musl) and presumably on windows does something similar to cause `libwinpthread`. I'm doing some reading about thread local storage here:

https://maskray.me/blog/2021-02-14-all-about-thread-local-storage

The symbol is being referenced because we're using `-fPIC`, and it seems like this is a reference that cannot be made weak and is libc-dependent (meaning it differs between musl and glibc).

Configuring libxml2 `--without-tls` avoids this issue, which is now what #3032 does.

flavorjones:

Closes #3031

update libxml to 2.12.0 and libxslt to 1.1.39, though note that libxslt was also previously updated to 1.1.39 on the v1.15.x branch (#3025)

- https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.12.0
- https://gitlab.gnome.org/GNOME/libxslt/-/releases/v1.1.39

This approach attempts to explicitly link with `-lpthread` to handle libxml2 changes to thread local storage. See #3031 for more discussion.


flavorjones:
Need to rebase once https://github.com/sparklemotion/nokogiri/pull/3028 is merged

flavorjones:
OK, so just a summary of what's going on in this PR as of 985d15bb ...

- libxml 2.12.0 does some new things with respect to threads (see https://gitlab.gnome.org/GNOME/libxml2/-/commit/c19771c1f13de9196f98260d142d8c8672eb5733)
- this is causing problems on systems with glibc < v2.34 (which merged libpthread into libc, see https://developers.redhat.com/articles/2021/12/17/why-glibc-234-removed-libpthread)
- 985d15bb attempts to explicitly link with libpthread to handle both pre- and post-merge glibc
- but this causes an issue installing precompiled gems on musl systems

The error on musl systems is:

```
Error loading shared library ld-linux-x86-64.so.2: No such file or directory
```

which I think is being pulled in by explicitly linking to libpthread on a glibc system.

So I can think of two ways forward now:

- complete the work started at https://github.com/rake-compiler/rake-compiler-dock/issues/75 to support musl platforms
- back off this PR and instead implement another patch to vendored libxml2

The musl support feels a little risky, though, and might require bundler changes? I just don't know. Which makes me want to fall back to patching libxml2 and tackling musl support separately. :thinking:

flavorjones:
Closing because #3032 is a more promising approach.

PR #3027, opened 2023-11-17
ci: add stimulus_reflex to downstream suite

### LINKED ISSUE #3024: Add downstream test coverage for stimulus reflex
https://github.com/stimulusreflex/stimulus_reflex

cc @marcoroth

flavorjones:

Closes #3024

cc @marcoroth

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -51,6 +51,9 @@ jobs:
           - url: https://github.com/sparklemotion/mechanize
             name: mechanize
             command: "bundle exec rake test"
+          - url: https://github.com/stimulusreflex/stimulus_reflex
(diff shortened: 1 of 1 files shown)


.github/workflows/downstream.yml
```diff
@@ -51,6 +51,9 @@ jobs:
           - url: https://github.com/sparklemotion/mechanize
             name: mechanize
             command: "bundle exec rake test"
+          - url: https://github.com/stimulusreflex/stimulus_reflex
+            name: stimulus_reflex
+            command: "bundle exec rake test"
```
marcoroth, line 56: In case it's causing any issues: you might want to only run `bundle exec rake ruby_test` since the default rake task also includes the JavaScript tests.

PR #3032, opened 2023-11-18
dep: libxml 2.12.0 upgrade (with libxml2 patch for weakref)

flavorjones:

#3031

This approach uses the patches submitted upstream at:

- https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/229
- https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/230

and turns off libxml2's thread-local storage feature, which is an issue when precompiling for musl on glibc systems. If and when we can build a separate musl native gem, we can revisit this decision.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,7 +110,7 @@ jobs:
         if: matrix.sys == 'disable'
         with:
           path: ports
-          key: ports-ubuntu-${{matrix.ruby}}-${{hashFiles('dependencies.yml', 'patches/**/*.patch')}}
+          key: ports-ubuntu-${{matrix.ruby}}-${{hashFiles('dependencies.yml', 'patches/**/*.patch', 'ext/nokogiri/extconf.rb')}}
       - run: bundle install --local || bundle install
       - run: bundle exec rake compile -- --${{matrix.sys}}-system-libraries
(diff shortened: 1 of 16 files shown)


flavorjones:
OK, so even with the patches applied, we're still pulling in `ld-linux-x86-64.so.2` because the new thread local storage feature ("TLS") results in a reference to `__tls_get_addr` that is non-weak. You can read more about this at https://maskray.me/blog/2021-02-14-all-about-thread-local-storage

Turning off TLS by configuring libxml2 with `--without-tls` avoids this problem. We'll try that next.

flavorjones:
CI didn't kick off, so the manual run is https://github.com/sparklemotion/nokogiri/actions/runs/6916120729

flavorjones:
OK, so even with the patches applied, we're still pulling in `ld-linux-x86-64.so.2` because the new thread local storage feature ("TLS") results in a reference to `__tls_get_addr` that is non-weak. I am assuming something similar is happening on windows with `libwinpthread`.

You can read more about TLS at https://maskray.me/blog/2021-02-14-all-about-thread-local-storage

Turning off TLS by configuring libxml2 with `--without-tls` avoids this problem. We'll try that next.

flavorjones:
This seems like the winner.

PR #3007, opened 2023-10-16
added fuzzing support

fuzzy-boiii23a:
This just adds fuzzing support such that if any changes are made to libgumbo it's easy to fuzz these changes before merging something into the main branch.

PR #3036, opened 2023-11-20
fix: gumbo memory leak on abandoned tags

flavorjones:

The fuzzer in #3007 found a memory leak when incomplete tags are encountered.

The repro added to the memory leak test suite is:

```html
<asdfasdfasdfasdfasdfasdfasdfasdfasdfasdf foo="bar
```

which will leak the tag name without this fix.

cc @stevecheckoway 

Yes, but added it to the memory leak suite which can only be run manually today (see #1603 for context on plans to improve it).

No functional change.

### DIFF
--- a/gumbo-parser/src/tokenizer.c
+++ b/gumbo-parser/src/tokenizer.c
@@ -506,6 +506,7 @@ static void abandon_current_tag(GumboParser* parser) {
   for (unsigned int i = 0; i < tag_state->_attributes.length; ++i) {
     gumbo_destroy_attribute(tag_state->_attributes.data[i]);
   }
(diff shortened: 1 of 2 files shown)


stevecheckoway (approved):
LGTM

stevecheckoway:
I'm a little surprised this bug is there. It must have been introduced after the last time I fuzzed nokogumbo. I appreciate the work @fuzzy-boiii23a has done here.

flavorjones:
@stevecheckoway maybe in 5c6af1e7?

flavorjones:
Closed https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=63319

PR #3038, opened 2023-11-26
ext: ruby 3.3 support

flavorjones:

Add precompiled native support for Ruby 3.3.

Note that this is still experimental! It is precompiling using a rake-compiler-dock prerelease that uses 3.3.0-preview3.

Testing against 3.3.0-preview3 or head

N/A

### DIFF
--- a/.cross_rubies
+++ b/.cross_rubies
@@ -30,3 +30,11 @@
 3.2.0:x86-mingw32
 3.2.0:x86_64-darwin
 3.2.0:x86_64-linux
+3.3.0:aarch64-linux
+3.3.0:arm-linux
+3.3.0:arm64-darwin
+3.3.0:x64-mingw-ucrt
+3.3.0:x86-linux
+3.3.0:x86-mingw32
+3.3.0:x86_64-darwin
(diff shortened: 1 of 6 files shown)


flavorjones:
I've updated this branch to use rake-compiler-dock v1.4.0.rc1

flavorjones:
Please note this won't get released until one of two things happens:

- the Ruby Core team freezes the ABI (binary interface) for Ruby 3.3 (see https://bugs.ruby-lang.org/issues/19980)
- Ruby 3.3.0 final release arrives

agrobbin:
@flavorjones looks like with the [release of 3.3 RC1 this morning](https://www.ruby-lang.org/en/news/2023/12/11/ruby-3-3-0-rc1-released/), the ABI was frozen!

> After the release of RC1, we will avoid introducing ABI incompatibilities wherever possible. If we need to do, we’ll announce it in the release note.

With that, the corresponding [issue](https://bugs.ruby-lang.org/issues/19980) you linked above has been closed.

agrobbin:
Thanks @flavorjones!!

PR #3093, opened 2024-01-13
maintain oss-fuzz target building upstream

fuzzy-boiii23a:
This pull request is needed so that i can update the oss-fuzz builder located at https://github.com/fuzzy-boiii23a/oss-fuzz/tree/master/projects/nokogiri which makes it much more seamless to add new fuzz targets in this nokogiri repo which will automatically be pulled by oss-fuzz and fuzzed instead of maintaining the fuzzer in the oss-fuzz repo.

### DIFF
--- a/gumbo-parser/fuzzer/build-ossfuzz.sh
+++ b/gumbo-parser/fuzzer/build-ossfuzz.sh
@@ -1,0 +1,7 @@
+cd src
+make
+cd ../
+
+$CXX $CXXFLAGS -o $OUT/parse_fuzzer fuzzer/parse_fuzzer.cc src/libgumbo.a $LIB_FUZZING_ENGINE
+cp fuzzer/gumbo.dict $OUT/parse_fuzzer.dict
+cp fuzzer/gumbo_corpus.zip $OUT/parse_fuzzer_seed_corpus.zip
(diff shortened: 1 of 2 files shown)


fuzzy-boiii23a:
For example: https://github.com/fuzzy-boiii23a/oss-fuzz/commit/9d656738cca2c41f321cf8b133915089d6e85ac5

gumbo-parser/fuzzer/build-ossfuzz.sh
```diff
@@ -0,0 +1,7 @@
+cd src
+make
+cd ../
+
+$CXX $CXXFLAGS -o $OUT/parse_fuzzer fuzzer/parse_fuzzer.cc src/libgumbo.a $LIB_FUZZING_ENGINE
+cp fuzzer/gumbo.dict $OUT/parse_fuzzer.dict
+cp fuzzer/gumbo_corpus.zip $OUT/parse_fuzzer_seed_corpus.zip
```
stevecheckoway, line 7: Looks like the last line is missing a newline.

stevecheckoway:
Other than the missing newline, LGTM.

gumbo-parser/fuzzer/build-ossfuzz.sh
fuzzy-boiii23a: fixed (i hope) thanks for taking a look @stevecheckoway !

stevecheckoway:
Heh, now it has an extra newline, but that's fine. (Some tools treat the [POSIX requirement](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html) that all lines in a text file end in a newline seriously was the only reason I mentioned it.)

fuzzy-boiii23a:
> Heh, now it has an extra newline, but that's fine. (Some tools treat the [POSIX requirement](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap03.html) that all lines in a text file end in a newline seriously was the only reason I mentioned it.)

I used shfmt (https://github.com/mvdan/sh) and i also checked and there is a new line in the file locally unsure why git is not picking this up, sorry

flavorjones:
Thank you for your work on the fuzzer! LGTM.

PR #3109, opened 2024-01-22
test: handle upstream libxml2 error message changes

flavorjones:

Upstream libxml2 (pre-2.13) is changing some error messages. Let's make the tests more robust around those messages.


flavorjones:
upstream workflow run is: https://github.com/sparklemotion/nokogiri/actions/runs/7618098281

PR #2100, opened 2020-10-16
Compile native extensions with '-O2 -g' flags

ilyazub:
Resolves #2022.

In my case parsing of a big HTML file decreased from 1.1 seconds to 210 ms.

@flavorjones Regarding the final challenge, I haven't cleaned up `process_recipe` and `configure_options`, but I can try if you add some guidance or vision of the result.

And thanks for the detailed issue description. It was pretty straightforward to make this change.

PR #3111, opened 2024-01-26
feat: first-class musl support

flavorjones:

Ship `linux-musl` gems built specifically for musl libc systems.

Although there are no existing issues that this addresses, we've been bitten in the past by musl incompatibilities and this should prevent those sorts of problems going forward.

Also: upgrade to rake-compiler-dock 1.5.0.rc1

Test coverage has been updated.

No functional changes.


flavorjones:
Found a pretty significant blocker this weekend. We probably cannot release this unless https://github.com/rubygems/rubygems/issues/7432 is addressed in bundler or rubygems.

And even then, we probably need additional documentation to help people resolve issues.

Putting this on hold until we know more.

flavorjones:
Picking this back up again since the fix to the rubygems issue mentioned above was released in [bundler 2.5.6](https://github.com/rubygems/rubygems/blob/master/bundler/CHANGELOG.md#256-february-6-2024).

flavorjones:
I'm proceeding carefully here since an exploration shows that the version of rubygems and bundler that ships with Ruby 3.0 by default don't work well with the `-linux-musl` platform name, see https://github.com/rake-compiler/rake-compiler-dock/issues/117

My instinct is to wait until we drop support for Ruby 3.0 in Nokogiri before we ship first-class musl support, since we don't know of any issues with the current linux gem.

flavorjones:
Additional note that I had forgotten about: There are some things we're doing to work around precompiled-musl-support-from-a-gnu-toolchain.

Specifically I'm thinking of:

1. We are compiling with `-U_FORTIFY_SOURCE` to avoid issues on musl (see #2100 and 0b49210c). We should remove this workaround and compile with FORTIFY_SOURCE.
2. I think the thread-local storage feature being turned off in 004e6a6e is also because of musl, see #3031

flavorjones:
Tagging for v1.18 which is when we're planning to drop support for Ruby 3.0.

flavorjones:
Superseded by #3375

PR #3117, opened 2024-01-31
Implement `#clone` correctly with respect to singleton classes

### LINKED ISSUE #316: clone & singleton classes
If a specific Nokogiri::HTML::Document extends a module M, a clone of that document won't.

I expected otherwise, since in Ruby all builtin classes, as well as user created classes will copy the singleton_class when cloning.

Thakns!

require 'nokogiri'
require 'open-uri'

module M; def foo; end end

doc = Nokogiri::HTML(open('http://www.google.com/search?q=tenderlove'))
doc.extend M
doc.clone.respond_to? :foo # => false, should be true

flavorjones:

Fixes #316

Classes this PR updates:

- [x] `XML::Node`
- [x] `XML::Document`
- [x] `XML::NodeSet`

Yes.

The fix applies to both implementations.

### DIFF
--- a/lib/nokogiri/xml/document.rb
+++ b/lib/nokogiri/xml/document.rb
@@ -19,6 +19,10 @@ module Nokogiri
       NCNAME_CHAR       = NCNAME_START_CHAR + "\\-\\.0-9"
       NCNAME_RE         = /^xmlns(?::([#{NCNAME_START_CHAR}][#{NCNAME_CHAR}]*))?$/
 
+      OBJECT_DUP_METHOD = Object.instance_method(:dup)
+      OBJECT_CLONE_METHOD = Object.instance_method(:clone)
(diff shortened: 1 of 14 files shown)


lib/nokogiri/xml/document.rb
```diff
@@ -180,6 +184,38 @@ def initialize(*args) # :nodoc: # rubocop:disable Lint/MissingSuper
...
+      #   dup(level) → Nokogiri::XML::Document
+      #
+      # Duplicate this node.
+      #
+      # [Parameters]
+      # - +level+ (optional Integer). 0 is a shallow copy, 1 (the default) is a deep copy.
+      # [Returns] The new Nokogiri::XML::Document
+      #
+      def dup(level = 1)
+        copy = OBJECT_DUP_METHOD.bind_call(self)
```
flavorjones, line 199: I don't particularly like the `bind_call` hack here and in and `#clone`, but I can't come up with anything better.

PR #3121, opened 2024-02-04
Treat reaching the maximum parse depth as EOF

stevecheckoway:
The parser is fragile and hitting the maximum tree depth has exposed several bugs. By treating this condition as if it is an EOF means that we're back in the normal parsing condition.

This does expose a weakness in Gumbo's design: A spec bug that results in skipped steps or failure to reprocess a token (or if we fail to implement the spec in such a way), then the result can be crashes. Not ideal.

Fixes: oss-fuzz-66107

oss-fuzz-66107

Nope! I'm sorry to punt on testing again but I didn't take the time to understand why this bug occurred so I didn't produce a minimal test case. Treating hitting the max depth as EOF just seemed like the best way to fix this whole class of bugs by turning the exceptional (> max depth) parsing case into the normal parsing case.

C

### DIFF
--- a/gumbo-parser/src/parser.c
+++ b/gumbo-parser/src/parser.c
@@ -4762,7 +4762,18 @@ GumboOutput* gumbo_parse_with_options (
         adjusted_current_node &&
           adjusted_current_node->v.element.tag_namespace != GUMBO_NAMESPACE_HTML
       );
-      gumbo_lex(&parser, &token);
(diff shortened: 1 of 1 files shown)


flavorjones:
@stevecheckoway Thank you! No sweat on the testing; I feel like things the fuzzer finds aren't always great to add to a functional test suite, anyway.

I manually tested this branch against the fuzzer repro and it looks great! :heart:

PR #3129, opened 2024-02-09
backport nokogiri 1.11.0 to allow Ruby 2.4.5

mikeatlas:

### DIFF
--- a/.cross_rubies
+++ b/.cross_rubies
@@ -22,3 +22,8 @@
 2.5.0:x86_64-linux-gnu
 2.5.0:x86_64-darwin
 2.5.0:arm64-darwin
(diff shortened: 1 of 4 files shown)


flavorjones:
Sorry, I will not support older versions of Ruby. The effort to precompile for those platforms is too great.

mikeatlas:
@flavorjones sorry about this PR - no expectation on my part that you'd support 2.4.5 any longer. I had intended to open the PR on my own fork but the GitHub UI default option sends it to the original repo. 

And yes, I am trying to backport for that CVE, although it's one of dozens across many gems in a legacy application that need vuln updates. Thanks for the pointers to the mitigation suggestions.

PR #3134, opened 2024-02-21
dep: update minimum java version from 7 to 8

flavorjones:

- Modern OpenJDK (e.g., 21) won't build for Java 7 making development hard
- JRuby 9.4 only supports Java 8 and later, and Nokogiri only supports JRuby 9.4+
- Java 7 extended support finally ended in July 2022
  - https://support.oracle.com/knowledge/Middleware/2809708_1.html

See related commit 08e55603 which updated the minimum to 7 from 6 in 2020 (v1.11.0).

cc @headius for review

Non-functional change.

Non-functional change.

### DIFF
--- a/rakelib/extensions.rake
+++ b/rakelib/extensions.rake
@@ -376,8 +376,8 @@ if java?
 
     ext.ext_dir = "ext/java"
     ext.lib_dir = "lib/nokogiri"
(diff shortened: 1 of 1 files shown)


headius:
Totally ok!

PR #3135, opened 2024-02-21
dev: improve some libgumbo debug messages

flavorjones:

This is a commit from a debugging branch I've had locally for some time. I wanted to better understand the state transitions and found these changes to be helpful for my understanding. In case they help someone in the future, I thought I'd open a PR.

Non-functional change.

Non-functional change.


stevecheckoway (approved):
This looks good to me.

Issue #3133, opened 2024-02-20
Gem contains hardcoded path to `/home/flavorjones`


paulmenzel:
Using OpenProject with *Nokogiri*  1.14.2, we noticed, the automounter trying to mount `/home/flavorjones`:

    2024-02-20T11:58:08+01:00 platsch  automount[96539]: attempting to mount entry /home/flavorjones
    2024-02-20T11:58:08+01:00 platsch  automount[96539]: key "flavorjones" not found in map source(s).
    2024-02-20T11:58:08+01:00 platsch  automount[96539]: failed to mount /home/flavorjones

This is „caused“ by `bin/rake jobs:workoff`.

    $ fd nokogiri
    ruby-3.2.1/lib/ruby/gems/3.2.0/cache/nokogiri-1.14.2-x86_64-linux.gem
    ruby-3.2.1/bin/nokogiri
    ruby-3.2.1/lib/ruby/gems/3.2.0/specifications/nokogiri-1.14.2-x86_64-linux.gemspec
    ruby-3.2.1/lib/ruby/gems/3.2.0/gems/ruby-saml-1.15.0/gemfiles/nokogiri-1.5.gemfile
    ruby-3.2.1/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2-x86_64-linux/
    ruby-3.2.1/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2-x86_64-linux/lib/xsd/xmlparser/nokogiri.rb
    ruby-3.2.1/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2-x86_64-linux/lib/nokogiri/
    ruby-3.2.1/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2-x86_64-linux/lib/nokogiri/jruby/nokogiri_jars.rb
    ruby-3.2.1/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2-x86_64-linux/lib/nokogiri/3.2/nokogiri.so
    ruby-3.2.1/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2-x86_64-linux/lib/nokogiri/3.1/nokogiri.so
    ruby-3.2.1/lib/ruby/gems/3.2.0/gems/nokogiri-1.14.2-x86_64-linux/lib/nokogiri/3.0/nokogiri.so
[...]

PR #3137, opened 2024-02-21
ext(wip): don't set LIBPATH to avoid -Wl,-rpath linker args

flavorjones:
draft PR to explore solutions to #3133 


flavorjones:
Iteration 1 was to skip setting `$LIBPATH` in the extconf, which results in failures on platforms that need to find libz, as well as when `--disable-static` is used.

https://github.com/sparklemotion/nokogiri/actions/runs/7995462149?pr=3137

flavorjones:
Iteration 2 was to set `$LIBPATH` to an empty array right before generating the Makefile, so that intermediate libraries (libxslt, libxml2) could build against libz, but then the final `nokogiri.so` would link statically. However, this fails because it uses _shared_ libraries for `libz` and does not statically link as we desire.

https://github.com/sparklemotion/nokogiri/actions/runs/7995800240/job/21837134836

This issue is closely related to the set of challenges described in https://github.com/flavorjones/mini_portile/issues/118 and which remain unfixed. I need to do some more thinking about how mini_portile and/or rake-compiler-dock can and should support better behavior to support statically-linking these libraries; ideally when statically-linking a DLL the RPATH should be empty.

There's one more thing I will try, which is to run `patchelf` within the rake-compiler-dock container after linking Nokogiri's extensions, which addresses the symptom locally.

Issue #3160, opened 2024-03-18
[install/bug] Alternate toolchain specified in ENV works, but does not work when passed as build arguments


ziggythehamster:
**Have you read and followed the installation tutorial at http://www.nokogiri.org/tutorials/installing_nokogiri.html?**

- [x] Yes!

---

We are using a non-default toolchain, so e.g. `ld` is available at `gcc10-ld`. For most gems with native code, they either use RbConfig to locate the toolchain (`RbConfig::CONFIG['LD'] # => "gcc10-ld"`) or allow you to specify them as build arguments, e.g. `gem install foo -- LD=gcc10-ld` or `bundle config build.foo LD=gcc10-ld`. Nokogiri seems to pass along `CC` but won't pass along `LD` or `AR` to the libxml2 build.

Nokogiri _will_ pass these along if they are defined in ENV, however.

This results in this not working:

`gem install --platform ruby nokogiri -- AR=gcc10-ar CC=gcc10-cc CXX=gcc10-c++ LD=gcc10-ld`

But this works:

`AR=gcc10-ar CC=gcc10-cc CXX=gcc10-c++ LD=gcc10-ld gem install --platform ruby nokogiri`

The former would be preferred, because I can configure `BUNDLE_BUILD__NOKOGIRI` in the "user" bundler config file to correctly specify these. The latter requires someone who is `bundle install`'ing to specify the toolchain (which they should not need to know about).

I think that this would be a rather simple change to `extconf.rb`, but I'm not sure if this would be an accepted change or not.

flavorjones:
@ziggythehamster Thanks for opening this, I think this is just an oversight. I would absolutely review to a PR to pass these other values through!

flavorjones:
A fix for this was merged in #3165! Thanks, @ziggythehamster!

PR #3165, opened 2024-04-02
Make extconf.rb honor AR and LD variables in addition to CC

ziggythehamster:

#3160

Note I also have branches that fix this on v1.16.x and v1.15.x, but I wanted to make sure I actioned feedback in one place before I opened additional MRs.

Also note that Nokogiri doesn't seem to honor variables passed as gem configuration args, and I didn't add that here because I'm not entirely sure what the "canonical" way to do this is with mkmf/mini_portile. In other situations where I need to provide the toolchain (rdkafka/karafka-rdkafka), there seems to be no special code to convert `FOO=BAR` to an env var.

That said, it's not actually necessary in my case because using `AR` and `LD` from `RbConfig` is sufficient. So one might say if you need to do something different from that, maybe it's actually a good thing that you would need to do `AR=x gem install nokogiri` rather than `gem install nokogiri -- AR=x`. I can take a stab at making this work, though, if the team thinks that both ways should work.

N/A but I assume if it breaks the build in some configuration I didn't test, GitHub Actions will be mad about it :).

No

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -137,6 +137,9 @@ NOKOGIRI_HELP_MESSAGE = <<~HELP
       NOKOGIRI_USE_SYSTEM_LIBRARIES
           Equivalent to `--enable-system-libraries` when set, even if nil or blank.
 
+      AR
+          Use this path to invoke the library archiver instead of `RbConfig::CONFIG['AR']`
+
       CC
           Use this path to invoke the compiler instead of `RbConfig::CONFIG['CC']`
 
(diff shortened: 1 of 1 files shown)


flavorjones:
Thank for opening this. I'm on holiday for a few more days and will review as soon as I'm back.

One note, though: the 1.15 branch is unsupported and I don't want to cut any more releases on that series. Hope that's not an unpleasant surprise.

ziggythehamster:
Meh, I suppose I have no choice but to live with it :).

flavorjones:
I'll take care of the changelog entry once this merges. Thank you for contributing this!

flavorjones:
(And if you really need a patch release for v1.16.x I'd do it, just please open a PR on the `v1.16.x` branch to let me know.)

PR #3195, opened 2024-05-17
test: update tests to handle upstream libxml2

flavorjones:

upstream integration testing with libxml2, see gnome/libxml2@4fefba4c


flavorjones:
Upstream CI is https://github.com/sparklemotion/nokogiri/actions/runs/9130983439

PR #3201, opened 2024-05-21
Pass encoding as a positional argument

stevecheckoway:

The `encoding` keyword is ignored. An encoding argument should be passed
as a positional argument instead.

This is fixing the tests, but it does raise the question what, precisely, the
tests were testing before.

No behavioral changes.


flavorjones:
Superseded by #3204 which includes these changes.

PR #3217, opened 2024-06-05
tidy: remove code related to Ruby < 3.0

flavorjones:

There is still some code in Nokogiri related to support or testing on Ruby < 3.0. Since the gemspec was updated to `require_ruby_version >= 3.0` in cc788e34, this is safe to do now.


merlinwizzard (commented):
fix all vunrabilitys

Issue #2718, opened 2022-12-08
[build bug] Modifying gumbo source doesn't cause rebuilds


stevecheckoway:
**Please describe the bug**

Modifying the gumbo source files doesn't cause `rake compile:nokogiri` to rebuild the extension.

**Help us reproduce what you're seeing**

1. Compile the extension with `rake compile:nokogiri`
2. Touch a gumbo source file: `touch gumbo-parser/src/util.h`
3. Recompile the extension with `rake compile:nokogiri`

```
$ bundle exec rake compile:nokogiri
/usr/bin/make install target_prefix=
install -c -p -m 755 nokogiri.bundle /Users/steve/programming/nokogiri/lib/nokogiri
cp tmp/arm64-darwin21/nokogiri/3.1.2/nokogiri.bundle tmp/arm64-darwin21/stage/lib/nokogiri/nokogiri.bundle
$ touch gumbo-parser/src/util.h
$ bundle exec rake compile:nokogiri
cp gumbo-parser/src/util.h tmp/arm64-darwin21/stage/gumbo-parser/src/util.h
/usr/bin/make install target_prefix=
install -c -p -m 755 nokogiri.bundle /Users/steve/programming/nokogiri/lib/nokogiri
cp tmp/arm64-darwin21/nokogiri/3.1.2/nokogiri.bundle tmp/arm64-darwin21/stage/lib/nokogiri/nokogiri.bundle
```

Note that the extension was already compiled for the first invocation so nothing needed to be done. In the second compilation, it also thinks there's nothing to do. (The same holds true if you actually modify the file, not merely touch it, but touching is traditionally sufficient for build systems.)

**Expected behavior**

I expect the gumbo library to be recompiled (or at least the parts of it that were changed) and the `nokogiri.bundle` to be rebuilt.

**Environment**

- I'm building the `main` branch.
- ruby 3.1.2p20 (2022-04-12 revision 4491bb740a) [arm64-darwin21]
- macOS 13.0.1

flavorjones:
Thanks for opening this issue up, this has been mildly annoying me, too.

It's a side effect of how I ended up implementing the libgumbo build, as a separate library and not part of the extension -- using the [`packaged_tarball` strategy](https://github.com/flavorjones/ruby-c-extensions-explained/tree/main/packaged_source).

[...]

PR #3221, opened 2024-06-07
Autotools gumbo

flavorjones:

See conversation at #2718, this simplifies how we build libgumbo both during installation (extconf) and during development.

Primary author is @stevecheckoway

Existing test coverage should be fine.

No behavior changes.


flavorjones:
I'm a little concerned about introducing autoconf to the development build process; I made some additional changes to help run it better in CI, but I think we probably also need to add a test job that installs the gem on a system without autoconf just to make sure we're packaging everything correctly.

flavorjones:
@stevecheckoway I wonder if we should split gumbo out into a separate repository? While it might make integration testing a little bit harder,

1. it would make the build system more straightforward (parallel construction to libxml2 & libxslt)
2. it would make the library available for third-party usage (and I think we might have the best-maintained fork at this point)

Just an idle thought on a Saturday morning.

Moka-Rezq (commented):
@@

Moka-Rezq (commented):
@@

flavorjones:
@stevecheckoway Bumping this in case you have any thoughts on what I suggested? I'm not sure how to unblock this one right now.

Issue #1941, opened 2019-11-10
Cannot build on FreeBSD 12


theirix:
**Have you read and followed the installation tutorial at http://www.nokogiri.org/tutorials/installing_nokogiri.html?**

Yes.

**What is the complete output of `gem install`?**

```
% gem install nokogiri -v '=1.10.5'
Building native extensions. This could take a while...
ERROR:  Error installing nokogiri:
	ERROR: Failed to build gem native extension.

    current directory: /usr/local/lib/ruby/gems/2.5/gems/nokogiri-1.10.5/ext/nokogiri
/usr/local/bin/ruby25 -I /usr/local/lib/ruby/site_ruby/2.5 -r ./siteconf20191110-25375-1hxb03q.rb extconf.rb
checking if the C compiler accepts ... yes
Building nokogiri using packaged libraries.
Using mini_portile version 2.4.0
checking for gzdopen() in -lz... yes
checking for iconv... yes
************************************************************************
IMPORTANT NOTICE:

Building Nokogiri with a packaged version of libxml2-2.9.10
with the following patches applied:
	- 0001-Revert-Do-not-URI-escape-in-server-side-includes.patch
	- 0002-Remove-script-macro-support.patch
	- 0003-Update-entities-to-remove-handling-of-ssi.patch

Team Nokogiri will keep on doing their best to provide security
updates in a timely manner, but if this is a concern for you and want
to use the system library instead; abort this installation process and
reinstall nokogiri as follows:

    gem install nokogiri -- --use-system-libraries
        [--with-xml2-config=/path/to/xml2-config]
        [--with-xslt-config=/path/to/xslt-config]

If you are using Bundler, tell it to use the option:

    bundle config build.nokogiri --use-system-libraries
    bundle install

Note, however, that nokogiri is not fully compatible with arbitrary
versions of libxml2 provided by OS/package vendors.
************************************************************************
Extracting libxml2-2.9.10.tar.gz into tmp/amd64-portbld-freebsd12/ports/libxml2/2.9.10... OK
Running git apply with /usr/local/lib/ruby/gems/2.5/gems/nokogiri-1.10.5/patches/libxml2/0001-Revert-Do-not-URI-escape-in-server-side-includes.patch... OK
Running git apply with /usr/local/lib/ruby/gems/2.5/gems/nokogiri-1.10.5/patches/libxml2/0002-Remove-script-macro-support.patch... OK
[...]

PR #3230, opened 2024-06-12
dep: update libxml2 to 2.13.0 and libxslt to 1.1.40

flavorjones:

- https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.13.0
- https://gitlab.gnome.org/GNOME/libxslt/-/releases/v1.1.40


flavorjones:
Looks like we need to wait for a companion libxslt release with gnome/libxslt@75967fc0 to address the compilation issues.

flavorjones:
Added libxslt upgrade to 1.1.40 to this PR

flavorjones:
@rgrove Take note that `sanitize` is going to have some test failures once this makes it into a release (likely v1.17.0 later in the summer):

https://github.com/sparklemotion/nokogiri/actions/runs/9487728125/job/26145008195?pr=3230

```
  1) Failure:
Sanitize::Transformers::CleanDoctype::when :allow_doctype is true#test_0002_should not allow obviously invalid doctype declarations in documents [/__w/nokogiri/nokogiri/sanitize/test/test_clean_doctype.rb:47]:
--- expected
+++ actual
@@ -1 +1 @@
-"<!DOCTYPE html><html>foo</html>"
+"<!DOCTYPE blah><html>foo</html>"
```

rgrove:
> @rgrove Take note that `sanitize` is going to have some test failures once this makes it into a release (likely v1.17.0 later in the summer)

Thanks for the heads up! I'll look into this.

flavorjones:
@rgrove Hmm, I'm digging in and something weird is going on, it's not Sanitize.

flavorjones:
@rgrove Ahhhh ... gnome/libxml2@a3713f78

> Disallow xmlNodeSetName on DTD nodes. DTD nodes don't store the name in a dictionary. Calling xmlNodeSetName with a DTD node could result in an invalid free.

rgrove:
Fun! So it looks like Sanitize may need a small change after all. 🙂

flavorjones:
The [downstream Mechanize test failure](https://github.com/sparklemotion/nokogiri/actions/runs/9487728125/job/26145009739?pr=3230) is due to gnome/libxml2@8d0aaf4b95eaaaaadb4d793b8a3cb05b124ad1a2

Not totally sure what the relationship is between the change and the behavior, but I'm also not totally sure I care, we've seen edge case encoding changes like this before and the Mechanize test in question is more descriptive than prescriptive ... sigh. I'll probably just change the test.

flavorjones:
Ah, this is because Mechanize's `encoding_error?` method is checking the error message contents ... blurgh. Will update that and I think it should be fine.

flavorjones:
See https://github.com/sparklemotion/mechanize/pull/644

flavorjones:
I've opened https://github.com/rgrove/sanitize/pull/238 on sanitize as a potential workaround for the DTD name-change limitation.

flavorjones:
The failure showing up on freebsd is because my upstream fix from https://github.com/sparklemotion/nokogiri/issues/1941 shipped in v2.13.0! Removing the patch in `patches/libxml2/0003-libxml2.la-is-in-top_builddir.patch`! :rocket:

PR #3231, opened 2024-06-12
Support parsing noscript content in script-enabled mode

### LINKED ISSUE #3178: Feature Request: Option to Treat `noscript` Tags as Script Enabled

stevecheckoway:
This change adds support for parsing noscript content with the scripting flag enabled. In this case, Gumbo will treat the content of the noscript element as raw text, exactly as the browser would.

Closes #3178 

Yes.

It adds support for this parsing option to libgumbo only.

### DIFF
--- a/test/html5/test_api.rb
+++ b/test/html5/test_api.rb
@@ -172,7 +172,6 @@ class TestHtml5API < Nokogiri::TestCase
     img = doc.at("/html/body/noscript/img")
     refute_nil(img)
   end
-  
 
   ["pre", "listing", "textarea"].each do |tag|
     define_method("test_serialize_preserve_newline_#{tag}".to_sym) do
--- a/test/html5/test_api.rb
+++ b/test/html5/test_api.rb
@@ -117,7 +117,7 @@ class TestHtml5API < Nokogiri::TestCase
     # Start tag 'img' isn't allowed here
     # End tag 'noscript' isn't allowed here
     # End tag head isn't allowed here
-    assert(noscript.children.empty?)
+    assert_empty(noscript.children)
     img = doc.at("/html/body/img")
     refute_nil(img)
   end
@@ -130,7 +130,7 @@ class TestHtml5API < Nokogiri::TestCase
     noscript = doc.at("/html/head/noscript")
     assert_equal(0, doc.errors.length, doc.errors.join("\n"))
     assert_equal(1, noscript.children.length)
-    assert(noscript.children.first.is_a?(Nokogiri::XML::Text))
+    assert_kind_of(Nokogiri::XML::Text, noscript.children.first)
   end
 
   def test_parse_noscript_as_elements_in_body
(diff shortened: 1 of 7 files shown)


stevecheckoway:
Suggestions on a better option name are especially welcome.

I didn't want to use something like `scripting_enabled` since that strikes me as misleading.

stevecheckoway:
I just realized that I haven't had the html5lib tests running so this is really not ready for review yet. Gotta figure out what I did wrong here first.

stevecheckoway:
It seems that we're now passing the scripting enabled html5lib tree construction tests.

test/html5/test_api.rb
```diff
@@ -104,6 +104,66 @@ def test_serialization_encoding
...
+    refute_nil(img)
+  end
+
+  def test_parse_noscript_as_text_in_head
+    # In contrast to the previous test, when the scripting flag is enabled, the content
+    # of the noscript element is parsed as raw text.
+    html = "<!DOCTYPE html><head><noscript><img src=!></noscript></head>"
+    doc = Nokogiri::HTML5(html, parse_noscript_content_as_text: true, max_errors: 100)
+    noscript = doc.at("/html/head/noscript")
+    assert_equal(0, doc.errors.length, doc.errors.join("\n"))
```
flavorjones, line 131: Here and in the other tests, you may want to consider:

```suggestion
    assert_empty(doc.errors)
```

test/html5/test_api.rb
```diff
@@ -104,6 +104,66 @@ def test_serialization_encoding
...
+    img = doc.at("/html/body/noscript/img")
+    refute_nil(img)
+  end
+
+  def test_parse_noscript_as_text_in_body
+    html = "<!DOCTYPE html><body><noscript><img src=!></noscript></body>"
+    doc = Nokogiri::HTML5(html, parse_noscript_content_as_text: true, max_errors: 100)
+    noscript = doc.at("/html/body/noscript")
+    assert_equal(0, doc.errors.length, doc.errors.join("\n"))
+    assert_kind_of(Nokogiri::XML::Text, noscript.children.first)
```
flavorjones, line 149: Is it worth also asserting that there's only one child?

test/html5/test_tree_construction.rb
```diff
@@ -161,7 +166,7 @@ module Html5libTestCaseParser
   class BadHtml5libFormat < RuntimeError; end
 
   def self.parse_test(test_data)
-    test = { script: :both }
+    test = { script: :off }
```
flavorjones, line 169: The `html5lib-tests` docs say

> Then there \*may\* be a line that says "\#script-off" or
> "\#script-on". If a line that says "\#script-off" is present, the
> parser must set the scripting flag to disabled. If a line that says
> "\#script-on" is present, it must set it to enabled. Otherwise, the
> test should be run in both modes.

Should we be running all the tests that are not marked `on` or `off` in both modes?

flavorjones:
This PR should also update the docstring in `lib/nokogiri/html5.rb`.

Let me know if you'd like me to put in any of this work? I'm happy to do as much or as little as you like. ♥

flavorjones:
> Suggestions on a better option name are especially welcome

The name `parse_noscript_as_text` is not awful, and does what it says. If I was going to suggest anything, it would be a vaguer description like `emulate_browser` or `pretend_scripting_is_enabled`.

Or if we want to be funny, `wwtbd` ("What Would The Browser Do?")

test/html5/test_api.rb
```diff
@@ -104,6 +104,66 @@ def test_serialization_encoding
     assert_match("ฉันไม่พูดภาษาไทย", html2)
   end
 
+  def test_parse_noscript_as_elements_in_head
```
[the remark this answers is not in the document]
stevecheckoway: Done

test/html5/test_api.rb
stevecheckoway: Done.

test/html5/test_api.rb
stevecheckoway: Yes. Good call. Done.

test/html5/test_tree_construction.rb
stevecheckoway: I'd forgotten that the absence of `#script-off`/`#script-on` meant we were supposed to run it in both modes (which does explain the `:both`).

I think we should run the tests in both modes. How do you feel about putting that change off for a short while? I've got some refactoring/clean up to the tree construction tests that I was hoping land and it should be easy to make the change to that.

I've published the clean up branch (which was based on an earlier version of this PR) here https://github.com/stevecheckoway/nokogiri/compare/script-enabled-noscript...stevecheckoway:nokogiri:better-error-reporting The high-level idea is to have just a single class, `TestHtml5TreeConstruction`, and define all of the methods using `it` to give them descriptive names that include both the `.dat` file and the line number of the start of the test data.

The `tests.each_with_index` loop can then define all the relevant tests (script on vs. off, fragment with context node or not, etc.)

stevecheckoway:
> This PR should also update the docstring in `lib/nokogiri/html5.rb`.

I did that as well. But I left the existing option name. I'll defer to your experience in API design.

test/html5/test_tree_construction.rb
stevecheckoway: Or I guess I can add that commit to this PR. Whichever is easiest for you!

flavorjones: Yup, happy to deal with `:both` in a separate PR!

flavorjones (approved):
LGTM! I added a commit to fix up some of the rdoc formatting (which is famously challenging to write correctly without looking at the generated docs in a browser) and a CHANGELOG entry.

flavorjones:
The changelog entry isn't worded perfectly ... I'll fix it in postproduction

test/html5/test_tree_construction.rb
flavorjones: see https://github.com/sparklemotion/nokogiri/pull/3277

Issue #3183, opened 2024-04-24
feature: accept a context tag name to `Nokogiri::HTML5::DocumentFragment.parse`


flavorjones:
Currently, the XML and HTML4 document fragment parse methods accept a context tag name, but the HTML5 class does not.

This has made implementing code like https://github.com/flavorjones/nokogiri-html5-inference a little bit more challenging. It's a useful API and we should just support it.

This would be a good first issue for someone!

sharvy:
Hi @flavorjones, if no one has started working on this issue, I would love to work on it.

flavorjones:
@sharvy Ah, thank you so much for offering!

I apologize, but I already implemented this feature in #3246 (see eca68e7e), but I neglected to close this issue because I referenced the wrong issue id in that PR.

This feature will be in v1.17 when it ships, hopefully at the end of July.

Are you interested in this feature specifically? Or can I point you at some other issue(s) that might be a good "first issue"?

sharvy:
@flavorjones thanks for the update! Happy to hear that it's already done. And yes, please show/assign me some other GF issues! I appreciate the guidance 🙏🙂

Issue #3200, opened 2024-05-21
Deprecate positional hash arguments in the API


flavorjones:
Currently we've still got a few methods that use the pre-Ruby-3.0 convention of using a positional hash argument to pass options.

For example:

```ruby
# lib/nokogiri/html5/document_fragment.rb
def initialize(doc, tags = nil, ctx = nil, options = {})
  ...
end

...

def self.parse(tags, encoding = nil, options = {})
  ...
end
```

I'd like to start planning to deprecate these positional hashes in favor of true keyword arguments.

Note that we can't simply move to keyword arguments, since calls that use explicit hashes would stop working:

```ruby
Nokogiri::HTML5::DocumentFragment.parse(frag, nil, {max_errors: 1})
# ArgumentError: wrong number of arguments (given 3, expected 1..2)
#     lib/nokogiri/html5/document_fragment.rb:59:in `parse'
```

So we'll need to do an intermediate step that looks something like:

```patch
diff --git a/lib/nokogiri/html5/document_fragment.rb b/lib/nokogiri/html5/document_fragment.rb
index 1cd4b28d..6d3f1fe1 100644
--- a/lib/nokogiri/html5/document_fragment.rb
+++ b/lib/nokogiri/html5/document_fragment.rb
@@ -56,7 +56,12 @@ def serialize(options = {}, &block) # :nodoc:
       end
 
       # Parse a document fragment from +tags+, returning a Nodeset.
-      def self.parse(tags, encoding = nil, **options)
+      def self.parse(tags, encoding = nil, positional_options_hash = nil, **options)
+        unless positional_options_hash.nil?
+          warn("ClassName.methodname: Passing options as an explicit hash is deprecated. Use keyword arguments instead. This will become an error in a future release.", uplevel: 1, category: :deprecated)
+          options.merge!(positional_options_hash)
+        end
+
         doc = HTML5::Document.new
         tags = HTML5.read_and_encode(tags, encoding)
         doc.encoding = "UTF-8"
```

flavorjones:
Closing this, related work at #3323 mostly addressed this in a backwards-compatible way.

PR #3246, opened 2024-06-21
Fix fragment quirks mode, and make it easier to specify fragment context

flavorjones:

Coming from the discussion at #3203, I wanted to improve the fragment parsing API

- as discussed in #2646, parse in no-quirks mode if a context element name is provided (not a context `Node`, just the name)
- allow passing `:context` kwarg to `DocumentFragment.new` and `.parse`
- deprecate the positional options hash to `.parse` per notes at #3200

Included additional coverage for the API changes

HTML5 is only in CRuby.


stevecheckoway (approved):
LGTM

flavorjones:
This closed #3183

Issue #3244, opened 2024-06-21
downstream distro: No rule to make target 'nokogiri.so'


gouravkhunger:
Hi!

Context: I am trying to create installable [ruby bootstraps](http://github.com/jekyllex/ruby-android) for [android architectures](https://developer.android.com/ndk/guides/abis). I have [altered](https://github.com/jekyllex/ruby-android/blob/main/patches/ruby/bundled_gems.patch) the default bundled gems so that I can customize what gems are pre-installed. I wish to include nokogiri as well. To do so I use the the [termux-packages](https://github.com/termux/termux-packages) build system to build the dependencies `zlib`, `libiconv`, `libxml2`, `libxslt`.

The build system places the built files at `/data/data/<app-package>/files/usr/`, which is auto included into the flags when I build [ruby](https://github.com/termux/termux-packages/tree/master/packages/ruby) from their patches. So enabling `--use-system-libraries` does pick up the dependencies during the build.

So far things look fine as the parts are being compiled correctly. But I get an error at the final step

```
...
compiling /home/builder/.termux-build/ruby/src/.bundle/gems/nokogiri-1.16.5/ext/nokogiri/xml_cdata.c
compiling /home/builder/.termux-build/ruby/src/.bundle/gems/nokogiri-1.16.5/ext/nokogiri/xml_comment.c
compiling /home/builder/.termux-build/ruby/src/.bundle/gems/nokogiri-1.16.5/ext/nokogiri/xml_document.c
/home/builder/.termux-build/ruby/src/.bundle/gems/nokogiri-1.16.5/ext/nokogiri/xml_document.c:92:7: warning: '__xmlDeregisterNodeDefaultValue' is deprecated [-Wdeprecated-declarations]
  if (xmlDeregisterNodeDefaultValue) {
      ^
...
/data/data/sh.gourav.jekyllex/files/usr/include/libxml2/libxml/tree.h:682:10: note: expanded from macro 'XML_GLOBALS_TREE'
         XML_DEPRECATED)
         ^
/data/data/sh.gourav.jekyllex/files/usr/include/libxml2/libxml/xmlexports.h:69:43: note: expanded from macro 'XML_DEPRECATED'
    #define XML_DEPRECATED __attribute__((deprecated))
                                          ^
1 warning generated.
compiling /home/builder/.termux-build/ruby/src/.bundle/gems/nokogiri-1.16.5/ext/nokogiri/xml_document_fragment.c
[...]

PR #3249, opened 2024-06-23
fix: No rule to make target 'nokogiri.so' on android

gouravkhunger:

On android builds of ruby, `DLLIB` does not direct to a valid `nokogiri.so` path but `TARGET_SO` does. This PR fixes it for the `clean-ports` step.

Reference #3244

Not sure what kind of tests should be added. I have tested this on my [build system](https://github.com/jekyllex/ruby-android/actions/runs/9634684158) and it passes.

Does not change the behaviour, but allows the compilation.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -205,6 +205,10 @@ def openbsd?
   RbConfig::CONFIG["target_os"].include?("openbsd")
 end
 
+def android?
+  RbConfig::CONFIG["target_os"].include?("android")
+end
+
 def aix?
   RbConfig::CONFIG["target_os"].include?("aix")
 end
@@ -586,6 +590,8 @@ def do_clean
     Pathname.glob(pwd.join("tmp", "*", "ports")) do |dir|
       FileUtils.rm_rf(dir, verbose: true)
     end
+
+    exit!(0) if android?
 
     if config_static?
       # ports installation can be safely removed if statically linked.
@@ -1165,7 +1171,7 @@ if config_clean?
     mk.print(<<~EOF)
 
       all: clean-ports
-      clean-ports: $(DLLIB)
+      clean-ports: #{android? ? "$(TARGET_SO)" : "$(DLLIB)"}
       \t-$(Q)$(RUBY) $(srcdir)/extconf.rb --clean --#{static_p ? "enable" : "disable"}-static
     EOF
   end


ext/nokogiri/extconf.rb
```diff
@@ -587,6 +591,8 @@ def do_clean
       FileUtils.rm_rf(dir, verbose: true)
     end
 
+    exit!(0) if android?
+
```
gouravkhunger, line 595: No matter what I tried, I kept getting:

```
/home/builder/.termux-build/ruby/src/lib/fileutils.rb:1624:in `stat': No such file or directory @ rb_file_s_stat - /home/builder/.termux-build/ruby/src/.bundle/gems/nokogiri-1.16.5/ports/archives/libxml2-2.12.7.tar.xz (Errno::ENOENT)
  from /home/builder/.termux-build/ruby/src/lib/fileutils.rb:1624:in `block in install'
  from /home/builder/.termux-build/ruby/src/lib/fileutils.rb:2469:in `block in fu_each_src_dest'
  from /home/builder/.termux-build/ruby/src/lib/fileutils.rb:2485:in `fu_each_src_dest0'
  from /home/builder/.termux-build/ruby/src/lib/fileutils.rb:2467:in `fu_each_src_dest'
  from /home/builder/.termux-build/ruby/src/lib/fileutils.rb:1623:in `install'
  from /home/builder/.termux-build/ruby/src/tool/rbinstall.rb:190:in `install'
  from /home/builder/.termux-build/ruby/src/tool/rbinstall.rb:881:in `block in extract_files'
  from /home/builder/.termux-build/ruby/src/tool/rbinstall.rb:876:in `each'
  from /home/builder/.termux-build/ruby/src/tool/rbinstall.rb:876:in `extract_files'
  from /home/builder/.termux-build/ruby/src/lib/rubygems/installer.rb:861:in `extract_files'
  from /home/builder/.termux-build/ruby/src/lib/rubygems/installer.rb:328:in `install'
  ...
```

For example, this snippet with multiple safe execution checks (`begin`...`rescue`, `File.directory?`) didn't fix it too:

```rb
    Pathname.glob(root.join("ports", config_static? ? "" : "archives")) do |dir|
      begin
        FileUtils.rm_rf(dir, verbose: true) if File.directory?(dir)
      rescue Errno::ENOENT => e
        puts "Error: #{e.message}"
      end
    end
```

I guess the file/directory exists but the process doesn't have the required permissions to delete those files.

For now I've exited from the program at this checkpoint, but I suppose there could be a better fix.

flavorjones:
I've created https://github.com/sparklemotion/nokogiri/pull/3250 which is a simpler proposed solution.

ext/nokogiri/extconf.rb
flavorjones: This is not the right solution and breaks the general case. Let's move the conversation back to #3244 or to #3250 if there are problems with that solution.

PR #3252, opened 2024-06-24
test: raise exception on assertionless tests

flavorjones:

Raise exception if a test makes no assertions.

See https://railsatscale.com/2024-01-25-catching-assertionless-tests/

And this PR cleans up all of the assertionless tests that were found.

- some were not useful (doing a thing in a loop might have been useful
  early in nokogiri development, but not anymore) and were deleted
- some needed explicit `refute_raises`
- some needed explicit `refute_valgrind_errors`
- some just needed rewriting with assertions

### DIFF
--- a/test/helper.rb
+++ b/test/helper.rb
@@ -206,11 +206,15 @@ module Nokogiri
         end
       end
 
+      super
+
+      if !skipped? && !error? && assertions.zero?
+        raise(Minitest::Assertion, "Test is missing assertions")
+      end
+
       if Nokogiri.uses_libxml?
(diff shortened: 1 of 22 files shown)


flavorjones:
cc @nvasilevski :heart:

nvasilevski:
Exciting, thanks! ❤️ 

Sorry about the tests that had to become more verbose but I genuinely believe that setting explicit assertions helps with readability especially for new contributors

flavorjones:
@nvasilevski I recently looked at https://github.com/seattlerb/minitest-proveit/blob/master/lib/minitest/proveit.rb which might be what I move to at some point

PR #3261, opened 2024-06-24
fix: Schema#validate returns errors on empty and malformed files

### LINKED ISSUE #642: XML::Schema#validate(filename) returns 0 errors for malformed documents
It seems that `Nokogiri::XML::Schema#validate` for filenames (not `Nokogiri::XML::Document`s) does not return any errors for malformed XML documents: only for well-formed documents that do not conform to the provided schema.

``` ruby
require 'nokogiri'

Nokogiri::VERSION
# => "1.5.3.rc3"

RUBY_VERSION
# => "1.9.3"

bad_xml_data = 'xyz'
xsd_data = <<-XSD
<?xml version="1.0" encoding="UTF-8"?>
<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema"
  targetNamespace="http://www.example.org/contactExample"
  xmlns:tns="http://www.example.org/contactExample"
  elementFormDefault="qualified">
  <xs:element name="Contacts"></xs:element>
</xs:schema>
XSD

malformed_xml_path = '/tmp/malformed.xml'

File.open(malformed_xml_path, 'w') { |f| f.write(bad_xml_data) }
# => 3

schema = Nokogiri::XML::Schema.new(xsd_data)
# => #<Nokogiri::XML::Schema:0x007feea9c40c80>

schema.validate(malformed_xml_path)
# => []

schema.valid?(malformed_xml_path)
# => true
```

flavorjones:

`Schema#validate` now returns errors on empty and malformed files. Previously, it would return errors on empty/malformed Documents, but not when reading from files.

Closes #642

Yes!

This fixes different bugs on CRuby and JRuby, but both impls agree with each other now.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -36,6 +36,7 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 
 * `Node#clone`, `NodeSet#clone`, and `*::Document#clone` all properly copy the metaclass of the original as expected. Previously, `#clone` had been aliased to `#dup` for these classes (since v1.3.0 in 2009). [#316, #3117] @flavorjones
 * CSS queries for pseudo-selectors that cannot be translated into XPath expressions now raise a more descriptive `Nokogiri::CSS::SyntaxError` when they are parsed. Previously, an invalid XPath expression was evaluated and a hard-to-understand XPath error was raised by the query engine. [#3193] @flavorjones
+* `Schema#validate` returns errors on empty and malformed files. Previously, it would return errors on empty/malformed Documents, but not when reading from files. [#642] @flavorjones
 * [CRuby] libgumbo (the HTML5 parser) treats reaching max-depth as EOF. This addresses a class of issues when the parser is interrupted in this way. [#3121] @stevecheckoway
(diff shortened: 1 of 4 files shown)


jlahtinen:
It seems like previously in jruby

```
document = Nokogiri::XML("")
fail "document has xml syntax errors" if document.errors.any?
fail "document has schema errors" unless Nokogiri::XML::Schema(File.open("schema.xsd")).validate(document)
```

did not fail but now fails to schema errors.

So I think it would be maybe a good idea to rephrase  changelog a little: "but not when reading from files" part.

flavorjones:
@jlahtinen I acknowledge that this changelog entry could have been more complete with respect to the JRuby implementation, but JRuby makes up a vanishingly small percentage of Nokogiri users, and the entirety of code contributors, so I won't apologize for simplifying the changelog entry to improve readability for 99.5% of users who are on CRuby. 🤷

I only seem to hear from JRuby users when Nokogiri fails to meet expectations. I would love to hear from JRuby users who want to help [improve JRuby support](https://github.com/sparklemotion/nokogiri/issues?q=is%3Aissue%20state%3Aopen%20label%3Aplatform%2Fjruby).

jlahtinen:
@flavorjones thanks for the response. Makes sense.

PR #3271, opened 2024-07-03
tests: handle upstream libxml2 returning null for UTF-8 enc handler

flavorjones:

Example failure: https://github.com/sparklemotion/nokogiri/actions/runs/9774049613/job/26981486984

Let's just work around it in the tests. The behavior we're testing isn't specific to UTF-8, so let's just swap in ISO-2022-JP.

gnome/libxml2@6d8427dc

Tagging @nwellnhof just for awareness.

Yes

N/A


nwellnhof:
What behavior has changed exactly in libxml2? The issue with `xmlFindCharEncodingHandler` should have been fixed here: https://gitlab.gnome.org/GNOME/libxml2/-/commit/c4d8343b7c56a5fd00fcd7edb8e2ceb786081570

PR #3275, opened 2024-07-03
ci: build a container for libxml2 to use in its pipeline

flavorjones:

Build a container for libxml2 to use in its pipeline

See:
- https://gitlab.gnome.org/GNOME/libxml2/-/issues/758
- https://gitlab.gnome.org/GNOME/libxml2/-/merge_requests/271


flavorjones:
Pipeline run is here: https://github.com/sparklemotion/nokogiri/actions/runs/9782778129

PR #3291, opened 2024-07-15
ci: update suppression to match new stack trace in Ruby 3.4-dev

flavorjones:

I think the stack trace changed with ruby/ruby@51bd8165, see for example https://github.com/sparklemotion/nokogiri/actions/runs/9935752042/job/27442553528

I'm not sure why this is the only leak showing up now, but am deferring further study until I devote some time to cleaning up known leaks where we raise exceptions in Ruby callbacks (from libxml2), see #2096 and #1610

### DIFF
--- a/suppressions/ruby.supp
+++ b/suppressions/ruby.supp
@@ -98,26 +98,25 @@
   fun:evaluate
 }
 {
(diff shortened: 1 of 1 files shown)


flavorjones:
Upstream workflow is https://github.com/sparklemotion/nokogiri/actions/runs/9941574034

PR #3316, opened 2024-10-15
Make available the XPath path of the node where a SyntaxError occurred during Schema validation

ryanong:

When validating an XML document with a schema it is difficult to impossible to know where the error happened in the document. With the path on the error, we can identify where in the document the error occurred.

I believe so. I updated the test in the schema to check to see if path errors were being returned correctly.

This change the C implementation gently. Setting the path during error creation.
I looked if xerces-j could do it, but I couldn't find anything on the SAXException that would help.

### DIFF
--- a/lib/nokogiri/xml/syntax_error.rb
+++ b/lib/nokogiri/xml/syntax_error.rb
@@ -19,23 +19,18 @@ module Nokogiri
         end
       end
 
-      # What part of libxml2 raised this error (enum xmlErrorDomain)
       attr_reader :domain
-      # libxml2 error code (enum xmlParserErrors)
       attr_reader :code
-      # libxml2 error level (enum xmlErrorLevel)
       attr_reader :level
       attr_reader :file
       attr_reader :line
-      # libxml2 path of the node in the tree that caused the error
+      # ⚠ path functionality is not available when running JRuby.
+      #
+      # path of the node that caused the error when validating a `Nokogiri::XML::Document`
       attr_reader :path
-      # libxml2 extra string information
       attr_reader :str1
-      # libxml2 extra string information
       attr_reader :str2
-      # libxml2 extra string information
       attr_reader :str3
-      # libxml2 extra extra number information
       attr_reader :int1
       attr_reader :column
 
--- a/lib/nokogiri/xml/syntax_error.rb
+++ b/lib/nokogiri/xml/syntax_error.rb
@@ -24,9 +24,14 @@ module Nokogiri
       attr_reader :level
       attr_reader :file
       attr_reader :line
-      # ⚠ path functionality is not available when running JRuby.
+
+      # The XPath path of the node that caused the error when validating a `Nokogiri::XML::Document`.
       #
-      # path of the node that caused the error when validating a `Nokogiri::XML::Document`
+      # This attribute will only be non-nil when the error is emitted by `Schema#validate` on
+      # Document objects. It will return `nil` for DOM parsing errors and for errors emitted during
+      # Schema validation of files.
+      #
+      # ⚠ `#path` is not supported on JRuby, where it will always return `nil`.
       attr_reader :path
       attr_reader :str1
       attr_reader :str2
--- a/lib/nokogiri/xml/syntax_error.rb
+++ b/lib/nokogiri/xml/syntax_error.rb
@@ -19,14 +19,23 @@ module Nokogiri
         end
       end
 
+      # What part of libxml2 raised this error (enum xmlErrorDomain)
       attr_reader :domain
+      # libxml2 error code (enum xmlParserErrors)
       attr_reader :code
+      # libxml2 error level (enum xmlErrorLevel)
       attr_reader :level
       attr_reader :file
       attr_reader :line
+      # libxml2 path of the node in the tree that caused the error
+      attr_reader :path
+      # libxml2 extra string information
       attr_reader :str1
+      # libxml2 extra string information
       attr_reader :str2
+      # libxml2 extra string information
       attr_reader :str3
+      # libxml2 extra extra number information
       attr_reader :int1
       attr_reader :column
 
--- a/test/html4/test_document.rb
+++ b/test/html4/test_document.rb
@@ -24,7 +24,6 @@ module Nokogiri
           refute_empty(errors, "has errors")
           errors.each do |error|
             assert_equal(error.to_s.chomp, error.to_s)
-            assert_nil(error.path)
           end
         end
 
--- a/test/html4/test_document.rb
+++ b/test/html4/test_document.rb
@@ -24,6 +24,7 @@ module Nokogiri
           refute_empty(errors, "has errors")
           errors.each do |error|
             assert_equal(error.to_s.chomp, error.to_s)
+            assert_nil(error.path)
           end
         end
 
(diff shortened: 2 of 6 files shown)


flavorjones:
@ryanong Thanks. Looks like some rubocop failures and appveyor is showing test failures.

test/xml/test_schema.rb
```diff
@@ -171,6 +171,10 @@ class TestNokogiriXMLSchema < Nokogiri::TestCase
 
         assert(errors = xsd.validate(tempfile.path))
         assert_equal(2, errors.length)
+        assert_equal(
```
flavorjones, line 174: In this particular test case, `error->node` looks like it's `NULL` (on my system, anyway).

You may want to write a new test case that's crafted so that `error->node` is present (and honestly I'm not sure when it's present or not, I'd have to dig into the libxml2 code). Let me know if you need assistance.

lib/nokogiri/xml/syntax_error.rb
```diff
@@ -19,14 +19,23 @@ def aggregate(errors)
...
       attr_reader :line
+      # libxml2 path of the node in the tree that caused the error
+      attr_reader :path
+      # libxml2 extra string information
       attr_reader :str1
+      # libxml2 extra string information
       attr_reader :str2
+      # libxml2 extra string information
       attr_reader :str3
+      # libxml2 extra extra number information
```
flavorjones, line 38: Thank you for documenting these! ♥ ♥ ♥

test/xml/test_schema.rb
flavorjones: Interesting. I'm running two ways: `bundle exec rake test` (whole suite) and `bundle exec rake test A=-n/validate_invalid_file/` (individual test).

flavorjones: Are you running with the packaged libxml2 or are you using your system library?

ryanong: never mind... needs different test

flavorjones: I'm betting you're running with system libxml2 old enough that it doesn't have this commit in it yet: https://github.com/GNOME/libxml2/commit/e20c9c148c725e2933efa143ee6a543a5cae4204

(looks like 2.9.10 or earlier)

ryanong: I just put the test in the wrong place, it needed to go in the one above. it was failing locally too

ryanong: It appears errors only render when validating a document and not a file. I am guessing when validating a document a node exists where as a file there is no node.

ryanong:
It looks like path only returns when validating an in memory document and not a file. Should I document this in the path description?

flavorjones (changes requested):
I'd like some DOM parser tests for `SyntaxError#path`, too. Specifically I'd like separate tests for XML and HTML4 to make sure those parsers' behavior is captured.

Don't worry about JRuby, I can spike on support and if I can't make it work I'll make the decision about whether to just document it as "libxml only".

flavorjones:
> It looks like path only returns when validating an in memory document and not a file. Should I document this in the path description?

That's strange, if you can commit a test for each (one passing, one failing) I'd like to check if that's intended libxml2 behavior.

flavorjones:
> I'd like some DOM parser tests for SyntaxError#path, too. Specifically I'd like separate tests for XML and HTML4 to make sure those parsers' behavior is captured.

Can you please also exercise in-memory and from-file parsing for these two parsers as well, just to capture the current behavior. Thanks.

ryanong:
It appears that html doesn't raise `Nokogiri::XML::SyntaxError` It uses `Nokogiri::SyntaxError`

I'll add some tests for doc parsing for xml

ryanong:
I think node is returning null because `validate_file` uses `xmlSchemaValidateFile` which uses sax parsing and doesn't have a node

ryanong:
I'm attempting to validate an XML file with an XSD and then associate the errors with the inputs that create the file.
```ruby
users = [
  { id: "1", name: "bob" },
  { id: "2", name: "boboboboboboboboboobobobbob"},
]
doc = build_doc(users)
# Nokogiri::XML
# <users>
#   <user id="1">
#     <name>bob</name>
#   </user>
#   <user id="2">
#     <name>boboboboboboboboboobobobbob</name>
#   </user>
# </users>

raw_errors = validate(doc)
errors = []
raw_errors.each do |error|
  id = doc.at_xpath(error.path).xpath("ancestor::user")["id"]
  user = users.find { |u| u[:id] == id }
  errors << { error: error, parent: user }
end
```

That way I can return each error to associated user.

Is there a way to know what node there is at a given line a column? I imagine I could do that by walking the entire tree and indexing it and matching against that but this feels cleaner

lib/nokogiri/xml/syntax_error.rb
flavorjones, line 38: These comments, though probably an improvement over nothing, still don't feel very informative. Over time I'm trying to evolve Nokogiri towards being less-libxml2-specific and I'm not sure which of these attributes would make that cut.

Let's keep the `:path` one since we're adding it, but let's not tie it to libxml2 in the description. And I think the rest of these comments can be omitted in the name of making it easier for me to `:nodoc:` them later if I decide to.

test/html4/test_document.rb
```diff
@@ -24,6 +24,7 @@ def test_exceptions_remove_newlines
           refute_empty(errors, "has errors")
           errors.each do |error|
             assert_equal(error.to_s.chomp, error.to_s)
+            assert_nil(error.path)
```
flavorjones, line 27: Let's remove this assertion. This is not the right test for it (exercising newlines in error messages) and we have adequate coverage in the other tests.

test/xml/test_schema.rb
```diff
@@ -171,6 +175,10 @@ class TestNokogiriXMLSchema < Nokogiri::TestCase
...
+          ["/purchaseOrder/billTo/state", "/purchaseOrder/shipTo/state"],
+          errors.map(&:path).sort,
+        )
```
flavorjones, line 181: Let's change this back to asserting that the paths are nil, since the file parsing case won't have a DOM node to refer to, and please document that for the `#path` attribute.

flavorjones (changes requested):
Ah, I see. Sorry, it's been a long week for me and I'm just now understanding fully what you're trying to do and why it's not working for input files. All of this makes sense.

I'm fine with documenting that `SyntaxError#path` is populated only when there is an actual DOM object to reference ("best efforts") and that it doesn't work in JRuby.

In other places I've used a string like this to document differences between xerces/neko and libxml2/gumbo:

```
⚠ HTML5 functionality is not available when running JRuby.
```

so something like that is probably fine for this.

lib/nokogiri/xml/syntax_error.rb
ryanong: This makes sense, thanks for all your time mike.

As for which ones that make the cut, I would probably say code and domain don't nor all the str#

I imagine the ideal API instead of returning code + str# + int# it would be a new error object with the values per code. This would be relevant to all libraries. The mapping would be annoying but it would be a great user experience.

Will make changes now

flavorjones (approved):

flavorjones:
@ryanong Thank you again! Will be in the v1.17 release (hopefully soon).

flavorjones:
Changelog updated in 1e06cd16

PR #3322, opened 2024-11-12
dep: update vendored libxml2 to v2.13.5 (main)

flavorjones:

https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.13.5


flavorjones:
We'll have to wait for the gnome mirrors to get populated. I'll kick CI off again later.

PR #3331, opened 2024-11-14
Update XML::Node.parse and XML::Node#canonicalize for keyword args

infews:
KEYWORD ARGS ALL THE THINGS!

YAS, sez @flavorjones 


flavorjones:

flavorjones:
I'm going to wait to merge this until I figure out Document#canonicalize (which requires more invasive changes to the C code), so this won't land in v1.17.0 but I'll target v1.18.0.

PR #3335, opened 2024-11-14
Updates HTML5::DocumentFragment.new, #parse with keyword arguments;

infews:
Keywooooooorrrrrrrrd Arguments!

This PR is not complete. DocumentFragment.new is in C, and needs to be updated to call HTML5::DocumentFragment#initialize without the positional options hash.


flavorjones:

flavorjones:
@infews I'm merging these commits into https://github.com/sparklemotion/nokogiri/pull/3355 so I can have one giant PR for all the Document and DocumentFragment updates!

Issue #3323, opened 2024-11-14
Introduce support for keyword arguments


flavorjones:
The Nokogiri API was largely designed and implemented for Ruby 1.9, which pre-dates keyword argument support and parameter forwarding with `...`. As a result, many methods are using positional arguments where keyword arguments are probably better.

For example, `Nokogiri::XML::Document.parse` is declared as:

```ruby
def parse(string_or_io, url = nil, encoding = nil, options = ParseOptions::DEFAULT_XML)
```

I would like to update these method signatures to support keyword arguments without dropping support (yet) for positional arguments. In the case of `Document.parse`, this might look like:

```ruby
def parse(string_or_io,
          _url = nil, _encoding = nil, _options = XML::ParseOptions::DEFAULT_XML,
          url: _url, encoding: _encoding, options: _options)
```

I would also like to update the docstrings to describe the keyword arguments, and remove documentation for the positional arguments.

Another example is `Nokogiri::XML()` which is defined as:

```ruby
def XML(thing, url = nil, encoding = nil, options = XML::ParseOptions::DEFAULT_XML, &block)
  Nokogiri::XML::Document.parse(thing, url, encoding, options, &block)
end
```

Methods that are forwarding positional arguments should be updated to more cleanly forward either keyword arguments or positional arguments, and avoid redeclaring parameter default values:

```ruby
def XML(...)
  Nokogiri::XML::Document.parse(...)
end
```

flavorjones:
For an example of the kind of documentation I'd like to see, take a look at https://github.com/sparklemotion/nokogiri/pull/3324 which updates `Nokogiri::XML::Document.parse`.

flavorjones:
The list of methods that need to be updated is at https://docs.google.com/spreadsheets/d/1pWM7N5D9073J8IRVO0yghIySTYhsJoO73sslAVxEO-A/edit?gid=0#gid=0

We'll work on this for RubyConf 2024 hack day!

flavorjones:
TODO: link back to this issue in doc strings for people to discover some documentation about the positional arguments

PR #3336, opened 2024-11-14
Update HTML4::DocumentFragment#initialize & #parse to take kwargs

MattJones:
https://github.com/sparklemotion/nokogiri/issues/3323

Yes and No. Test coverage is included for `#parse`. But the the `DocumentFragment.new` class method calls a C method, which is not getting touched here. Thus I can't add in tests using `.new`.
@flavorjones : you'll need to update the C code to accept keyword args. Then, you can update [this call](https://github.com/sparklemotion/nokogiri/blob/main/lib/nokogiri/html4/document_fragment.rb#L85). Also you'll want to add some tests. I added two in `test_document_fragment` before I realized they wouldn't work. One for "without a context node" and one for "with a context node".

I.... don't think so?

### DIFF
--- a/lib/nokogiri/html4/document_fragment.rb
+++ b/lib/nokogiri/html4/document_fragment.rb
@@ -87,7 +87,7 @@ module Nokogiri
 
       # It's recommended to use either DocumentFragment.parse or XML::Node#parse rather than call this
       # method directly.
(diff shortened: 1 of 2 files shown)


flavorjones:

flavorjones:
@MattJones I'm merging these commits into #3335 so I can have one giant PR for all the Document and DocumentFragment updates!

PR #3338, opened 2024-11-15
extconf: respect AR and RANLIB in recipes on darwin

joshheinrichs-shopify:

When using Nix on Darwin, we ideally should be latching onto the Nix-provided AR and RANLIB. Without this, when blocking Xcode via `sandbox-exec`, I see errors like this while attempting to install the gem, since we end up using the default `ar` and `runlib`
```
xcode-select: note: No developer tools were found, requesting install.
If developer tools are located at a non-default location on disk, use `sudo xcode-select --switch path/to/Xcode.app` to specify the Xcode that you wish to use for command line developer tools, and cancel the installation dialog.
See `man xcode-select` for more details.
```

No but this seems relatively harmless?

No

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -209,8 +209,12 @@ def aix?
   RbConfig::CONFIG["target_os"].include?("aix")
 end
 
-def nix?
+def unix?
   !(windows? || solaris? || darwin?)
+end
+
+def nix?
+  ENV.key?("NIX_CC")
 end
 
 def truffle?
@@ -705,7 +709,7 @@ end
 
 # Add SDK-specific include path for macOS and brew versions before v2.2.12 (2020-04-08) [#1851, #1801]
 macos_mojave_sdk_include_path = "/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/libxml2"
-if config_system_libraries? && darwin? && Dir.exist?(macos_mojave_sdk_include_path)
+if config_system_libraries? && darwin? && Dir.exist?(macos_mojave_sdk_include_path) && !nix?
   append_cppflags("-I#{macos_mojave_sdk_include_path}")
 end
 
@@ -828,7 +832,7 @@ else
       end
     end
 
-    unless nix?
+    unless unix?
       libiconv_recipe = process_recipe(
         "libiconv",
         dependencies["libiconv"]["version"],
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -928,7 +928,8 @@ else
     end
 
     if darwin? && !cross_build_p
-      recipe.configure_options += ["RANLIB=/usr/bin/ranlib", "AR=/usr/bin/ar"]
+      recipe.configure_options << "RANLIB=/usr/bin/ranlib" unless ENV.key?("RANLIB")
+      recipe.configure_options << "AR=/usr/bin/ar" unless ENV.key?("AR")
     end
 
     if windows?
@@ -969,7 +970,8 @@ else
     cflags = concat_flags(ENV["CFLAGS"], "-O2", "-U_FORTIFY_SOURCE", "-g")
 
     if darwin? && !cross_build_p
-      recipe.configure_options += ["RANLIB=/usr/bin/ranlib", "AR=/usr/bin/ar"]
+      recipe.configure_options << "RANLIB=/usr/bin/ranlib" unless ENV.key?("RANLIB")
+      recipe.configure_options << "AR=/usr/bin/ar" unless ENV.key?("AR")
     end
 
     if windows?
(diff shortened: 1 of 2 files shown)


ext/nokogiri/extconf.rb
```diff
@@ -928,7 +928,10 @@ def configure
     end
 
     if darwin? && !cross_build_p
-      recipe.configure_options += ["RANLIB=/usr/bin/ranlib", "AR=/usr/bin/ar"]
+      recipe.configure_options += [
+        "RANLIB=#{ENV.fetch("RANLIB", "/usr/bin/ranlib")}",
```
flavorjones, line 932: For libxml2's `configure` script, if an environment variable like `RANLIB` is set, then it doesn't need to be passed explicitly to the configure command. Can you restructure this to something like the following

```ruby
recipe.configure_options << "RANLIB=/usr/bin/ranlib" unless ENV.key?("RANLIB")
recipe.configure_options << "AR=/usr/bin/ar" unless ENV.key?("AR")
```

ext/nokogiri/extconf.rb
```diff
@@ -969,7 +972,10 @@ def configure
     cflags = concat_flags(ENV["CFLAGS"], "-O2", "-U_FORTIFY_SOURCE", "-g")
 
     if darwin? && !cross_build_p
-      recipe.configure_options += ["RANLIB=/usr/bin/ranlib", "AR=/usr/bin/ar"]
+      recipe.configure_options += [
+        "RANLIB=#{ENV.fetch("RANLIB", "/usr/bin/ranlib")}",
```
flavorjones, line 976: Same here for libxslt's `configure` script.

flavorjones (changes requested):
Hi @joshheinrichs-shopify, thanks for opening this.

Do you know if there are any github actions that we could use to test Nix-on-Darwin? (For example, we're using `vmactions/freebsd-vm@v1` to test FreeBSD.)

I put in a lot of work over the past few years to thoroughly exercise compilation on supported platforms, and that coverage has allowed me to confidently make changes to `extconf.rb`. I'd prefer not to merge this without adding some test coverage to prevent regressing, if possible.

flavorjones:
I'm also curious: can you share why you're not using precompiled gems?

flavorjones (approved):
Looks good! I kicked off CI.

ext/nokogiri/extconf.rb
```diff
@@ -928,7 +928,8 @@ def configure
     end
 
     if darwin? && !cross_build_p
-      recipe.configure_options += ["RANLIB=/usr/bin/ranlib", "AR=/usr/bin/ar"]
+      recipe.configuration_options << "RANLIB=/usr/bin/ranlib" unless ENV.key?("RANLIB")
```
flavorjones, line 931:
```suggestion
      recipe.configure_options << "RANLIB=/usr/bin/ranlib" unless ENV.key?("RANLIB")
```

ext/nokogiri/extconf.rb
```diff
@@ -969,7 +970,8 @@ def configure
     cflags = concat_flags(ENV["CFLAGS"], "-O2", "-U_FORTIFY_SOURCE", "-g")
 
     if darwin? && !cross_build_p
-      recipe.configure_options += ["RANLIB=/usr/bin/ranlib", "AR=/usr/bin/ar"]
+      recipe.configuration_options << "RANLIB=/usr/bin/ranlib" unless ENV.key?("RANLIB")
```
flavorjones, line 973:
```suggestion
      recipe.configure_options << "RANLIB=/usr/bin/ranlib" unless ENV.key?("RANLIB")
```

ext/nokogiri/extconf.rb
joshheinrichs-shopify: 🤦 thanks

flavorjones:
Kicked off CI again.

flavorjones:
@joshheinrichs-shopify CI is failing. Looks like a tricky one. My guess is that there's inconsistent header usage: `xmlSwitchEncodingName` was added in libxml2 v2.13.0, and so the extconf is checking for its existence using a "modern" libxml2 (maybe Nokogiri's own vendored headers?), but then compiling against something older than that (maybe nix's system headers?). I have no idea what nix package is being linked- and compiled-against so I can't verify that hunch, but something like that is likely the cause.

joshheinrichs-shopify:
hmm looks like that failing test is latching onto xcode for some reason. taking a look...

ext/nokogiri/extconf.rb
```diff
@@ -209,10 +209,14 @@ def aix?
   RbConfig::CONFIG["target_os"].include?("aix")
 end
 
-def nix?
+def unix?
```
flavorjones, line 212: LOL, naming things is hard

flavorjones:
Just kicked CI and it looks like the darwin-nix-enabled build is green! If everything goes green I'll merge.

flavorjones:
Thank you! This will be in v1.17.0 when it ships (hopefully very very soon).

PR #3352, opened 2024-12-06
ci: skip downstream sanitize test for now

flavorjones:

Downstream sanitize tests are failing right now after https://github.com/sparklemotion/nokogiri/pull/3348 was released in v1.16.8


flavorjones:
Closed in favor of #3353

PR #3360, opened 2024-12-09
[DOC] Parse options doc

BurdetteLamar:

Improvements to doc for XML::ParseOptions.

N/A.  All doc.

No.

### DIFF
--- a/lib/nokogiri/xml/parse_options.rb
+++ b/lib/nokogiri/xml/parse_options.rb
@@ -1,167 +1,399 @@
 # coding: utf-8
 # frozen_string_literal: true
 
+# :markup: markdown
+
 module Nokogiri
   module XML
-    # Options that control the parsing behavior for XML::Document, XML::DocumentFragment,
-    # HTML4::Document, HTML4::DocumentFragment, XSLT::Stylesheet, and XML::Schema.
     #
-    # These options directly expose libxml2's parse options, which are all boolean in the sense that
-    # an option is "on" or "off".
+    # \Class to house options for parsing \XML or \HTML4 (but not \HTML5).
     #
-    # 💡 Note that HTML5 parsing has a separate, orthogonal set of options due to the nature of the
-    # HTML5 specification. See Nokogiri::HTML5.
+    # ## About the Examples
     #
-    # ⚠ Not all parse options are supported on JRuby. Nokogiri will attempt to invoke the equivalent
-    # behavior in Xerces/NekoHTML on JRuby when it's possible.
+    # Examples on this page assume that the following code has been executed:
     #
-    # == Setting and unsetting parse options
+    # ```
+    # require 'nokogiri' # Make Nokogiri available.
+    # include Nokogiri   # Allow omitting leading 'Nokogiri::'.
+    # xml_s = "<root />\n"
+    # File.write('t.xml', xml_s)
+    # html_s = "<html />\n"
+    # File.write('t.html', html_s)
+    # ```
     #
-    # You can build your own combinations of parse options by using any of the following methods:
+    # Examples executed via `IRB` (interactive Ruby) display \ParseOptions instances
+    # using method #inspect.
     #
-    # [ParseOptions method chaining]
+    # ## Methods That Use \XML::ParseOptions
     #
-    #   Every option has an equivalent method in lowercase. You can chain these methods together to
-    #   set various combinations.
+    # Each of the methods listed below performs parsing for an \XML or \HTML4 source:
     #
-    #     # Set the HUGE & PEDANTIC options
-    #     po = Nokogiri::XML::ParseOptions.new.huge.pedantic
-    #     doc = Nokogiri::XML::Document.parse(xml, nil, nil, po)
+    # - Each requires a leading argument that specifies the source of the text to be parsed;
+    #   except as noted, the argument's value may be either:
     #
-    #   Every option has an equivalent <code>no{option}</code> method in lowercase. You can call these
-    #   methods on an instance of ParseOptions to unset the option.
+    #     - A string.
+    #     - An open IO stream (must respond to methods `read` and `close`).
     #
-    #     # Set the HUGE & PEDANTIC options
-    #     po = Nokogiri::XML::ParseOptions.new.huge.pedantic
+    #     Examples:
     #
-    #     # later we want to modify the options
-    #     po.nohuge # Unset the HUGE option
-    #     po.nopedantic # Unset the PEDANTIC option
+    #     ```
+    #     XML::parse(xml_s)
+    #     HTML4.parse(html_s)
+    #     XML::parse(File.open('t.xml'))
+    #     HTML4.parse(File.open('t.html'))
+    #     ```
     #
-    #   💡 Note that some options begin with "no" leading to the logical but perhaps unintuitive
-    #   double negative:
+    # - Each accepts a trailing optional argument `options`
+    #   (or keyword argument `options`)
+    #   that specifies parsing options;
+    #   the argument's value may be either:
     #
-    #     po.nocdata # Set the NOCDATA parse option
-    #     po.nonocdata # Unset the NOCDATA parse option
+    #     - An integer: see [Bitmap Constants](rdoc-ref:ParseOptions@Bitmap+Constants).
+    #     - An instance of \ParseOptions: see ParseOptions.new.
     #
-    #   💡 Note that negation is not available for STRICT, which is itself a negation of all other
-    #   features.
+    #     Examples:
     #
+    #     ```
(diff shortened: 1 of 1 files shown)


flavorjones:
Thank you for opening this! I apologize for not having made time to review this yet. I will do so in the next few days!

lib/nokogiri/xml/parse_options.rb
```diff
@@ -1,167 +1,406 @@
...
-    #     po.nohuge # Unset the HUGE option
-    #     po.nopedantic # Unset the PEDANTIC option
+    # ```
+    # ParseOptions.new(ParseOptions::DTDLOAD)
+    # # => #<Nokogiri::XML::ParseOptions: ... strict, dtdload>
+    # ParseOptions.new(ParseOptions::DTDATTR)
+    # # => #<Nokogiri::XML::ParseOptions: ... strict, dtdattr>
+    # ```
     #
-    #   💡 Note that some options begin with "no" leading to the logical but perhaps unintuitive
```
flavorjones, line 41: I think we lost this note in the PR's changes. I think it's useful/helpful to people to explicitly show them how to negate one of the negatively-named options.

lib/nokogiri/xml/parse_options.rb
```diff
@@ -1,167 +1,406 @@
...
+    # ```
     #
     class ParseOptions
-      # Strict parsing
+
+      # Strict parsing; do not recover from errors in input.
       STRICT      = 0
 
-      # Recover from errors. On by default for XML::Document, XML::DocumentFragment,
-      # HTML4::Document, HTML4::DocumentFragment, XSLT::Stylesheet, and XML::Schema.
```
flavorjones, line 72: I think we need to keep the note on each constant that states whether the bit is ON or OFF by default (and which macros have which value if it varies).

lib/nokogiri/xml/parse_options.rb
```diff
@@ -1,167 +1,406 @@
...
     #
-    #     po.nocdata # Set the NOCDATA parse option
-    #     po.nonocdata # Unset the NOCDATA parse option
+    # ```
+    # options = ParseOptions::BIG_LINES | ParseOptions::COMPACT | ParseOptions::NOCDATA
+    # ParseOptions.new(options)
+    # # => #<Nokogiri::XML::ParseOptions: ... strict, nocdata, compact, big_lines>
+    # ```
     #
-    #   💡 Note that negation is not available for STRICT, which is itself a negation of all other
```
flavorjones, line 47: I think we need to keep this note somewhere, doesn't need to be a separate call-out though.

lib/nokogiri/xml/parse_options.rb
```diff
@@ -186,21 +425,88 @@ def #{constant.downcase}?
...
+      # options.strict? # => true
+      # ```
       def strict?
         @options & RECOVER == STRICT
       end
 
+      # :call-seq:
+      #    self == object
+      #
+      # Returns whether `object` is equal to `self` (`object.to_i == to_i`):
```
flavorjones, line 464: I think this is a confusing statement, we can't both test for identity (`self`) and the value of the integer.

I would describe this as "Returns true if the same options are set in `self` and `other`."

flavorjones (commented):
Sorry it took so long to circle back on this PR. Thank you so much for the time you've invested!

lib/nokogiri/xml/parse_options.rb
BurdetteLamar: We have this:

    #   Note that there is no unsetter `nostrict`,
    #   but the setter `recover` serves the same purpose:
    #

lib/nokogiri/xml/parse_options.rb
```diff
@@ -1,167 +1,406 @@
...
+# :markup: markdown
+
 module Nokogiri
   module XML
-    # Options that control the parsing behavior for XML::Document, XML::DocumentFragment,
-    # HTML4::Document, HTML4::DocumentFragment, XSLT::Stylesheet, and XML::Schema.
     #
-    # These options directly expose libxml2's parse options, which are all boolean in the sense that
-    # an option is "on" or "off".
+    # \Class to house options for parsing \XML or \HTML4 (but not \HTML5).
```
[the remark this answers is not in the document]
BurdetteLamar: Changed to "contain."

lib/nokogiri/xml/parse_options.rb
```diff
@@ -1,167 +1,406 @@
...
+    #     XML::parse(xml_s, options: XML::ParseOptions::STRICT)
+    #     HTML4::parse(html_s, options: XML::ParseOptions::BIG_LINES)
+    #     XML::parse(xml_s, options: XML::ParseOptions.new.strict)
+    #     HTML4::parse(html_s, options: XML::ParseOptions.new.big_lines)
+    #     ```
+    #
+    # - Each (except as noted) accepts a block that allows parsing options to be specified;
+    #   see [Options-Setting Blocks](rdoc-ref:ParseOptions@Options-Setting+Blocks).
     #
-    # 💡 Note that HTML5 parsing has a separate, orthogonal set of options due to the nature of the
```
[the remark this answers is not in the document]
BurdetteLamar: Modified.

lib/nokogiri/xml/parse_options.rb
```diff
@@ -1,167 +1,406 @@
...
-    # HTML5 specification. See Nokogiri::HTML5.
+    # The methods:
     #
-    # ⚠ Not all parse options are supported on JRuby. Nokogiri will attempt to invoke the equivalent
```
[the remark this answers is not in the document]
BurdetteLamar: Modified.

lib/nokogiri/xml/parse_options.rb
```diff
@@ -1,167 +1,406 @@
...
-    # behavior in Xerces/NekoHTML on JRuby when it's possible.
+    # - Nokogiri.HTML.
```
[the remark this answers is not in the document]
BurdetteLamar: Removed.

lib/nokogiri/xml/parse_options.rb
```diff
@@ -1,167 +1,406 @@
...
+    # - Nokogiri.HTML4.
```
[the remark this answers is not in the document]
BurdetteLamar: Removed.

lib/nokogiri/xml/parse_options.rb
```diff
@@ -1,167 +1,406 @@
...
+    # OLD10,
+    # PEDANTIC,
+    # RECOVER,
+    # SAX1,
+    # STRICT,
+    # XINCLUDE.
     #
-    #   Most parsing methods will accept a block for configuration of parse options, and we
-    #   recommend chaining the setter methods:
+    # There are also several "shorthand" constants that can set multiple options:
```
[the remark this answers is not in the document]
BurdetteLamar: Modified.

lib/nokogiri/xml/parse_options.rb
BurdetteLamar: Restored.

lib/nokogiri/xml/parse_options.rb
BurdetteLamar: Restored.

flavorjones (approved):
This looks great! I rebased onto origin/main and made a few small whitespace changes to please rubocop and make sure it's indented consistently.

flavorjones:
Thank you so much, @BurdetteLamar!

nwellnhof:
@flavorjones Some additional notes from me:

- DTDATTR and DTDVALID imply DTDLOAD and are unsafe as well. 
- SAX1 should probably not be exposed.
- NODICT should probably not be exposed.
- XINCLUDE, NOXINCNODE and NOBASEFIX are only used by the XML Reader and XInclude API.
- HUGE is safe these days.

flavorjones:
@nwellnhof Thanks for the additional notes, either @BurdetteLamar or I will circle back on these docs and make some more changes.

Do you know offhand when HUGE became safe to enable? I can enable it by default if Nokogiri is built with an appropriate version of libxml2.

nwellnhof:
I also finalized the design of new options for 2.14, so I simply leave this here:

- UNZIP: Enable decompression. This option has no real effect for now. The plan is that users who really need decompression start to add the option. At a later point, it will be required to enable decompression.
- NO_SYS_CATALOG: Don't use system catalogs when resolving DTDs or entities.
- CATALOG_PI: Enable `oasis-xml-catalog` PIs. This is a really obscure feature that should have never been enabled by default. I don't think your users need it.

flavorjones:
I've opened #3439 describing followup actions from Nick's notes above.

PR #3366, opened 2024-12-10
ci: test rake-compiler-dock edge in the upstream workflow

flavorjones:

In CI, test against rake-compiler-dock edge/snapshot so we can make releases more confidently.


flavorjones:
This is a good start, but putting on pause until dropping 3.0 and handling the rcd 1.6.0 upgrade.

flavorjones:
Will revisit in a new PR.

PR #3380, opened 2024-12-15
Support compiled XPath expressions

flavorjones:

There has been some discussion, summarized at #3266, about exposing libxml2's support for compiled XPath expressions. The idea is that, if you have a complex expression that you use a lot and you don't want to pay the cost of parsing/compiling it multiple times, then you can compile it once and presumably your document search will be faster.

This PR implements a new T_DATA class, `XML::XPath::Expression`, which stores the result of compiling an XPath expression via `xmlXPathCompile`. The `XPathContext` class knows how to accept either a String or an Expression.

However, I'm not seeing noticeable improvements in speed, though my benchmark may not capture the benefits.

I'm posting this as a draft in case someone wants to write me a benchmark that shows compiled XPath expressions are compellingly faster than just using Strings. Right now, based on what I'm seeing, I'm not at all sure the complexity is worth the benefit.

Yes.

This is an optimization available on CRuby only; though the idea is that the shorthand methods `Nokogiri::XML::XPath.expression` and `Nokogiri::CSS.selector` will be no-ops on JRuby (returning the string argument) and code that uses Expressions could be portable across both implementations.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,11 @@ This release ships separate precompiled GNU and Musl gems for all linux platform
 This release drops precompiled native platform gems for `x86-linux` and `x86-mingw32`. **These platforms are still supported.** Users on these platforms must install the "ruby platform" gem which requires a compiler toolchain. See [Installing the `ruby` platform gem](https://nokogiri.org/tutorials/installing_nokogiri.html#installing-the-ruby-platform-gem) in the installation docs. (#3369, #3081)
 
 
+### Improved
+
+* CSS and XPath queries are faster now that `Node#xpath`, `Node#css`, and related functions are re-using the underlying xpath context object (which is expensive to initialize). We benchmarked a 2.8x improvement for a 6kb file, and a more modest 1.3x improvement for a 70kb file. (#3378) @flavorjones
+
+
 ## v1.17.2 / 2024-12-12
 
 ### Fixed
--- a/ext/java/nokogiri/XmlXpathContext.java
+++ b/ext/java/nokogiri/XmlXpathContext.java
@@ -179,7 +179,7 @@ public class XmlXpathContext extends RubyObject
     final NokogiriXPathFunctionResolver fnResolver = NokogiriXPathFunctionResolver.create(handler);
     try {
       return tryGetNodeSet(context, expr, fnResolver);
-    } catch (TransformerException ex) {
+    } catch (TransformerException | RuntimeException ex) {
       throw XmlSyntaxError.createXMLXPathSyntaxError(context.runtime,
           (expr + ": " + ex.toString()),
           ex).toThrowable();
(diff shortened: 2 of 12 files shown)


flavorjones:
An example benchmark script:

```ruby
#! /usr/bin/env ruby

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "nokogiri", path: "."
  gem "benchmark-ips"
end

doc_large = Nokogiri::HTML5.parse(File.read(File.join(__dir__, "../test/files/tlm.html")))
doc_small = Nokogiri::HTML5.parse(File.read(File.join(__dir__, "../test/files/noencoding.html")))

Benchmark.ips do |x|
  x.warmup = 0
  expression_str = "//p[nokogiri-builtin:css-class(@class,'br0') and count(preceding-sibling::*)=0]"
  expression_comp = Nokogiri::XML::XPath::Expression.new(expression_str)

  x.report("small: compiled") do
    doc_small.xpath(expression_comp).length == 0 or raise("nope")
  end

  x.report("small: string") do
    doc_small.xpath(expression_str).length == 0 or raise("nope")
  end

  x.compare!
end
```

flavorjones:
Updated benchmark script, using one extremely small expression as advised by Nick in #3378:

```ruby
#! /usr/bin/env ruby

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "nokogiri", path: "."
  gem "benchmark-ips"
end

doc = Nokogiri::HTML5.parse(File.read(File.join(__dir__, "../test/files/tlm.html")))

expressions = [
  ["//@href", 203],
  ["//*[nokogiri-builtin:css-class(@class,'br0') and count(preceding-sibling::*)=0]", 8],
]

expressions.each do |expression_str, count|
  Benchmark.ips do |x|
    x.warmup = 0
    expression_comp = Nokogiri::XML::XPath::Expression.new(expression_str)

    x.report("string") do
      doc.xpath(expression_str).length == count or raise("nope #{doc.xpath(expression_comp).length}")
    end

    x.report("compiled") do
      doc.xpath(expression_comp).length == count or raise("nope #{doc.xpath(expression_comp).length}")
    end

    x.compare!
  end
end
```

Result:

```
              string     18.797k (± 5.6%) i/s   (53.20 μs/i) -     91.300k in   4.978472s
            compiled     18.887k (± 5.6%) i/s   (52.95 μs/i) -     91.721k in   4.978564s

Comparison:
            compiled:    18887.3 i/s
              string:    18797.1 i/s - same-ish: difference falls within error

              string      1.707k (± 3.1%) i/s  (585.67 μs/i) -      8.518k in   4.998124s
            compiled      1.713k (± 3.9%) i/s  (583.92 μs/i) -      8.536k in   4.997768s

Comparison:
            compiled:     1712.6 i/s
              string:     1707.5 i/s - same-ish: difference falls within error
```

So I'm still not seeing much/any benefit to compiled expressions. Will continue to leave this PR open to invite folks to write a benchmark that demonstrates that this is faster in a real-world benchmark (macro or micro).

PR #3386, opened 2024-12-20
Reset xmlXPathContext state variables before each evaluation

flavorjones:

In #3378 we implemented re-used of xmlXPathContext objects for performance.

If we're re-using the xmlXPathContext object, there's a chance that the context variables will be trashed by recursive custom functions.

In https://github.com/sparklemotion/nokogiri/pull/3378#issuecomment-2557001734, @nwellnhof advised:

> Note that if you use a single XPath context and support custom XPath
> extension functions, a custom function could evaluate XPath
> expressions recursively which will lead to corruption of context
> variables. This is mostly due to some design mistakes in libxml2.

So let's set these context variables back to their default.

Functional test coverage should be sufficient to show the features still work, but I did not add test coverage that would have revealed a bug.

CRuby only fix

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Nokogiri follows [Semantic Versioning](https://semver.org/), please see the [REA
 
 ---
 
-## 1.18.0.rc1 / 2024-12-16
+## v1.18.0.rc1 / 2024-12-16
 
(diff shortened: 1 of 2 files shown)


flavorjones:
After chatting more with Nick, this approach is not right, closing.

PR #3393, opened 2024-12-26
perf: html5 attribute parsing

### LINKED ISSUE #2568: fix O(n^2) behavior when checking for duplicate attributes (libgumbo)
**Please describe the bug**

Originally https://github.com/rubys/nokogumbo/issues/144

Code was added to limit the number of attributes supported per element to prevent DoS attacks: https://github.com/rubys/nokogumbo/pull/143

That safety limit is here: https://github.com/sparklemotion/nokogiri/blob/main/gumbo-parser/src/tokenizer.c#L792

It would be great to support more attributes by addressing performance concerns in the implementation.

flavorjones:

Addressing quadratic performance of HTML5 attribute parsing as described in #2568 and https://github.com/rubys/nokogumbo/issues/144.

Closes #2568

### Some benchmarks

<details><summary>Here is the benchmark script.</summary>

```ruby
#!/usr/bin/env ruby

require "bundler/inline"

gemfile do
  source "https://rubygems.org"
  gem "nokogiri", path: "."
  gem "benchmark-ips"
end

def html_with_attributes(count)
  html = "<div"
  count.times do |j|
    html << " fake-attr-#{j}"
  end
  html << ">"
end

html = {}
Benchmark.ips do |x|
  x.warmup = 0

  [ 1, 2, 4, 8, 16, 32, 64, 128, 256, 512, 1024, 2048, 4096, 8192, 16384 ].each do |attribute_count|
    html[attribute_count] = html_with_attributes(attribute_count)

    x.report "#{attribute_count.to_s.rjust(7)} attributes" do
      Nokogiri::HTML5(html[attribute_count], max_attributes: 100_000)
    end
  end
end
```

</details>

before:

```
        1 attributes    242.976k (± 9.3%) i/s    (4.12 μs/i) -    936.599k in   4.838811s
        2 attributes    219.456k (± 9.4%) i/s    (4.56 μs/i) -    837.644k in   4.858317s
        4 attributes    191.173k (± 9.3%) i/s    (5.23 μs/i) -    715.392k in   4.882336s
        8 attributes    144.813k (±12.2%) i/s    (6.91 μs/i) -    516.194k in   4.912577s
       16 attributes     83.396k (±17.2%) i/s   (11.99 μs/i) -    238.564k in   4.958331s
       32 attributes     55.800k (± 9.9%) i/s   (17.92 μs/i) -    190.871k in   4.970063s
       64 attributes     27.381k (±14.7%) i/s   (36.52 μs/i) -     91.867k in   4.981936s
      128 attributes     12.707k (± 9.4%) i/s   (78.70 μs/i) -     49.906k in   4.990254s
      256 attributes      4.723k (± 8.1%) i/s  (211.74 μs/i) -     19.657k in   4.995888s
      512 attributes      1.417k (±10.6%) i/s  (705.61 μs/i) -      6.339k in   4.998084s
     1024 attributes    382.078 (±11.5%) i/s    (2.62 ms/i) -      1.811k in   5.000356s
     2048 attributes    119.825 (±10.8%) i/s    (8.35 ms/i) -    582.000 in   5.003555s
     4096 attributes     29.034 (± 6.9%) i/s   (34.44 ms/i) -    144.000 in   5.003602s
     8192 attributes      6.492 (± 0.0%) i/s  (154.04 ms/i) -     33.000 in   5.111381s
    16384 attributes      1.478 (± 0.0%) i/s  (676.68 ms/i) -      8.000 in   5.417232s
```

after:

```
        1 attributes    229.084k (±10.4%) i/s    (4.37 μs/i) -    868.071k in   4.846385s
        2 attributes    213.788k (± 9.9%) i/s    (4.68 μs/i) -    804.683k in   4.859819s
        4 attributes    181.602k (±10.7%) i/s    (5.51 μs/i) -    652.740k in   4.886700s
        8 attributes    137.811k (± 9.9%) i/s    (7.26 μs/i) -    500.690k in   4.911176s
       16 attributes     95.462k (± 9.9%) i/s   (10.48 μs/i) -    338.045k in   4.938699s
       32 attributes     61.388k (± 9.5%) i/s   (16.29 μs/i) -    216.911k in   4.961587s
       64 attributes     34.368k (±10.2%) i/s   (29.10 μs/i) -    121.985k in   4.976219s
      128 attributes     17.315k (±10.9%) i/s   (57.75 μs/i) -     60.716k in   4.987542s
      256 attributes      9.011k (± 9.2%) i/s  (110.98 μs/i) -     32.443k in   4.992878s
      512 attributes      4.552k (±11.8%) i/s  (219.69 μs/i) -     16.522k in   4.995834s
     1024 attributes      2.255k (±13.6%) i/s  (443.38 μs/i) -      8.303k in   4.997861s
     2048 attributes      1.091k (±17.6%) i/s  (916.91 μs/i) -      4.132k in   4.998710s
     4096 attributes    544.840 (±22.0%) i/s    (1.84 ms/i) -      2.101k in   5.006943s
     8192 attributes    282.789 (±22.3%) i/s    (3.54 ms/i) -      1.089k in   5.000604s
    16384 attributes    134.991 (±27.4%) i/s    (7.41 ms/i) -    542.000 in   5.001360s
```

In graphical form:

image

Just the "after", showing linear performance now:

image

### Analysis

Analyzing the parsing of a tag with 16000 attributes (using `samply`), we see two hotspots:

image

1. the `strlen`/`memcmp` checks in libgumbo's `tokenizer.c:finish_attribute_name` which checks for duplicates in the attribute list: https://github.com/sparklemotion/nokogiri/blob/729c96c05f692d69e4d1d3be6d4b9524bdbe9322/gumbo-parser/src/tokenizer.c#L804-L820
2. libxml2's `xmlNewPropInternal` traversing the singly-linked properties list to append each new property (to maintain parse order)

This PR addresses (1) by introducing a hashmap and using it if there are 16 or more attributes.

This PR addresses (2) by inlining most of `xmlNewNsProp` but keeping a pointer to the last property to avoid re-traversing the linked list each time.

After those changes, the flamegraph looks much better, with no obvious hotspots:

image

### hashmap

Keeping in mind we may still someday want to publish our libgumbo fork as a separate library, I opted to bring in a new self-contained hashmap library from https://github.com/tidwall/hashmap.c

I'm using [xxHash3](https://github.com/Cyan4973/xxHash) which seems to be pretty fast, but I'm open to other opinions on what hash function to use.

## Have you included adequate test coverage?

Yes.

## Does this change affect the behavior of either the C or the Java implementations?

HTML5 is CRuby-only.


flavorjones:
@stevecheckoway I would really love your feedback on this when you get a few minutes. ❤️

ext/nokogiri/libxml2_polyfill.c
```diff
@@ -112,3 +112,14 @@ xmlSwitchEncodingName(xmlParserCtxtPtr ctxt, const char *encoding)
   return (xmlSwitchToEncoding(ctxt, handler));
 }
 #endif
+
+#ifndef HAVE_XMLADDIDSAFE
+int
+xmlAddIDSafe(xmlAttrPtr attr, const xmlChar *value) {
+  xmlIDPtr id = xmlAddID(NULL, attr->doc, value, attr);
```
stevecheckoway, line 119: This can also fail and return NULL due to allocation failure. I think that in this case that will cause `xmlAddIDSafe` to return that the ID already exists.

gumbo-parser/src/string_set.c
```diff
@@ -0,0 +1,46 @@
...
+
+static uint64_t
+string_hash2(const char* str, size_t len, uint64_t seed0, uint64_t seed1) {
+  return hashmap_xxhash3(str, len, seed0, seed1);
+}
+
+GumboStringSet *
+gumbo_string_set_new(size_t cap)
+{
+  return (GumboStringSet*)hashmap_new(sizeof(char *), cap, SEED0, SEED1, string_hash, string_compare, NULL, NULL);
```
stevecheckoway, line 28: The cast for the return value shouldn't be necessary.

stevecheckoway (changes requested):
I'm glad you looked into fixing this. I was hesitant to reach into the internals of libxml2. It's often  unclear which fields of the structs are not going to change.

flavorjones:
@stevecheckoway Thanks for the review. Yes, I was sloppy with malloc failures to get something up and running; I'll fix.

I was super curious about what you think of introducing the hashmap into libgumbo? Just want to make sure you don't think it's a silly thing to do.

stevecheckoway:
@flavorjones I think adding the hashmap is better than going with the red-black tree I implemented for this exact task before I discovered that libxml2 also had quadratic behavior.

ext/nokogiri/libxml2_polyfill.c
flavorjones: This one we can't do anything about. If we're using this polyfill then we're using libxml2 2.12 or earlier which doesn't have gnome/libxml2@e34a49b7 to return a separate error code if a memory allocation fails.

stevecheckoway: This may not be worth dealing with but if `xmlAddID()` returns `NULL`, then we could do something like
```c
xmlError *lastError = xmlGetLastError();

if (lastError && lastError->code == XML_ERR_NO_MEMORY && lastError->level = XML_ERR_FATAL) {
    // Handle the OOM condition
}
```
That's from looking at https://github.com/GNOME/libxml2/blob/068abc0b75995c3d1576c078f01c76cc0814b280/valid.c#L56

PR #3397, opened 2024-12-30
ci: re-enable downstream sanitize test

rgrove:

In #3348, downstream CI tests for Sanitize were temporarily disabled because the changes in that PR (intentionally) caused some of Sanitize's tests to fail.

As of Sanitize 7.0.0, tests are passing again with the latest Nokogiri and it's safe to re-enable the downstream tests. 🎉


flavorjones:
Thank you @rgrove!

PR #3427, opened 2025-02-05
[DOC] Doc for XML::Node#<<

BurdetteLamar:
Revises doc for XML::Node#<<.

Also adds heading "Subclasses" near the top of the class doc, to make it a link target.


BurdetteLamar:
Should I be adding "[no CI]" or some such to the base desciption?

BurdetteLamar:
Needs new examples.

PR #3431, opened 2025-02-13
[DOC] Revision for XML::Node.add_child

BurdetteLamar:


BurdetteLamar:
@flavorjones, much that follows will depend on this PR (b/c the example code will be cited).

lib/nokogiri/xml/node.rb
```diff
@@ -192,16 +198,47 @@ def decorate!
...
+      #   add_child(object) -> object
+      #
+      # Appends one or more children (as specified by the given +object+) to +self+;
+      # returns +object+.
+      #
+      # When +object+ is an XML::Node of any kind,
+      # adds the node as the last child of +self+;
+      # sets the parent of that node as +self+:
+      #
+      #   doc = XML::Document.new
```
flavorjones, line 211: Creating a bare Document this way, and calling Node methods on it, feels very foreign to me and I'm not sure it clearly illustrates that we're talking about Nodes and not Documents in this documentation. Especially since Document#add_child has limitations (a document can't have multiple root nodes).

It feels much more natural to me to do something like this instead:

```ruby
doc = XML::Document.parse("<root></root>")
```

which has implications for documentation below ...

lib/nokogiri/xml/node.rb
```diff
@@ -192,16 +198,47 @@ def decorate!
...
+      #   # Move all book nodes from bookstore_doc to other_doc.root.
+      #   bookstore_doc = fresh_bookstore_doc
+      #   book_nodes = bookstore_doc.search('//book')
+      #   other_doc.root.add_child(book_nodes)
+      #   bookstore_doc.search('//book').size # => 0
+      #   other_doc.search('//book').size     # => 4
+      #
+      # When +object+ is a string,
+      # creates a new XML::Node from the string
+      # and adds the node as the last child of +self+:
```
flavorjones, line 234: I don't think this illustrates the feature very well, which is intended to parse a string into a node (or nodes) and add them. How about something like:

```ruby
doc = XML::Document.parse("<root></root>")
doc.root.add_child("<child>first</child><child>second</child>")
doc.root.children.size # => 2
```

lib/nokogiri/xml/node.rb
```diff
@@ -192,16 +198,47 @@ def decorate!
...
-      #
-      # +node_or_tags+ can be a Nokogiri::XML::Node, a ::DocumentFragment, a ::NodeSet, or a String
-      # containing markup.
-      #
-      # Returns the reparented node (if +node_or_tags+ is a Node), or NodeSet (if +node_or_tags+ is
-      # a DocumentFragment, NodeSet, or String).
-      #
-      # Also see related method +<<+.
+      # :call-seq:
+      #   add_child(object) -> object
```
flavorjones, line 202: I've been gradually introducing a documentation convention to give a brief overview of arguments and argument types, which I'd like to ask you to use (or improve upon).

Good examples of this convention are in this file, and applying it here might look like:

```
[Parameters]
- +object+ (String, Node, NodeSet) The Node or Nodes to be added as children of this node. A String will be parsed as a fragment and the resulting NodeSet will be added as children.
```

followed by the prose explanation and examples below. WDYT?

lib/nokogiri/xml/node.rb
```diff
@@ -192,16 +198,47 @@ def decorate!
...
+      #
+      # When +object+ is an XML::Node of any kind,
+      # adds the node as the last child of +self+;
+      # sets the parent of that node as +self+:
+      #
+      #   doc = XML::Document.new
+      #   doc.children.size      # => 0
+      #   root_ele = XML::Element.new('root', doc)
+      #   root_ele.parent        # => nil
+      #   doc.add_child(root_ele)
```
flavorjones, line 215: I think this example should show adding a child to the root node, and not adding a root node to the parent.

lib/nokogiri/xml/node.rb
```diff
@@ -192,16 +198,47 @@ def decorate!
...
+      # Appends one or more children (as specified by the given +object+) to +self+;
+      # returns +object+.
+      #
+      # When +object+ is an XML::Node of any kind,
+      # adds the node as the last child of +self+;
+      # sets the parent of that node as +self+:
+      #
+      #   doc = XML::Document.new
+      #   doc.children.size      # => 0
+      #   root_ele = XML::Element.new('root', doc)
```
flavorjones, line 213: This also feels very foreign to me, see notes above.

doc/examples/bookstore_setup.rb
```diff
@@ -0,0 +1,37 @@
+require 'nokogiri'          # Make nokogiri available.
+include Nokogiri            # Allow omitting leading Nokogiri::.
```
flavorjones, line 2: I realize you're likely doing this as an editorial decision, to make lines shorter in the examples.

However, I'm not sure it's worth the character savings, because it means that the examples are no long copy-pasteable into ordinary application code. I would honestly prefer not taking shortcuts like this in order to have clearer examples.

In fact, what would you think of just having the bookstore XML be the context, and dropping the `fresh_bookstore_doc` method? Examples would then just start with

```ruby
doc = Nokogiri::XML::Document(BOOKSTORE_XML)
```

instead of

```ruby
doc = fresh_bookstore_doc
```

which isn't much longer but makes the examples more readable because there is less context the reader needs to understand.

flavorjones (commented):
In general, I like the idea of having a real-world document that the examples can demonstrate upon.

I've asked some questions and made some small suggestions. Thank you again for taking the time to work on Nokogiri! I really appreciate it.

lib/nokogiri/xml/node.rb
BurdetteLamar: What you're suggesting fits well with my own personal philosophy.  (That's how I wrote in various DEC programmer's reference manuals back at the Dawn of Time.)

However, I think you should consider the Ruby[ Documentation Guide](https://docs.ruby-lang.org/en/master/contributing/documentation_guide_md.html), first drafted by @jeremyevans and often discussed and augmented since.  It represents consensus in the Ruby community, which I think has value in itself.  But it also gives a consistent "Feel" to the Ruby documentation that readers might appreciate also finding in the gems' documentation.

But of course, "your call."

flavorjones: Do you mean [this section in the Documentation Guide](https://docs.ruby-lang.org/en/master/contributing/documentation_guide_md.html#label-Argument+Description+-28if+necessary-29)? I don't find it specifies much of anything beyond "use a labeled list", but to be clear I'm fine with that if that's what you want to go with.

The general principal is: give a brief summary of arguments and their expected types. I'm not finicky about the particulars of the style we use to present it.

flavorjones: Also, the "if necessary" bit in that section makes me groan a bit. I get it, we don't need to over-specify things for simple methods with simple arguments.

But all of these Node methods in particular take complex arguments (could be a Node, could be a NodeSet, could be a String that is cast as a NodeSet via fragment parsing) and so, yes, I think for those kinds of methods, documenting the arguments is necessary. WDYT?

flavorjones:
@BurdetteLamar Oh! I don't actually have any feelings here. I suppose if it's XML then `<root/>` is fine! 😃

lib/nokogiri/xml/node.rb
BurdetteLamar: Au naturelle.

lib/nokogiri/xml/node.rb
BurdetteLamar: Fixed.

doc/examples/bookstore_setup.rb
BurdetteLamar: Fixed.

flavorjones (changes requested):
This looks good to me, adding the examples is an obvious improvement over what was there.

I would still like to have a brief summary of the types that the input argument can be. Again, I don't feel strongly about the style used to present it, but I do feel strongly that the types and behaviors are complex enough that they justify being summarized very briefly for users who are looking for API reference material and not an explanation. We need to provide both, I think.

lib/nokogiri/xml/node.rb
```diff
@@ -192,16 +198,57 @@ def decorate!
...
+      #   doc.root.children.to_a
+      #   # => [#(Element:0x593990 { name = "foo", children = [ #(Text "FOO")] })]
+      #   doc.root.add_child('<bar>BAR</bar><baz>BAZ</baz>')
+      #   doc.root.children.to_a
+      #   # =>
+      #   [#(Element:0x593990 { name = "foo", children = [ #(Text "FOO")] }),
+      #    #(Element:0x5992dc { name = "bar", children = [ #(Text "BAR")] }),
+      #    #(Element:0x5994bc { name = "baz", children = [ #(Text "BAZ")] })]
+      #
+      # Raises ArgumentError if +object+ is not a node, a nodeset, or a string.
```
flavorjones, line 251: In general, I think information about return types, exceptions, and blocks should be the top of the docstring near the information about argument types.

lib/nokogiri/xml/node.rb
```diff
@@ -192,16 +198,57 @@ def decorate!
...
-      #
-      # +node_or_tags+ can be a Nokogiri::XML::Node, a ::DocumentFragment, a ::NodeSet, or a String
-      # containing markup.
-      #
-      # Returns the reparented node (if +node_or_tags+ is a Node), or NodeSet (if +node_or_tags+ is
-      # a DocumentFragment, NodeSet, or String).
-      #
-      # Also see related method +<<+.
+      # :call-seq:
+      #   add_child(object) -> node or nodeset
```
flavorjones, line 202: When possible, I'd like to use the proper constant name when we reference classes.

```suggestion
      #   add_child(object) -> Node or NodeSet
```

lib/nokogiri/xml/node.rb
```diff
@@ -192,16 +198,57 @@ def decorate!
...
+      #
+      # Appends child nodes to +self+.
+      #
+      # Argument +object+ is one of:
+      #
+      # - A Nokogiri::XML::Node (of any kind).
+      # - A Nokogiri::XML::NodeSet.
+      # - A string.
```
flavorjones, line 210:
```suggestion
      # - A String.
```

lib/nokogiri/xml/node.rb
```diff
@@ -192,16 +198,57 @@ def decorate!
...
-      # Returns the reparented node (if +node_or_tags+ is a Node), or NodeSet (if +node_or_tags+ is
-      # a DocumentFragment, NodeSet, or String).
-      #
-      # Also see related method +<<+.
+      # :call-seq:
+      #   add_child(object) -> node or nodeset
+      #
+      # Appends child nodes to +self+.
+      #
+      # Argument +object+ is one of:
```
flavorjones, line 206: Ugh, reading the previous documentation, I'm now realizing that object could _also_ be a DocumentFragment

lib/nokogiri/xml/node.rb
BurdetteLamar: Again, different from what's seen in Ruby core and standard libraries.  And again, I'll do as you say.

lib/nokogiri/xml/node.rb
BurdetteLamar: Same as above.

BurdetteLamar:
As above, I'll do as you say.  Thanks for taking time on this PR, as it establishes [what we're trying to do](https://www.youtube.com/watch?v=1jnxrqeOYPM).

BurdetteLamar:
Converting to draft (again), while I mull things over.

BurdetteLamar:
@flavorjones, I have greatly modified this PR.  It can (I hope) serve as a model for what's to come.

A few points:

- This now has a "reference" style of opening, consisting of a 1-line description, **Arguments:**, **Returns:**, **Raises:**.  Other methods may also need **Options:**, **Yields:**.
- The first example (arg is a Node) carries the main load;  it shows in great detail how the method works.  The other examples (other arg types) have less detail.

BurdetteLamar:
@flavorjones, let's wait on this.  I've only now discovered `CONTRIBUTING.md@Documentation`)(!).  I will study it, and maybe offer some ideas.

Issue #3432, opened 2025-02-15
`HTMLElementDescription` is defined under `::Struct` namespace


viralpraxis:
The only constant definied with [special](https://rubyapi.org/3.4/o/struct#method-c-new) "class name" form of `Struct.new` is `HTMLElementDescription`: https://github.com/sparklemotion/nokogiri/blob/d5d64bb597bcc915d7588bddb5ecea90617f9b4c/lib/nokogiri/html4/element_description_defaults.rb#L11-L27

```console
ruby -rnokogiri -e 'p Struct.constants'
[:HTMLElementDescription]
```

It’s not a big deal — I’m just curious whether this was intentional or if there’s a specific reason for it. If it can be fixed, I’ll to submit a PR, otherwise, feel free to close this. Thanks!

This patch seems to pass the tests:

```patch
diff --git a/lib/nokogiri/html4/element_description_defaults.rb b/lib/nokogiri/html4/element_description_defaults.rb
index 79a8b810..ebd1f6d0 100644
--- a/lib/nokogiri/html4/element_description_defaults.rb
+++ b/lib/nokogiri/html4/element_description_defaults.rb
@@ -8,8 +8,7 @@ module Nokogiri
       # and we don't want to clobber any methods that have been
       # defined there.
 
-      Desc = Struct.new(
-        "HTMLElementDescription",
+      HTMLElementDescription = Struct.new(
         :name,
         :startTag,
         :endTag,
@@ -2024,7 +2023,7 @@ module Nokogiri
       ].each do |descriptor|
         name = descriptor[0]
 
-        d = Desc.new(*descriptor)
+        d = HTMLElementDescription.new(*descriptor)
 
         # flatten all the attribute lists (Ruby1.9, *[a,b,c] can be
         # used to flatten a literal list, but not in Ruby1.8).
```

flavorjones:
This is not intentional, and seems like it was introduced in 2010 in 96df1268.

[Searching github doesn't show any code that is using](https://github.com/search?q=language%3Aruby+HTMLElementDescription+path%3A*.rb&type=code) the `Struct::HTMLElementDescription` constant, so it's probably safe to just remove the `class_name` from the argument list to `Struct.new`.

If you'd like to open a PR for this, I'd gladly review and accept it. Please don't change the name of the constant that the struct is assigned to, though. Thanks.

PR #3433, opened 2025-02-15
Drop `Struct::HTMLElementDescription` class

viralpraxis:

Fixes https://github.com/sparklemotion/nokogiri/issues/3432

Not applicable here?

Seems like it does not.

No idea 🙂 

Should there be a changelog entry for this?

### DIFF
--- a/lib/nokogiri/html4/element_description_defaults.rb
+++ b/lib/nokogiri/html4/element_description_defaults.rb
@@ -9,7 +9,6 @@ module Nokogiri
       # defined there.
 
       Desc = Struct.new(
-        "HTMLElementDescription",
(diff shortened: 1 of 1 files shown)


viralpraxis:
Seems like CI failure is unrelated

flavorjones:
@viralpraxis yep, there's a bug in qemu that manifests in aarch64 docker containers as a segfault during compilation. More info [here](https://github.com/tonistiigi/binfmt/issues/215) but no need to worry about it.

flavorjones:
Thank you again!!

flavorjones:
Changelog updated in c11e91cf

Issue #3439, opened 2025-02-19
Update parse options to reflect modern libxml usage


flavorjones:
## Context

In #3360 the libxml2 maintainer left some suggestions about how we're exposing and documenting some of the parse options.

He mentioned:

- DTDATTR and DTDVALID imply DTDLOAD and are unsafe as well.
- SAX1 should probably not be exposed.
- NODICT should probably not be exposed.
- XINCLUDE, NOXINCNODE and NOBASEFIX are only used by the XML Reader and XInclude API.
- HUGE is safe these days (since 2.10)

and some forward-looking statements about the upcoming 2.14 release:

- UNZIP: Enable decompression. This option has no real effect for now. The plan is that users who really need decompression start to add the option. At a later point, it will be required to enable decompression.
- NO_SYS_CATALOG: Don't use system catalogs when resolving DTDs or entities.
- CATALOG_PI: Enable oasis-xml-catalog PIs. This is a really obscure feature that should have never been enabled by default. I don't think your users need it.

## Actions

I think the actions I'd like to take re: documentation:

- Make the following bits `:nodoc:`: SAX1, NODICT
- Update documentation for DTDATTR and DTDVALID to imply DTDLOAD and include safety warnings
  - And double-check that these are all off by default
- Update documentation for the XINCLUDE set to specify they're only used by `Reader` and `Node#process_xincludes`

And the functional action I'd like to take:

- Add HUGE to all the default bitsets if the libxml2 version is >= 2.10.0

I'd like to wait until the UNZIP bit is useful before adding it. We don't expose the catalog bits, so nothing to do there.

BurdetteLamar:
Assign to me, if you like.  I can make the first pass at these changes.

flavorjones:
@BurdetteLamar Assigned you, thanks! If you want to skip the HUGE bit, that would be fine, I may want to do some additional testing and can spin it out into a separate issue.

PR #3447, opened 2025-02-23
In XML::ParseOptions.rb put :nodoc: onto constants SAX1 and NODICT

BurdetteLamar:
Implements part of [Issue 3439](https://github.com/sparklemotion/nokogiri/issues/3439).

### DIFF
--- a/lib/nokogiri/xml/parse_options.rb
+++ b/lib/nokogiri/xml/parse_options.rb
@@ -280,9 +280,8 @@ module Nokogiri
       # Remove blank nodes. Off by default.
       NOBLANKS    = 1 << 8
 
-      # :nodoc:
       # Use the SAX1 interface internally. Off by default.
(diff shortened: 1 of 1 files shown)


flavorjones:
I rebased and moved the `:nodoc:` onto the same line as the constant. E.g. where it was

```ruby
      # :nodoc:
      # Use the SAX1 interface internally. Off by default.
      SAX1        = 1 << 9
```

it is now

```ruby
      # Use the SAX1 interface internally. Off by default.
      SAX1        = 1 << 9 # :nodoc:
```

The first form ends up dropping all the subsequent constants from the documentation. This may be an rdoc bug, but it is definitely unexpected behavior.

flavorjones:
Also @BurdetteLamar keep in mind for docs-only changes you can add

```
[skip ci]
```

in the commit message to skip CI and save some electricity

PR #3437, opened 2025-02-18
dep: update libxml2 to v2.13.6

flavorjones:
**What problem is this PR intended to solve?**

Update vendored libxml2 to v2.13.6

https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.13.6

Security note: the changelog mentions CVE-2025-24928 and CVE-2024-56171, but it's not obvious that those vulnerabilities affect Nokogiri users.

PR #3448, opened 2025-02-24
dep: update libxml2 to v2.13.6 (v1.17.x branch)

flavorjones:
https://gitlab.gnome.org/GNOME/libxml2/-/releases/v2.13.6

See related #3437 and #3438

I'm not making any kind of statement or promises about whether I'll cut security releases for v1.17.x in the future. I'm doing this because Mastodon 4.2 still supports Ruby 3.0 and its dependency on ruby-saml makes it potentially impacted by the underlying libxml2 fixes.

I know somebody out there, somewhere, is going to say "I'll stay on Ruby 3.0 if Mike is going to keep cutting security updates", and hoo boy that is NOT a bet you should be making. I am the most enthusiastic supporter of "dropping support for EOL versions of Ruby" that you will ever meet, and this is NOT going to continue.

I know somebody out there, somewhere, is going to try to convince me that because I made this one security update, I'm somehow _obligated_ to continue supporting the v1.17.x branch. If you feel the urge to send me a message like that, please restrain yourself and do not make me regret doing this thing.

### DIFF
--- a/.github/workflows/downstream.yml
+++ b/.github/workflows/downstream.yml
@@ -16,7 +16,6 @@ on:
     types: [opened, synchronize]
     branches:
       - '*'
-
 jobs:
   downstream:
     name: downstream-${{matrix.name}}
--- a/dependencies.yml
+++ b/dependencies.yml
@@ -1,8 +1,8 @@
 ---
 libxml2:
-  version: "2.13.5"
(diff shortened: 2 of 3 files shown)


flavorjones:
After chatting with one of the Mastodon maintainers, I'm not going to cut a release on this branch. They've shipped https://github.com/mastodon/mastodon/releases/tag/v4.2.17 which drops support for Ruby 3.0 and are encouraging users to upgrade Ruby. 👏👏

PR #3453, opened 2025-03-07
[skip ci] Doc for Node#add_child

BurdetteLamar:


BurdetteLamar:
@flavorjones, in the event, I'm less convinced about the extended-form "Related" I've shown here.  See what you think.

lib/nokogiri/xml/node.rb
```diff
@@ -191,17 +197,70 @@ def decorate!
...
+      #   src_child_node = doc.at_css('src_child')
+      #   dst_parent_node = doc.at_css('dst_parent')
+      #   dst_parent_node.add_child(src_child_node)
+      #
+      # When +object+ is a NodeSet (see {About the Examples}[Node.html#class-Nokogiri::XML::Node-label-About+the+Examples]),
+      # appends each of its nodes to the children of +self+;
+      # returns +object+:
+      #
+      #   bookstore_doc = Nokogiri::XML::Document.parse(BOOKSTORE_XML)
+      #   book_nodeset = bookstore_doc.search('//book')
```
flavorjones, line 235: I'd prefer if we did not use `search` in examples. I've been trying to discourage the use of that method for many years. Let's please use either `xpath` or `css`.

lib/nokogiri/xml/node.rb
```diff
@@ -191,17 +197,70 @@ def decorate!
...
+      #
+      # When +object+ is a +Node+,
+      # adds it as the last child of +self+;
+      # returns +object+:
+      #
+      #   xml = '<root><src_parent><src_child/></src_parent><dst_parent><dst_child/></dst_paren></root>'
+      #   doc = Nokogiri::XML::Document.parse(xml)
+      #   src_child_node = doc.at_css('src_child')
+      #   dst_parent_node = doc.at_css('dst_parent')
+      #   dst_parent_node.add_child(src_child_node)
```
flavorjones, line 228: I'd like these examples to also illustrate the final state of the document, after the changes were made. Maybe this could be as simple as something like this:

```
# doc.to_xml
# # => '<root>...</root>`
```

lib/nokogiri/xml/node.rb
```diff
@@ -191,17 +197,70 @@ def decorate!
...
+      #
+      # When +object+ is a String,
+      # creates a +NodeSet+ object from the string;
+      # appends each of its nodes to the children of +self+;
+      # returns the +NodeSet+:
+      #
+      #   doc = Nokogiri::XML::Document.parse('<root/>')
+      #   doc.root.add_child('<foo/><bar/>')
+      #
+      # [Related]
```
flavorjones, line 256: I agree this doesn't feel particularly high-value. I've already got methods bucketed by general function, e.g.:

```
      # :section: Manipulating Document Structure
```

lib/nokogiri/xml/node.rb
```diff
@@ -191,17 +197,70 @@ def decorate!
       end
 
       # :section: Manipulating Document Structure
-
-      ###
-      # Add +node_or_tags+ as a child of this Node.
+      0
```
flavorjones, line 200: Typo

flavorjones (changes requested):
In general, I like this addition to the documentation. I made a few small notes.

lib/nokogiri/xml/node.rb
BurdetteLamar: @flavorjones, going forward, do have a preference between the two?

flavorjones: I don't have a preference, honestly. Depending on the task, an xpath query or a css selector might be easier ... Or if we want to cater to the reader, XML examples might use XPath, and HTML examples might use CSS?

BurdetteLamar: Interesting, thanks.

BurdetteLamar:
@flavorjones, I have re-worked the examples to show before/after states.

lib/nokogiri/xml/node.rb
BurdetteLamar: Fixed.

lib/nokogiri/xml/node.rb
BurdetteLamar: Re-worked.

lib/nokogiri/xml/node.rb
BurdetteLamar: Revised.

lib/nokogiri/xml/node.rb
BurdetteLamar: Fixed.

flavorjones:
@BurdetteLamar This looks great! Merging. Thank you! 🙇

Reality2byte (commented):
I love this and want to help

Issue #3458, opened 2025-03-09
[bug] Windows namespace shadowing bug


johnnyshields:
The following test case illustrates a bug related to xmlns shadowing on Windows that doesn't happen on Non-Windows MRI or JRuby. This bug changes the semantic meaning of the output XML.

```ruby
  it 'redeclares xmlns definitions when shadowed' do
    doc = Nokogiri::XML::Builder.new do |xml|
      xml['dnd'].adventure('xmlns:dnd' => 'http://www.w3.org/dungeons#') do
        xml['dnd'].party('xmlns:dnd' => 'http://www.w3.org/dragons#') do
          xml['dnd'].members('xmlns:dnd' => 'http://www.w3.org/dragons#') do
            xml['dnd'].character('xmlns:dnd' => 'http://www.w3.org/dungeons#') do
              xml['dnd'].name('Nigel', 'xmlns:dnd' => 'http://www.w3.org/dungeons#')
            end
          end
        end
      end
    end.doc.to_xml(save_with: Nokogiri::XML::Node::SaveOptions::AS_XML)

    assert_includes(doc, '<dnd:adventure xmlns:dnd="http://www.w3.org/dungeons#">')
    assert_includes(doc, '<dnd:party xmlns:dnd="http://www.w3.org/dragons#">')
    assert_includes(doc, '<dnd:members>')
    # TODO: The Windows behavior here is incorrect.
    # See: https://github.com/sparklemotion/nokogiri/issues/3457
    if Nokogiri.windows?
      assert_includes(doc, '<dnd:character xmlns:dnd="http://www.w3.org/dungeons#">')
    else
      assert_includes(doc, '<dnd:character xmlns:dnd="http://www.w3.org/dungeons#">')
    end
    assert_includes(doc, '<dnd:name>Nigel</dnd:name>')
  end
```

flavorjones:
Simplifying this test as much as I can:

```ruby
  it "foo" do
    doc = Nokogiri::XML(<<~XML)
      <dnd:adventure xmlns:dnd="http://www.w3.org/dungeons#">
        <dnd:party xmlns:dnd="http://www.w3.org/dragons#">
          <dnd:members>
          </dnd:members>
        </dnd:party>
      </dnd:adventure>      
    XML

    dungeons_ns = doc.root.namespace_definitions.find { |ns| ns.prefix == "dnd" }
    parent = doc.xpath("//ns:members", ns: "http://www.w3.org/dragons#").first

[...]

PR #3460, opened 2025-03-09
fix(java): stop using the cache for Node#namespace_definitions

### LINKED ISSUE #2543: [bug] explore removing JRuby implementation's namespace cache
**Please describe the bug**

It's not clear to me why we're using a namespace cache in the JRuby implementation. It is causing problems in edge cases like #1247 that I think would be better solved by looking up the node's parent chain for an identical namespace definition a) when it is added, and b) when the node is reparented.

This is what we do in the CRuby implementation. It's more complicated (and likely still has some bugs) but it is obviously more correct in these edge cases.

An example taken from #1247 is:

```ruby
  it "handles multiple instances of a namespace definition" do
    # this describes behavior that is broken in JRuby related to the namespace cache and should be fixed
    doc = Nokogiri::XML::Document.parse("<root>")
    child1 = doc.create_element("a", "xmlns:foo" => "http://nokogiri.org/ns/foo")
    assert_equal(1, child1.namespace_definitions.length)
    child1.namespace_definitions.first.tap do |ns|
      assert_equal("foo", ns.prefix)
      assert_equal("http://nokogiri.org/ns/foo", ns.href)
    end

    child2 = doc.create_element("b", "xmlns:foo" => "http://nokogiri.org/ns/foo")
    pending_if("broken in JRuby related to the namespace cache and should be fixed", Nokogiri.jruby?) do
      assert_equal(1, child2.namespace_definitions.length)
      child2.namespace_definitions.first.tap do |ns|
        assert_equal("foo", ns.prefix)
        assert_equal("http://nokogiri.org/ns/foo", ns.href)
      end
    end
  end
```

In summary, the second node does not have a namespace definition in the JRuby implementation. This is clearly wrong.

A partial fix would be to reimplement `Node#namespace_definitions` to look at the attributes for a namespace definition:

```java
  @JRubyMethod
  public RubyArray<?>
  namespace_definitions(ThreadContext context)
  {
    // TODO: I think this implementation would be better but there are edge cases
    RubyArray<?> nsdefs = RubyArray.newArray(context.getRuntime());
    NamedNodeMap attrs = node.getAttributes();
    for (int j = 0 ; j < attrs.getLength() ; j++) {
      Attr attr = (Attr)attrs.item(j);
      if ("http://www.w3.org/2000/xmlns/" == attr.getNamespaceURI()) {
        nsdefs.append(XmlNamespace.createFromAttr(context.getRuntime(), attr));
      }
    }
    return nsdefs;
  }
```

but we also need to then implement something like `xml_node.c:relink_namespace` in order to make this match the CRuby behavior.

flavorjones:

In this PR, I've updated `Node#namespace_definitions` to no longer uses the namespace cache, meaning it uses the attributes defined on the node to determine the return value(s).

Removing this use of the cache revealed a separate bug in the `relink_namespace` java function, which suddenly became better tested on JRuby as a result of removing the cache usage (which was hiding the bug).

Finally, removing the cache also means that `relink_namespace` now needs to remove namespace definitions that are redundant (that is, the exact prefix and URI resolve in the parent) in order to match the CRuby implementation, so I've done that, too.

Closes #2543 

Yes.

Only the Java implementation, which now behaves like the CRuby implementation (and I've resolved a pending JRuby test).

cc @johnnyshields

### DIFF
--- a/ext/java/nokogiri/XmlNode.java
+++ b/ext/java/nokogiri/XmlNode.java
@@ -438,25 +438,24 @@ public class XmlNode extends RubyObject
       if (ns_inherit.isTrue()) {
         set_namespace(context, ((XmlNode)parent(context)).namespace(context));
       }
-      return;
-    }
+    } else {
+      String currentPrefix = e.getParentNode().lookupPrefix(nsURI);
+      String currentURI = e.getParentNode().lookupNamespaceURI(prefix);
+      boolean isDefault = e.getParentNode().isDefaultNamespace(nsURI);
 
-    String currentPrefix = e.getParentNode().lookupPrefix(nsURI);
-    String currentURI = e.getParentNode().lookupNamespaceURI(prefix);
-    boolean isDefault = e.getParentNode().isDefaultNamespace(nsURI);
-
-    // add xmlns attribute if this is a new root node or if the node's
-    // namespace isn't a default namespace in the new document
-    if (e.getParentNode().getNodeType() == Node.DOCUMENT_NODE) {
-      // this is the root node, so we must set the namespaces attributes anyway
-      e.setAttribute(prefix == null ? "xmlns" : "xmlns:" + prefix, nsURI);
-    } else if (prefix == null) {
-      // this is a default namespace but isn't the default where this node is being added
-      if (!isDefault) { e.setAttribute("xmlns", nsURI); }
-    } else if (!prefix.equals(currentPrefix) || nsURI.equals(currentURI)) {
-      // this is a prefixed namespace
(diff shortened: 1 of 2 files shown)


flavorjones:
When I merge this, then I can ship a fix for #3458

flavorjones:
Side note, I *think* that this change makes the serialization work done in #3456 unnecessary, because that PR is working around the unnecessary namespaces which are being cleaned up here. That said, I'm not absolutely confident that this PR isn't going to break something, so let's keep all the code for now.

PR #3462, opened 2025-03-10
fix(cruby): reconciliation of ns on unparented nodes

### LINKED ISSUE #3459: [bug] CRuby node reparenting namespace issue
**Please describe the bug**

```ruby
  it "foo" do
    doc = Nokogiri::XML(<<~XML)
      <dnd:adventure xmlns:dnd="http://www.w3.org/dungeons#">
        <dnd:party xmlns:dnd="http://www.w3.org/dragons#">
          <dnd:members>
          </dnd:members>
        </dnd:party>
      </dnd:adventure>      
    XML

    dungeons_ns = doc.root.namespace_definitions.find { |ns| ns.prefix == "dnd" }
    parent = doc.xpath("//ns:members", ns: "http://www.w3.org/dragons#").first

    node = doc.create_element("character")
    node.namespace = dungeons_ns
    parent.add_child(node)

    # passes, indicating it's not a reparenting issues
    assert_equal("http://www.w3.org/dungeons#", node.namespace.href)
    # fails, maybe this is a serialization issue? not sure.
    assert_includes(doc.to_xml, %Q{<dnd:character xmlns:dnd="http://www.w3.org/dungeons#"/>})
  end
```

**Additional context**

I'm pretty sure this is a bug in `relink_namespaces` which I've wanted to replace for quite a while, see #2919 for other options.

flavorjones:

Closes #3459

Yes.

Fixes the behavior of CRuby. JRuby was already OK.


flavorjones:
Ugh, this is going to need a bit more work. JRuby is failing, and Viewpoint (downstream) is failing.

PR #3468, opened 2025-03-14
dep: update minimum java version from 1.8 to 21

flavorjones:

Github Actions started failing on jruby-head builds this week with an error message like:

```
  bad class file: /home/runner/.rubies/jruby-head/lib/jruby.jar(/org/jruby/Ruby.class)
    class file has wrong version 65.0, should be 61.0
```

Example failure: https://github.com/sparklemotion/nokogiri/actions/runs/13852274600/job/38789085283#step:4:17

If I understand correctly, the class file version is correlated with the JDK version, so I'm trying this to see if it fixes the issues with jruby-head.

See related commit 86652b15 which updated the minimum from 1.7 to 1.8 in 2024, and 08e55603 which updated the minimum from 1.6 to 1.7 in 2020.

@headius am I doing this right, and what would you recommend as far as timing of when to drop support for 1.8 in Nokogiri?

### DIFF
--- a/rakelib/extensions.rake
+++ b/rakelib/extensions.rake
@@ -181,6 +181,14 @@ def java?
   RUBY_PLATFORM.include?("java")
 end
 
+def java_min_version
+  if java? && JRUBY_VERSION.start_with?("10.")
+    "21"
+  else
+    "1.8"
+  end
+end
+
 def add_file_to_gem(relative_source_path)
   if relative_source_path.nil? || !File.exist?(relative_source_path)
     raise "Cannot find file '#{relative_source_path}'"
@@ -331,8 +339,8 @@ if java?
 
     ext.ext_dir = "ext/java"
     ext.lib_dir = "lib/nokogiri"
-    ext.source_version = "1.8"
-    ext.target_version = "1.8"
+    ext.source_version = java_min_version
+    ext.target_version = java_min_version
     ext.classpath = ext.gem_spec.files.select { |path| File.fnmatch?("**/*.jar", path) }.join(":")
     ext.debug = true if ENV["JAVA_DEBUG"]
   end


headius:
Hey thanks for the ping!

I had hoped you might see things just work but perhaps there's more improvement needed for setup-ruby. In ruby/setup-ruby#721 and ruby/setup-ruby#722 we modified the install to try running JRuby with the default JAVA_HOME, and if it doesn't work we choose the installed Java 21 to run with. That should make things work with JRuby 10, which is compiled for 21+.

Nokogiri should not drop support for Java 8 (1.8) until we're fully done with JRuby 9.4.x, which won't happen until next year at the earliest. The released gem should be built and released with 8 on JRuby 9.4. But for testing against JRuby 10 (which is jruby-head now in setup-ruby), you'll want to run with 21. Nokogiri can be compiled with any release, but tests will need to run with 21 for JRuby.

I am going into a meeting but we can chat through this config later.

flavorjones:
I see. OK, I've updated this PR to make the min version conditional on the JRuby version:

```ruby
def java_min_version
  if java? && JRUBY_VERSION.start_with?("10.")
    "1.9"
  else
    "1.8"
  end
end
```

Let's see what CI says.

flavorjones:
Narp. Failure is the same, at https://github.com/sparklemotion/nokogiri/actions/runs/13865422261/job/38803387178

OK, I'm stuck and must have misunderstood the symptom/failure. LMK if you know what I should be doing.

headius:
Your conditional logic is correct but for a few details:

* There is no Java 1.9, it's 9,
* JRuby 10 requires Java 21  minimum, not 9.
* The nokogiri extension can be build with Java 8 (or 1.8, same thing) and run on JRuby 10 with Java 21+ just fine.
* We need the released extension to continue to support Java 8, so it should be built and released with JRuby 9.4 still.

I think if you make the min version for JRuby 10 to be Java "21" it should work ok.

FWIW I did confirm recently that nokogiri CI tests run properly on a local JRuby 10 (with Java 21+), so this is just a matter of getting the env set up right.

headius:
It may also be possible to build the extension for Java 8/1.8 using JRuby 10 on Java 21, but there's always a chance that dependencies on Java 21 APIs might sneak in.

flavorjones:
Bumped to 21. Upstream workflow (manual run) is https://github.com/sparklemotion/nokogiri/actions/runs/13913631993

flavorjones:
@headius - What's failing is only the jruby-head job (which I test along with other upstream edge builds (cruby, libxml, html5lib) in the "upstream" workflow). The jruby 9.4 job is fine. Not urgent!

headius:
@flavorjones I pushed #3477 to remove jar-dependencies pinning and fully switch the upstream build to Java 21 using the setup-java action (https://github.com/actions/setup-java). The jruby-head build is green there.

flavorjones:
I'll close this, then. Thank you!

PR #3471, opened 2025-03-16
[ci skip] Tweaks for Node subclass list

BurdetteLamar:


flavorjones:
Good catch! I'm also hoping to include this kind of information in the sidebar for all classes if/when I get to https://github.com/ruby/rdoc/issues/1201

PR #3475, opened 2025-03-17
wip: testing whether we still need IN_LIBXML for windows linkage

flavorjones:

In a comment in #23470, @nwellnhof mentioned his surprise that we were linking with `IN_LIBXML` on windows.

This change was introduced in 7ace7915 as part of the fix for #2167, in which I struggled to get symbols to resolve on Windows.

Lots has changed in the intervening three years, let's see what happens if I remove it now.


flavorjones:
Well, apparently we don't need it anymore! 🙌

nwellnhof:
> This change was introduced in https://github.com/sparklemotion/nokogiri/commit/7ace79155e80d46540c214a2cb625525c73bed3f as part of the fix for https://github.com/sparklemotion/nokogiri/issues/2167, in which I struggled to get symbols to resolve on Windows.

Okay, I see. If you want to link libxml2 statically into another Windows DLL and export its symbols, you probably need such a hack.

Issue #3455, opened 2025-03-08
MRI Ruby vs. JRuby XML child namespace output differences


johnnyshields:
## The Problem

JRuby includes redundant `xmlns:*` attributes on child notes, even when the parent node already has the namespace. MRI correctly omits redundant `xmlns:*` attributes on child nodes.

## Who cares?

I'm working to migrate [RubySaml to purely use Nokogiri](https://github.com/SAML-Toolkits/ruby-saml/pull/745). While technically this is not a _major_ functional issue, it is making my life difficult with test assertions about the structure of the XML.

## Show me

Here's a simple script which illustrates the problem:

```ruby
require 'nokogiri'

Nokogiri::XML::Builder.new do |xml|
  xml['ds'].Signature('xmlns:ds' => 'http://www.w3.org/2000/09/xmldsig#') do
    xml['ds'].SignedInfo do
    end
  end
end.doc.to_xml(save_with: Nokogiri::XML::Node::SaveOptions::AS_XML)

# MRI
# "<?xml version=\"1.0\"?>\n<ds:Signature xmlns:ds=\"http://www.w3.org/2000/09/xmldsig#\"><ds:SignedInfo/></ds:Signature>\n"

# JRuby
# "<?xml version=\"1.0\"?>\n<ds:Signature xmlns:ds=\"http://www.w3.org/2000/09/xmldsig#\"><ds:SignedInfo xmlns:ds=\"http://www.w3.org/2000/09/xmldsig#\"/></ds:Signature>"
```

flavorjones:
Hi @johnnyshields. I personally have no plans to fix bugs or add features to the JRuby implementation at this point, though I am willing to review and merge pull requests from contributors.

Can you help me understand better why this is an issue for you? Naively, I would think that you could simply avoid making assertions on the attributes that the JRuby backend is unnecessarily adding. If you can, please show an entire test case that demonstrates why it's difficult?

flavorjones:
For context, there are a lot of differences between the JRuby and CRuby implementations; but specifically namespaces are one of those things that are different, but not wrong.

See #1875 for a similar case where JRuby serializes differently than CRuby, but the document generated is _semantically_ correct. In a perfect world, the libraries would behave the same, and so I'm sorry if this is making things difficult for you as a user of Nokogiri.

We've historically declined to try to make the two backends serialize byte-for-byte identically, and this is reflected in the https://nokogiri.org/#guiding-principles of the project:

[...]

PR #3476, opened 2025-03-18
Fix MRI Ruby vs. JRuby XML child namespace output differences (backport v1.18x)

flavorjones:

Fix MRI Ruby vs. JRuby XML child namespace output differences

Fixes https://github.com/sparklemotion/nokogiri/issues/3455

Backport of #3456

cc @johnnyshields


flavorjones:
Shipped in https://github.com/sparklemotion/nokogiri/releases/tag/v1.18.5

PR #3477, opened 2025-03-18
Fixes for recent JRuby updates

headius:
This PR will contain fixes for recent JRuby updates, including:

* jar-dependencies unpinning
* running jruby-head builds on appropriate Java setup
* anything else I find that needs updating

### DIFF
--- a/.github/workflows/upstream.yml
+++ b/.github/workflows/upstream.yml
@@ -182,6 +182,10 @@ jobs:
       - uses: actions/checkout@v4
         with:
           submodules: true
+      - uses: actions/setup-java@v4
+        with:
+          distribution: zulu
+          java-version: 21
       - uses: ruby/setup-ruby@v1
         with:
           ruby-version: "jruby-head"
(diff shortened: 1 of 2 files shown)


headius:
Switching fully to Java 21 using https://github.com/actions/setup-java appears to have done the trick.

The problem here was as I suspected: Even if JRuby can find its Java with our JAVA_HOME tweaking (in setup-ruby), the system path will still point at the default Java on GHA (currently 17). By using setup-java we switch both JAVA_HOME and PATH to 21.

Most users running tests for a Ruby library against jruby-head probably won't need the command-line `java` or `javac` updated, so we'll wait and see if other projects run into this same problem.

headius:
Supersedes changes in #3468.

headius:
Actually this can be merged any time. There's enough small deprecations and IntelliJ code suggestions I'll put them in a separate PR.

headius:
Other PR for cleanup and improvements: https://github.com/sparklemotion/nokogiri/pull/3478

flavorjones:
Manually running the upstream build against a copy of this branch at https://github.com/sparklemotion/nokogiri/actions/runs/14021514530

PR #3478, opened 2025-03-18
Clean up and modernize the JRuby extension

headius:
This PR will do some cleanup of the JRuby extension.

The first pass will be improvements recommended by IntelliJ plus some TODO comments for deprecated methods in JRuby 10.

Deprecated JRuby calls have replacements in 10 that will soon be backported to JRuby 9.4. These calls should not be updated to use the new API until it has been in the wild in a 9.4 release for at least a year.

Additional passes will update style to match modern JRuby and Java recommendations and include some minor performance improvements.

### DIFF
--- a/ext/java/nokogiri/internals/NokogiriHelpers.java
+++ b/ext/java/nokogiri/internals/NokogiriHelpers.java
@@ -741,7 +741,7 @@ public class NokogiriHelpers
     if (ruby_encoding == null) { return str; }
     Charset encoding = Charset.forName(ruby_encoding);
     if (Charset.forName(parsed_encoding).compareTo(encoding) == 0) { return str; }
-    if (str.isEmpty()) { return str; } // no need to convert
+    if (str.length() == 0) { return str; } // no need to convert
     return NokogiriHelpers.nkf(context, encoding, str);
   }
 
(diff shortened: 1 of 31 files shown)


ext/java/nokogiri/internals/NokogiriHelpers.java
```diff
@@ -730,7 +741,7 @@ public class NokogiriHelpers
     if (ruby_encoding == null) { return str; }
     Charset encoding = Charset.forName(ruby_encoding);
     if (Charset.forName(parsed_encoding).compareTo(encoding) == 0) { return str; }
-    if (str.length() == 0) { return str; } // no need to convert
+    if (str.isEmpty()) { return str; } // no need to convert
```
flavorjones, line 744: For whatever reason, this is failing in the rake-compiler-dock:1.9.1-jruby container with:

```
ext/java/nokogiri/internals/NokogiriHelpers.java:744: error: cannot find symbol
    if (str.isEmpty()) { return str; } // no need to convert
           ^
  symbol:   method isEmpty()
  location: variable str of type CharSequence
```

See https://github.com/sparklemotion/nokogiri/actions/runs/13933282491/job/38995513854?pr=3478#step:4:137 for the CI failure.

I'm afraid I don't understand why this is, and I can't reproduce it on my normal development environment even if I use the same version of JRuby. But I'm going to add a commit to revert this, then this PR should be green.

headius: Oops, yep that was a method added in Java 15. The revert is correct.

headius:
I don't think I'll get back to this for the additional style and form changes for a while (got to get JRuby 10 finished!), so you can merge this when it goes green.

flavorjones:
@headius Thanks so much for spending some time cleaning this up!

PR #3480, opened 2025-03-22
doc: Nokogiri::XML::Node.traverse is recursively called on children first

AntonKhorev:

The documentation of [traverse](https://nokogiri.org/rdoc/Nokogiri/XML/Node.html#method-i-traverse) says "Yields self and all children to block recursively". I thought that means *first yield self, then recurse on children*. But actually it's the other way around: *first recurse on children, then yield self*. This PR changes the RDoc comment to clarify that.

not necessary for this PR

no

### DIFF
--- a/lib/nokogiri/xml/node.rb
+++ b/lib/nokogiri/xml/node.rb
@@ -1334,7 +1334,7 @@ module Nokogiri
       end
 
       ####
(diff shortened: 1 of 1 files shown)


flavorjones:
Thank you for catching this! ❤️

Issue #1534, opened 2016-09-19
Nokogiri::CSS::SyntaxError: unexpected '(' after ''


hassanakram:
Hi,
Nokogiri's `search` function is throwing error on xpath starting with `(`.
For example following string throws error mentioned in the title of this issue:

```
- (//div[@data-page='SRPLIST'])[index]
- (//div[class='someclass'])[index]
```
Is it a intentional feature of search function to exclude xpaths starting from `(` ?
I have looked into the source code, the constant [LOOKS_LIKE_XPATH](https://github.com/sparklemotion/nokogiri/blob/027a151102083258f5e297c4fecfddca56904a2a/lib/nokogiri/xml/searchable.rb#L13) in `Searchable` modules blacklists above valid xpath strings.
Should I add `(` in above constant to support xpaths starting from `(` or it would break something ?

flavorjones:
Hi, thanks for asking this question.

Please use #xpath if you know your search string is xpath. This avoids putting more guessing logic into the #search method, which by its nature is an inexact heuristic.

If this doesn't help you resolve your issue, please comment and I'm happy to reopen and have a deeper conversation about your use case.

hassanakram:
Hi @flavorjones Thanks for the reply.

Yeah after finding out this problem I shifted using specific `xpath` function. I just wanted to report this if this comes in category of bug, or might help someone with the same problem.

flavorjones:
@seanstory If you are interested in improving how the `search` function determines whether the string is CSS or XPath, you are welcome to open a pull request and I would be happy to review it.

I wish I could deprecate the `#search` method, but unfortunately there are other users who would be inconvenienced if I did so; but at the same time I'm not personally interested in making it better because ... there's `#css` and `#xpath` right there.

If you don't know if your query is a CSS selector or an XPath selector, I'd love to hear more about your use case.

seanstory:
Sure. Draft docs change: https://github.com/sparklemotion/nokogiri/pull/3482

> If you don't know if your query is a CSS selector or an XPath selector, I'd love to hear more about your use case.

Our Crawler product allows users to specify a selector. Since we realized we were just using the Nokogiri `Searchable.search()` function, which documented support for xpath as well, we similarly documented that xpath support was also available. We don't ask the users which they're providing. We'll adjust our implementation. See: https://www.elastic.co/guide/en/enterprise-search/current/crawler-extraction-rules.html#crawler-extraction-rules-content-fields

[...]

PR #3482, opened 2025-03-23
Note that `Searchable#search` is not ideal

seanstory:
Related to https://github.com/sparklemotion/nokogiri/issues/1534

Search function provides no indicator that it shouldn't be used for xpath selectors.

No, because it's a docs change

No


flavorjones (changes requested):
I don't think "incomplete XPath support" is a correct description of the issue encountered in #1534. The issue described is with the overly-simple regular expression used to detect whether the query is CSS or XPath; and interpreting incorrectly for some queries.

I've suggested wording that I think is both more correct and provides readers with the "why" behind the issues they may encounter.

PR #3507, opened 2025-04-16
[skip ci] Doc for XML::Node#prepend_child

BurdetteLamar:


lib/nokogiri/xml/node.rb
```diff
@@ -212,16 +212,105 @@ def add_child(node_or_tags)
...
+      #   # Before the move.
+      #   src_parent_node.children.map {|child| child.name } # => ["FOO", "BAR"]
+      #   dst_parent_node.children.map {|child| child.name } # => ["foo", "bar"]
+      #   src_node.parent.name                               # => "PARENT"
+      #   # Move the node.
+      #   dst_parent_node.prepend_child(src_node)
+      #   # After the move.
+      #   src_parent_node.children.map {|child| child.name } # => ["BAR"]
+      #   dst_parent_node.children.map {|child| child.name } # => ["FOO", "foo", "bar"]
+      #   src_node.parent.name                               # => "parent"
```
flavorjones, line 252: This example feels, stylistically, very foreign to me. I think it may be confusing to some readers that XML is case-sensitive, so I'd like to use consistent case. The naming "src" and "dst" may be confusing to non-english speakers. I would like to use whitespace to separate stanzas (setup, action, illustration). I also think some of this can be conceptually compressed (we don't _need_ separate variables for the input strings). I also think we can use briefer syntax for the `map` calls.

WDYT of:

```ruby
doc1 = Nokogiri::XML::Document.parse('<root><parent><foo/></parent></root>')
doc2 = Nokogiri::XML::Document.parse('<root><parent><bar/></parent></root>')

original_parent = doc1.at_xpath('//parent')
node = doc1.at_xpath('//parent/foo')      # Node to move.
target_parent = doc2.at_xpath('//parent') # Node to become parent.

# Before the move.
original_parent.children.map(&:name) # => ["foo"]
target_parent.children.map(&:name)   # => ["bar"]

# Move the node.
target_parent.prepend_child(node)

# After the move.
original_parent.children.map(&:name) # => []
target_parent.children.map(&:name)   # => ["foo", "bar"]
node.parent == target_parent         # => true
```

BurdetteLamar: I think I've done as much as I want to over here at nokogiri.

For my open PRs:   close, merge, or ignore;  no hard feelings, whatever you decide.

flavorjones: @BurdetteLamar I'm genuinely sorry you feel that way. Thanks for your time.

stevecheckoway: @BurdetteLamar I appreciate all of your contributions to Nokogiri and I want to wish you luck with your future projects!

BurdetteLamar: > @BurdetteLamar I'm genuinely sorry you feel that way. Thanks for your time.

@flavorjones, I would have loved to help with the documentation for Nokogiri, but we're somewhat at odds about how we might work together.

What I look for from a maintainer/owner:

- If a documentation PR is an improvement, and contains no new technical errors, it should be merged.  If it has new technical errors, they should be pointed out and corrected before merging.
- A maintainer/owner is a professional developer;  I am a professional documentor.  I expect considerable leeway when there's a difference of opinion about documentation style and effectiveness.

Think whether we might find ways to work together.  Nokigiri deserves good documentation.

Issue #3508, opened 2025-04-20
[BUG] A heap-buffer-overflow in `hashmap_get_with_hash` at `gumbo-parser/src/hashmap.c:337:34`


JJLeo:
## Description

- Version: Latest commit [a024cff](https://github.com/sparklemotion/nokogiri/commit/a024cff101211aedfcef8372427f3af07dbe881b)
- Environment: Ubuntu 20.04.6 LTS, Clang 18.1.8
- Fuzzing harness: https://github.com/sparklemotion/nokogiri/blob/main/gumbo-parser/fuzzer/parse_fuzzer.cc

**Please let me know if you encounter any issues reproducing it — I can upload a Docker image to help.**

## Steps to reproduce

```bash
cd nokogiri/gumbo-parser
export CC="clang"
export CXX="clang++"
export CFLAGS="-fsanitize=address -g -O0 -fno-omit-frame-pointer"
export CXXFLAGS="-fsanitize=address -g -O0 -fno-omit-frame-pointer -stdlib=libc++"
export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
export OUT=$PWD
make oss-fuzz CFLAGS="$CFLAGS" CXXFLAGS="$CXXFLAGS"
./parse_fuzzer $POC
```

## Sanitizer output

```

==16==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x615000000500 at pc 0x0000006d02a2 bp 0x7fff6aa90620 sp 0x7fff6aa8fdc8
READ of size 9 at 0x615000000500 thread T0
    #0 0x6d02a1 in strcmp (/out/parse_fuzzer.fuzz+0x6d02a1)
    #1 0x602d4d in hashmap_get_with_hash /src/nokogiri/gumbo-parser/src/hashmap.c:337:34
    #2 0x602d4d in hashmap_get /src/nokogiri/gumbo-parser/src/hashmap.c:348:12
    #3 0x602d4d in gumbo_string_set_contains /src/nokogiri/gumbo-parser/src/string_set.c:40:10
    #4 0x602d4d in finish_attribute_name /src/nokogiri/gumbo-parser/src/tokenizer.c:831:9
    #5 0x5704e0 in handle_attr_name_state /src/nokogiri/gumbo-parser/src/tokenizer.c:1812:7
    #6 0x4e3df6 in gumbo_lex /src/nokogiri/gumbo-parser/src/tokenizer.c:3444:26
[...]

Issue #3500, opened 2025-04-07
[BUG] A heap-buffer-overflow in `hashmap_set_with_hash` at `gumbo-parser/src/hashmap.c:286:5`


JJLeo:
## Description

- Version: Latest commit [a024cff](https://github.com/sparklemotion/nokogiri/commit/a024cff101211aedfcef8372427f3af07dbe881b)
- Environment: Ubuntu 20.04.6 LTS, Clang 18.1.8
- Fuzzing harness: https://github.com/sparklemotion/nokogiri/blob/main/gumbo-parser/fuzzer/parse_fuzzer.cc

**Please let me know if you encounter any issues reproducing it — I can upload a Docker image to help.**

## Steps to reproduce

```bash
cd nokogiri/gumbo-parser
export CC="clang"
export CXX="clang++"
export CFLAGS="-fsanitize=address -g -O0 -fno-omit-frame-pointer"
export CXXFLAGS="-fsanitize=address -g -O0 -fno-omit-frame-pointer -stdlib=libc++"
export LIB_FUZZING_ENGINE="-fsanitize=fuzzer"
export OUT=$PWD
make oss-fuzz CFLAGS="$CFLAGS" CXXFLAGS="$CXXFLAGS"
./parse_fuzzer $POC
```

## Sanitizer output

```

root@67d74d8702cc:/src/nokogiri/gumbo-parser# ./parse_fuzzer /root/nokogiri_crash.txt 
INFO: Running with entropic power schedule (0xFF, 100).
INFO: Seed: 3305694978
INFO: Loaded 1 modules   (28 inline 8-bit counters): 28 [0x5f0fccac1440, 0x5f0fccac145c), 
INFO: Loaded 1 PC tables (28 PCs): 28 [0x5f0fccac1460,0x5f0fccac1620), 
./parse_fuzzer: Running 1 inputs 1 time(s) each.
Running: /root/nokogiri_crash.txt
=================================================================
==5312==ERROR: AddressSanitizer: heap-buffer-overflow on address 0x502000000174 at pc 0x5f0fcc941fb2 bp 0x7fff0a23b0b0 sp 0x7fff0a23a870
READ of size 8 at 0x502000000174 thread T0
    #0 0x5f0fcc941fb1 in __asan_memcpy /src/llvm-project/compiler-rt/lib/asan/asan_interceptors_memintrinsics.cpp:63:3
[...]

PR #3524, opened 2025-06-30
gumbo: Make sure to use the char* pointer as the hashmap item

flavorjones:

Previously we were passing a character string of variable length, but always copying 8 bytes into the item. It's not really a problem if the length of the string is less than 8, but it's more serious if the length is longer than 8.

Fixes https://github.com/sparklemotion/nokogiri/issues/3500
Fixes https://github.com/sparklemotion/nokogiri/issues/3508

I've added test coverage that will fail under valgrind.

The issue is only with the C (gumbo) implementation.

### DIFF
--- a/gumbo-parser/src/string_set.c
+++ b/gumbo-parser/src/string_set.c
@@ -8,12 +8,12 @@
 
 static int
 string_compare(const void *a, const void *b, void *udata) {
-  return strcmp((const char *)a, (const char *)b);
+  return strcmp(*(const char **)a, *(const char **)b);
 }
 
 static uint64_t
 string_hash(const void *item, uint64_t seed0, uint64_t seed1) {
-  const char *str = (const char *)item;
+  const char *str = *(const char **)item;
   return hashmap_xxhash3(str, strlen(str), seed0, seed1);
 }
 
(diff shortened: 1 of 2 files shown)


stevecheckoway:
I only took a quick look (I'm on vacation without a computer) so I could be missing something. It looks like this is using the address of a local variable as the value in the hash set.

I'm assuming that the implementation  is going to be storing a copy of that address (I'm basing that on the sizeof(char *) in the call to create the hash set; I haven't looked at the actual API) and using it in the lookup. If I'm correct about that, then I think this is undefined behavior.

I can take a closer look next week if that would be helpful.

Cheers from Athens!

flavorjones:
@stevecheckoway Thanks, not urgent at all. I'd love your eyes but not worth interrupting vacation for.

The prescribed usage of the hashmap library is to use a fixed-length struct as the "item". I've cheated here a bit and am using the fixed-length pointer value as the "item". The char* points at the string allocated for the attribute name, and so I'm pretty sure it's OK to reference this string during tag tokenization.

The hashmap is thrown away in `emit_current_tag`, so the lifetime of the strings is longer than the lifetime of the hashmap. I think it's OK? But would appreciate your eyeballs when you're back.

PR #3531, opened 2025-07-29
CI: Remove ruby-3.1 from test matrix on Windows

larskanis:
Ruby-3.1 is not compatible to gcc-15 and patches will not be backported since it's EOL now. Unfortunately MSYS2 is a rolling release with no option for an older gcc. Althought setup-ruby installs an older gcc-14.2 package by default, it fails to install further packages per pacman due to dependencies:

```sh
  $ pacman.exe -Sy --noconfirm --noprogressbar --needed --disable-download-timeout  mingw-w64-ucrt-x86_64-libxml2 mingw-w64-ucrt-x86_64-libxslt
  resolving dependencies...
  looking for conflicting packages...
  error: failed to prepare transaction (could not satisfy dependencies)
  :: installing mingw-w64-ucrt-x86_64-gcc-libs (15.1.0-8) breaks dependency 'mingw-w64-ucrt-x86_64-gcc-libs=14.2.0-3' required by mingw-w64-ucrt-x86_64-gcc
```

No, but I removed some test coverage. 😒

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,7 +97,7 @@ jobs:
     outputs:
       # these are usually the same, but are different once we get to ruby release candidates
       setup_ruby: "['3.1', '3.2', '3.3', '3.4']"
-      setup_ruby_win: "['3.1', '3.2', '3.3', '3.4']"
+      setup_ruby_win: "['3.2', '3.3', '3.4']"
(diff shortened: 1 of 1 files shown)


flavorjones:
Thanks! Though I worked out a fix that I think is a bit better in #3542 that skips Ruby 3.1 builds against system libraries, but otherwise maintains coverage (against vendored libraries and installing precompiled).

PR #3535, opened 2025-08-06
Fix SIGINT during parsing crashing Ruby

### LINKED ISSUE #3528: [bug] SIGINT on a running process results in a segfault
**Please describe the bug**

Crash on Ctrl+C in scripts using Nokogiri. Easily reproducible in a couple of tries.

```sh
$ timeout -s SIGINT 0.5s ruby -r nokogiri -e "loop { Nokogiri.HTML5('') }"
[BUG] Segmentation fault
c:0001 p:0000 s:0003 E:0006b0 DUMMY  [FINISH]
```

**Expected behavior**

No crash

**Environment**

```
Nokogiri: 1.18.9
Description: ruby 3.4.5 (2025-07-16 revision 20cda200d3) +PRISM [arm64-darwin24]
System: macOS 15.5 (24F74)
Model : MacBookAir10,1
```

stevecheckoway:

Closes #3528

No tests were added. Crashes are nondeterministic and only seem to happen on macOS.

Fixes a bug in the C version.

### DIFF
--- a/ext/nokogiri/gumbo.c
+++ b/ext/nokogiri/gumbo.c
@@ -441,12 +441,21 @@ parse_continue(VALUE parse_args)
   } else {
     doc = new_html_doc(NULL, NULL, NULL);
   }
-  args->doc = doc; // Make sure doc gets cleaned up if an error is thrown.
+
+  // XXX This doesn't feel like the correct approach but seems to work.
+  // We're about to build the libxml2 data structure from the Gumbo data
+  // structure. If an exception is raised during tree building, control passes
+  // to `parse_cleanup()` which needs to free the newly allocated `xmlDoc`,
+  // `doc`. If the tree is successfully constructed, then
+  // `noko_xml_document_wrap()` will create a new Ruby object which will own
+  // `doc`. In that case, `parse_cleanup()` should not free `doc`.
+  args->doc = doc;
   build_tree(doc, (xmlNodePtr)doc, output->document);
+  args->doc = NULL;
   VALUE rdoc = noko_xml_document_wrap(args->klass, doc);
+
   rb_iv_set(rdoc, "@url", args->url_or_frag);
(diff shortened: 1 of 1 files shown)


stevecheckoway:
I'm not sure this is actually the correct approach but I'm not sure how signals interact with ruby extensions. I _think_ signals are handled by the runtime such that a signal received while an extension's code is running is not handled until the extension returns to the runtime or calls into it.

In this case, when we wrap a new `xmlDoc`, we're calling into the runtime to create a new object and run its initializer. If a signal was received during `build_tree()`, will the corresponding Ruby exception be raised when we create the new wrapper object or when we return from the extension? If the former, then I think we do need something like `rb_ensure()` or perhaps `rb_protect()`. If the latter, then this code can be simplified by moving the contents of the `parse_continue`, `fragment_continue`, and `parse_cleanup` into the respective functions.

flavorjones:
@stevecheckoway Thanks for debugging this issue and writing up this PR! I'll take a look shortly.

flavorjones:
@stevecheckoway This seems the right diagnosis and an adequate fix. If I could reproduce this issue on Linux (I still can't!) I'd be more comfortable hammering it into shape with rb_ensure or rb_protect, but without visibility into what's actually happening on the mac I'm fine shipping this PR as-is. WDYT?

The CI failures re: Ruby 3.1 on Windows and MacOS are not related, I'll try to clean that up in a separate PR.

stevecheckoway:
@flavorjones I think we should just merge this for now since it appears to fix the problem on macOS and I don't have any reason to suspect it could lead to additional problems.

It might be interesting to dig into this behavioral difference between macOS and Linux in the future but unfortunately, I don't have the time now.

flavorjones:
I've rebased this PR onto `main` which should address the flaky Ruby 3.1 tests.

PR #3559, opened 2025-09-20
fixes dangling-doc-comments warnings

yokolet:

When Nokogiri is compiled on JRuby by `bundle exec rake compile test` command, it raises a lot of warnings. Those are just warnings, but it's not easy to see what are happening during compilation. Among those warnings, this PR fixes dangling-doc-comments warnings.

No.

No.


flavorjones:
@yokolet Nice to see you again! ❤️ Thank you for this.

PR #3562, opened 2025-10-05
Fix serial warnings

yokolet:

This PR fixes all warnings of "warning: [serial] non-transient instance field of a serializable class declared with a non-serializable type", when Nokogiri is compiled on JRuby 9.4.

Fixing strategy is:
- The instance field type is from outside of Nokogiri: add transient declaration
- The instance field  type is from Nokogiri but make it serializable requires many changes: add transient declaration
- The instance field  type is from Nokogiri and make it serializable is simple: change the type to serializable

No.

It should not change any Java implementation behavior. The command `bundle exec rake clean compile test` result on JRuby 9.4 shows the changes don't break any.


flavorjones:
Thank you, @yokolet!

PR #3617, opened 2026-04-09
Hoist lambda to avoid EP escape

tekknolagi:
The CRuby VM will move the VM frame from the stack to the heap if a
`lambda` captures it. Since this `lambda` does not use any locals from
its parent frame, we can hoist it and avoid this overhead.

This should have no functional changes and does not affect C or Java
implementations.

### DIFF
--- a/lib/nokogiri/html5/node.rb
+++ b/lib/nokogiri/html5/node.rb
@@ -36,6 +36,10 @@ module Nokogiri
         result
       end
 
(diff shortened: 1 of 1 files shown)


XrXr:
This also removes a GC allocation per run

flavorjones:
The libxmlruby failure is unrelated, I'll fix that separately.

Issue #3576, opened 2025-11-15
[install] 1.8.10 cannot find nokogiri_gumbo.h termux


russano09:
**Have you read and followed the installation tutorial at http://www.nokogiri.org/tutorials/installing_nokogiri.html?**

- [x] Yes!

**What is the complete output of `gem install` or `bundle install`?**

```
% bundle install
Fetching gem metadata from https://rubygems.org/.......
Resolving dependencies...
Bundler itself does not use binstubs because its version is selected by RubyGems
Installing nokogiri 1.18.10 (was 1.18.9) with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

current directory:
/data/data/com.termux/files/home/scripts/test/vendor/bundle/ruby/3.4.0/gems/nokogiri-1.18.10/ext/nokogiri
/data/data/com.termux/files/usr/bin/ruby extconf.rb --use-system-libraries
--with-xml2-lib\=/data/data/com.termux/files/usr/lib
--with-xml2-include\=/data/data/com.termux/files/usr/include/libxml2/libxml
--with-xslt-lib\=/data/data/com.termux/files/usr/lib
--with-xslt-include\=/data/data/com.termux/files/usr/include/libxslt
--with-iconv-lib\=/data/data/com.termux/files/usr/lib
--with-iconv-include\=/data/data/com.termux/files/usr/include
--with-gumbo-lib\=/data/data/com.termux/files/usr/lib
--with-gumbo-include\=/data/data/com.termux/files/usr/include
checking for whether -std=c99 is accepted as CFLAGS... no
checking for whether -Wno-declaration-after-statement is accepted as CFLAGS... no
checking for whether -O2 is accepted as CFLAGS... no
checking for whether -g is accepted as CFLAGS... no
checking for whether -Winline is accepted as CFLAGS... no
checking for whether -Wmissing-noreturn is accepted as CFLAGS... no
checking for whether -Wconversion is accepted as CFLAGS... no
Building nokogiri using system libraries.
checking for gzdopen() in -lz... yes
checking for xmlParseDoc() in -lxml2... no
checking for xmlParseDoc() in -llibxml2... no
checking for libxml-2.0 using `pkg_config`... yes: ["-I/data/data/com.termux/files/usr/include/libxml2
-I/data/data/com.termux/files/usr/include ", "-L/data/data/com.termux/files/usr/lib", "-lxml2"]
checking for xmlParseDoc() in -lxml2... yes
[...]

Issue #3634, opened 2026-06-10
[install] Nokogiri 1.19.3 bundle install failing on MacOs, gem install is fine


sarah-fflow:
**Have you read and followed the installation tutorial at http://www.nokogiri.org/tutorials/installing_nokogiri.html?**

- Yes!

**What is the complete output of `gem install` or `bundle install`?**

gem install nokogiri
```
Successfully installed nokogiri-1.19.3-arm64-darwin
Parsing documentation for nokogiri-1.19.3-arm64-darwin
Done installing documentation for nokogiri after 0 seconds
1 gem installed
```

bundle install

```
Local specification for rexml-3.2.8 has different dependencies than the remote gem, ignoring it
Local specification for net-smtp-0.4.0.1 has different dependencies than the remote gem, ignoring it
WARN: Unresolved or ambiguous specs during Gem::Specification.reset:
      psych (>= 4.0.0)
      Available/installed versions of this gem:
      - 5.3.1
      - 5.1.2
WARN: Clearing out unresolved specs. Try 'gem cleanup <gem>'
Please report a bug if this causes problems.
Installing nokogiri 1.19.3 with native extensions
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    current directory: /Users/sarah/.rvm/gems/ruby-3.3.4/gems/nokogiri-1.19.3/ext/nokogiri
/Users/sarah/.rvm/rubies/ruby-3.3.4/bin/ruby extconf.rb --use-system-libraries
checking for whether -std=c99 is accepted as CFLAGS... no
checking for whether -Wno-declaration-after-statement is accepted as CFLAGS... no
checking for whether -O2 is accepted as CFLAGS... no
checking for whether -g is accepted as CFLAGS... no
checking for whether -Winline is accepted as CFLAGS... no
checking for whether -Wmissing-noreturn is accepted as CFLAGS... no
checking for whether -Wconversion is accepted as CFLAGS... no
checking for whether -Wno-error=unused-command-line-argument-hard-error-in-future is accepted as CFLAGS... no
checking for whether -Wno-unknown-warning-option is accepted as CFLAGS... no
checking for whether -I/Library/Developer/CommandLineTools/SDKs/MacOSX.sdk/usr/include/libxml2 is accepted as CPPFLAGS... no
Building nokogiri using system libraries.
checking for gzdopen() in -lz... yes
checking for xmlParseDoc() in -lxml2... yes
[...]

PR #3653, opened 2026-06-30
fix: fatal error: 'nokogiri_gumbo.h' file not found

lustickd:

Fixes issues #3576 and #3634.

append_cppflags is forcing -Werror which causes warning from Ruby to bubble up to error instead. This makes paths say they are not valid and don't get added to CPPFLAGS even though they should be fine, eg:

```
append_cppflags: checking for whether -I/.../ext/nokogiri/ports/arm64-darwin25/libgumbo/1.0.0-nokogiri/include is accepted as CPPFLAGS... -------------------- no
```

This change modifies $CPPFLAGS directly for libgumbo which skips setting -Werror. It's not necessary to use append_cppflags to check for compiler compatibility because all C compilers support `-I<dir>`.

No I only built it locally (ruby 3.3.7) to verify it succeeds:
```
gem install --local nokogiri-1.20.0.dev.gem --platform ruby
```

No.

### DIFF
--- a/ext/nokogiri/extconf.rb
+++ b/ext/nokogiri/extconf.rb
@@ -1114,7 +1114,7 @@ else
       end
     end
   end
-  append_cppflags("-I#{File.join(libgumbo_recipe.path, "include")}")
+  $CPPFLAGS << " -I#{File.join(libgumbo_recipe.path, "include")}"
   $libs = $libs + " " + File.join(libgumbo_recipe.path, "lib", "libgumbo.a")
   $LIBPATH = $LIBPATH | [File.join(libgumbo_recipe.path, "lib")]
   ensure_func("gumbo_parse_with_options", "nokogiri_gumbo.h")


lustickd:
Just to confirm, grepped the log and all calls to append_cflags and append_cppflags are failing:

```
append_cppflags: checking for whether  -I/opt/homebrew/include is accepted as CPPFLAGS... -------------------- no
append_cflags: checking for whether -std=c99 is accepted as CFLAGS... -------------------- no
append_cflags: checking for whether -Wno-declaration-after-statement is accepted as CFLAGS... -------------------- no
append_cflags: checking for whether -O2 is accepted as CFLAGS... -------------------- no
append_cflags: checking for whether -g is accepted as CFLAGS... -------------------- no
append_cflags: checking for whether -Winline is accepted as CFLAGS... -------------------- no
append_cflags: checking for whether -Wmissing-noreturn is accepted as CFLAGS... -------------------- no
append_cflags: checking for whether -Wconversion is accepted as CFLAGS... -------------------- no
append_cflags: checking for whether -Wno-error=unused-command-line-argument-hard-error-in-future is accepted as CFLAGS... -------------------- no
append_cflags: checking for whether -Wno-unknown-warning-option is accepted as CFLAGS... -------------------- no
append_cppflags: checking for whether -DNOKOGIRI_PACKAGED_LIBRARIES is accepted as CPPFLAGS... -------------------- no
append_cppflags: checking for whether -DNOKOGIRI_LIBXML2_PATCHES="\"\"" is accepted as CPPFLAGS... -------------------- no
append_cppflags: checking for whether -DNOKOGIRI_LIBXSLT_PATCHES="\"\"" is accepted as CPPFLAGS... -------------------- no
append_cppflags: checking for whether -I/Users/lustick/.local/share/mise/installs/ruby/3.3.7/lib/ruby/gems/3.3.0/gems/nokogiri-1.20.0.dev/ext/nokogiri/ports/arm64-darwin25/libgumbo/1.0.0-nokogiri/include is accepted as CPPFLAGS... -------------------- no
```

I'm not sure if any of the others matter except maybe `-I/opt/homebrew/include`, and I assume `-std=c99` and `-O2` are universally supported flags and wouldn't need the `try_cflags` check.

flavorjones:
@lustickd Thanks for the PR. I'm not necessarily opposed to making this change, but it does seem like it's addressing the symptom and not the root cause: Why is `-Werror` causing a problem in the first place?

The CI matrix is huge and covers every supported platform, and this problem literally never comes up. Help me understand the root causes, and then I can better understand why this is the correct fix. 🙏

lustickd:
It's not clear to me why our specific version of ruby has a warning, or why it needs to set -Werror to check if a flag is valid. But as far as I know `-I` is always valid so I don't think setting `-Werror` with append_cppflags is a reliable way of checking supported flags. IMO this is a bug within `append_cppflags`, it incorrectly says `-I` is not valid.

The readme says this should be supported on Ruby >= 3.2 and I'm testing with 3.3.7. If we need different versions for compilation because of Ruby warnings it would be nice to have a supported list of which versions work. Unfortunately our version of debian is too old to use the precompiled binaries. Interestingly it _does_ succeed while compiling with ruby 3.3.7 on debian, not sure why `-Werror` only fails on Mac.

flavorjones:
@lustickd Thanks for the reply.

> It's not clear to me why our specific version of ruby has a warning, or why it needs to set -Werror to check if a flag is valid.

Same here, which is why I asked. This is worth digging into, and I can't do myself without better instructions on how to reproduce the issue. Can you give me precise instructions on how to reproduce this?

> But as far as I know -I is always valid so I don't think setting -Werror with append_cppflags is a reliable way of checking supported flags. IMO this is a bug within append_cppflags, it incorrectly says -I is not valid.

Well `append_cppflags` is simply reflecting the fact that your Ruby was built with options that won't actually successfully compile on your platform! To be clear -- the `-Werror` is very likely something that your Ruby was built with, and so is injecting into the build process for all C extensions. We still have not gotten to the root cause (as noted above) and so I don't think the perspective of "this is a bug in append_cppflags" is helpful.

I want to impress upon you as strongly as possible that the CI matrix is extremely broad -- we test both source-compilation and native-installation on every version of Ruby and every supported platform. And this problem has not shown up. This STRONGLY suggests that the issue is in "how is your Ruby built", specifically "are there compiler options being propagated that do not actually work on my platform".

What I'm saying is: I suspect this problem has to do with how your Ruby is installed. Which is why I'm hesitant to merge any sort of fix that addresses symptoms when we have not yet identified the root cause(s).

lustickd:
It seems to be related to compiling ruby with clang21 on Darwin which causes this warning: `-Wdefault-const-init-field-unsafe`. https://bugs.ruby-lang.org/issues/21629

It would be appreciated if that didn't block including libgumbo header files. 😅

lustickd:
I got it to build successfully with this config:
```sh
bundle config build.nokogiri --with-cppflags="-Wno-default-const-init-field-unsafe"
```

Then `checking for whether -I/opt/homebrew/include is accepted as CPPFLAGS...` threw an error because it couldn't find `jemalloc/jemalloc.h` (which is in /opt/homebrew/include) so I had to set it manually with CPATH.
```
export CPATH=/opt/homebrew/include
```

stevecheckoway:
I can reproduce that using rbenv install 3.4.7. The fix is in Ruby 3.4.8 according to the [release notes](https://github.com/ruby/ruby/releases/tag/v3_4_8) though.

Is there a reason you cannot use the latest version of the 3.4 branch at least (which appears to be 3.4.10)?

lustickd:
Well I have a workaround already, just trying to help other people avoid this pitfall and this seemed like a simple fix. If you want it to fail on any ruby warning then should I exit at the top with an explicit error message that it won't compile due to a warning? I think that would be a lot less confusing than getting file not found errors.

stevecheckoway:
I believe that `-Werror` is being added by `append_cppflags()`. Specifically, [`append_cppflags()`](https://github.com/ruby/ruby/blob/a1392b966535f60d06d0da1db03df5890d325f18/lib/mkmf.rb#L719) calls [`try_cppflags()`](https://github.com/ruby/ruby/blob/a1392b966535f60d06d0da1db03df5890d325f18/lib/mkmf.rb#L710) which calls [`try_compile()`](https://github.com/ruby/ruby/blob/a1392b966535f60d06d0da1db03df5890d325f18/lib/mkmf.rb#L658) (by an alias). `try_compile()` is responsible for adding `-Werror` if its `werror` keyword argument is true.

The upshot here is that if we want to disable `-Werror`, we can do so by calling `append_cppflags(flag, werror: false)`.

It seems to me that `-Werror` is mostly useful for development and CI. End users, in contrast, don't benefit. Perhaps we could conditionally disable it.

lustickd:
Alternately to keep the -Werror behavior without failing on previous compile warnings, you could compile a minimal program with only the given flags. This is essentially the same thing `try_compile` is doing:

```rb
require 'open3'
require 'rbconfig'

$CPPFLAGS ||= ''

def add_cppflags(*flags)
  cc = RbConfig::CONFIG['CC'] || 'cc'
  werror = RbConfig::CONFIG['WERRORFLAG'] || '-Werror'
  cmd = "#{cc} #{werror} #{flags.join(' ')} -xc -c - -o /dev/null"
  _out, err, _status = Open3.capture3(cmd, stdin_data: "int main(void){return 0;}\n")
  if err.strip.empty?
    $CPPFLAGS << ' ' << flags.join(' ')
  else
    raise err
  end
end

add_cppflags('-O2', '-Winvalid', '-I/opt/homebrew/include')

puts $CPPFLAGS
```

Output for file named check_warnings.rb:
```sh
error: unknown warning option '-Winvalid' [-Werror,-Wunknown-warning-option]
1 error generated.
        from check_warnings.rb:18:in `<main>'
```

lustickd:
Another thing I was wondering is why are ruby headers set on CPPFLAGS? I thought this was using a C compiler, does it need headers/warnings from ruby? Maybe the whole issue can be avoided by not setting the headers from ruby on the C compiler flags.

stevecheckoway:
> Another thing I was wondering is why are ruby headers set on CPPFLAGS? I thought this was using a C compiler, does it need headers/warnings from ruby? Maybe the whole issue can be avoided by not setting the headers from ruby on the C compiler flags.

The header file the extension is including contains the structure definitions and function prototypes needed to interoperate with Ruby. Every Ruby extension includes `ruby.h`.

lustickd:
Sorry I meant does `append_cppflags` actually need to include `ruby.h` to check if the flag is valid? If the check didn't include ruby.h there would be no extra warning and it would compile successfully.

According to mkmf.log its trying to compile this to test the flag:
```c
#include "ruby.h"

int main(int argc, char **argv)
{
  return !!argv[argc];
}
```

IMO this would be more reliable:
```c
int main(int argc, char **argv)
{
  return !!argv[argc];
}
```

I think this is a bug within mkmf/append_cppflags because including a header file should have no effect on whether a compiler flag is valid. It's just a side effect of the shared `create_tmpsrc` method prepending `COMMON_HEADERS` to everything even if they don't need it.

stevecheckoway:
Ah sorry, I misunderstood.

I don't know if it's a bug or not, but mkmf is indeed the one that's including that header: https://github.com/ruby/ruby/blob/a1392b966535f60d06d0da1db03df5890d325f18/lib/mkmf.rb#L2863

jvcarli:
Can this please be merged? I stumbled across this problem months ago and fixed it with some ugly hacks. This PR fixed it. I am stuck with `ruby 3.2.2 [x64-mswin64_140]` so upgrading is not an option for me.

lustickd:
@stevecheckoway I think using `append_cppflags(flag, werror: false)` makes sense, want me to do that for the rest of the file too?

@flavorjones 
For reproducing it in the CI matrix it depends on the version of clang. So it would need a config with clang21 compiling ruby 3.2 or 3.3.

I'm not really sure how to set that up but here's my versions:
Ruby 3.3.7 [arm64-darwin25]
Apple clang version 21.0.0 (clang-2100.1.1.101)
Target: arm64-apple-darwin25.6.0

PR #3654, opened 2026-07-06
Harden workflows

flavorjones:

Harden github workflows. Pin all actions to commit SHAs, add a lint-actions job running actionlint and zizmor, and resolve zizmor findings across the ci, upstream, downstream, and generate-ci-images workflows


flavorjones:
BSD failures are because of https://github.com/ruby/rubygems/pull/9676

PR #3673, opened 2026-08-15
gumbo: fix foster parented text not being merged into preceding text node

txgk:
Hi, I'm maintainer of [gumbo-parser](https://codeberg.org/gumbo-parser/gumbo-parser) repo.

While tinkering with the Gumbo's test suite over at Codeberg, I came across a very nasty bug in text foster parenting that could never have been detected by the html5lib-based tree-construction tests: the Python bindings rebuild Gumbo's tree through html5lib's treebuilder, whose insertText() appends character data to a trailing text node, so Gumbo's erroneously split text nodes were merged back together before the tree was ever compared against the expected output.

I've compared the proposed change against current Chromium and Firefox, and it matches their behavior in every affected test case.

Chromium tree example:

image

Firefox tree example:

image

Best regards, Grigory

### DIFF
--- a/gumbo-parser/src/parser.c
+++ b/gumbo-parser/src/parser.c
@@ -973,7 +973,33 @@ static void maybe_flush_text_node_buffer(GumboParser* parser) {
   );
 
   InsertionLocation location = get_appropriate_insertion_location(parser, NULL);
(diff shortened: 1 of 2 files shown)


flavorjones:
@txgk Thanks for the patch. I'll take a look!

Issue #2801, opened 2023-02-28
[bug] HTML5 document encoding differs from HTML4


flavorjones:
**Please describe the bug**

The encoding of an HTML5 document differs from the encoding of an HTML4 document:

```
Nokogiri::HTML4::Document.parse(File.open(SHIFT_JIS_HTML)).encoding == "Shift_JIS"
Nokogiri::HTML5::Document.parse(File.open(SHIFT_JIS_HTML)).encoding == "UTF-8"
```

I haven't had time to dig into why this is (and whether it's intended behavior), so I'm opening this issue to look into it later. cc @stevecheckoway 

**Help us reproduce what you're seeing**

```ruby
#! /usr/bin/env ruby

require 'nokogiri'
require_relative 'test/helper'

class Test < Nokogiri::TestCase
  describe "document encoding" do
    it "should be shift_jis" do
      # passes
      assert_equal("Shift_JIS", Nokogiri::HTML4::Document.parse(File.open(SHIFT_JIS_HTML)).encoding)

      # fails
      assert_equal("Shift_JIS", Nokogiri::HTML5::Document.parse(File.open(SHIFT_JIS_HTML)).encoding)
    end
  end
end
```

**Expected behavior**

I think these should both be the same?

flavorjones:
There are a couple of things worth digging into here:

1. why isn't the HTML4 EncodingReader discovering the encoding in the "given a string" case?
2. looks like passing a File and an encoding isn't supported by HTML5 (it raises an exception)
3. can we make the HTML5 encoding detection match the HTML5 behavior? I think we might be able to use the EncodingReader if it's helpful.

stevecheckoway:
Since Gumbo doesn't support anything other than UTF-8, it performs the standard [encoding detection pre-scan](https://html.spec.whatwg.org/multipage/parsing.html#determining-the-character-encoding) that browsers are supposed to perform to decide on the encoding and then uses that to convert to UTF-8 to pass to Gumbo.

I'm wasn't sure what the `encoding` property is supposed to return so I [set it to the encoding of the strings in the Document](https://github.com/sparklemotion/nokogiri/blob/4ba43be4dcf5fb0c5c56110360c47db06908c1a0/lib/nokogiri/html5/document.rb#L124) itself rather than the encoding of the input document. If we want the latter, that is probably easy to change.

[...]

PR #3674, opened 2026-08-17
fix: html5 encoding detection for IO input

kataokatsuki:

#2801, items 2 and 3 in @flavorjones's comment there: passing a File together with an
encoding raises `TypeError`, and HTML5 encoding detection does not follow the HTML5
behavior. A Shift_JIS document passed as `File.open(path)` parses with replacement
characters, where the same bytes read with `File.binread` parse correctly.

`#reencode`, where a BOM or a `meta` charset declaration is read, only runs for input
tagged `ASCII-8BIT`, and `File.open(path).read` tags its result with
`Encoding.default_external`, which reflects the locale rather than the document. This PR
prefers a declared encoding instead, while preserving the IO's encoding when the
document declares none. An explicit `encoding:` argument to `parse` and an IO configured
for transcoding still take precedence. An IO's external encoding does not: Ruby records
nothing about whether it was chosen by the caller or inherited from the locale, so it
cannot stand in for what the document says about itself.

One input shape changes behavior: an IO whose document declares an encoding that cannot
decode its bytes now raises. Which error depends on how the bytes fail —
`Encoding::InvalidByteSequenceError` when the sequence is not valid in the declared
encoding, `Encoding::UndefinedConversionError` when the sequence is valid there but that
position is unassigned. Both match what `Nokogiri::HTML5(File.binread(path))` of the same
bytes already raises.

This leaves `doc.encoding` alone, so the assertions in the issue still report `"UTF-8"`.
Per @stevecheckoway's comment there, that value is deliberately the encoding of the
strings in the document, and how it should interact with serialization is still open.

Twelve tests in `test/html5/test_encoding.rb`; eight fail without this change, and four
cover behavior that must be preserved.

CRuby only. HTML5 is not available on JRuby.

### DIFF
--- a/lib/nokogiri/html5.rb
+++ b/lib/nokogiri/html5.rb
@@ -285,24 +285,33 @@ module Nokogiri
       def read_and_encode(string, encoding)
         # Read the string with the given encoding.
         if string.respond_to?(:read)
-          external = string.external_encoding if string.respond_to?(:external_encoding)
           internal = string.internal_encoding if string.respond_to?(:internal_encoding)
+          # A StringIO holds an already-decoded String and reports that String's encoding rather
+          # than a locale default, so wrapping a String in one must not change how it decodes.
+          wraps_a_string = StringIO === string
           string = string.read
 
           if encoding
-            string = string.dup
             string.force_encoding(encoding)
-          elsif internal.nil? && external == Encoding.default_external
-            # The IO was opened without an explicit encoding, so its external encoding is the
-            # process default and says nothing about the document. Prefer an encoding the document
-            # declares itself, and keep the IO's encoding when the document declares none.
+          elsif internal.nil? && !wraps_a_string
+            # Ruby records nothing about whether an IO's external encoding was chosen by the
+            # caller or inherited from the locale, so it cannot stand in for what the document
+            # says about itself. Prefer the document's own declaration, and let a caller who
+            # needs to override it say so with the `encoding:` argument.
+            #
+            # When the document declares nothing, keep the IO's encoding. The standard's answer
+            # there is windows-1252 and #reencode falls back to ISO_8859_1, so this is a third
+            # answer and a deliberate one: it is the conservative choice and it leaves documents
+            # that parse correctly today parsing the same way.
             declared = detect_encoding(string)
             if declared
-              string = string.dup
               begin
                 string.force_encoding(declared)
               rescue ArgumentError
-                # an unknown encoding name is not a reason to discard what the IO gave us
+                # The document named an encoding Ruby does not know. #reencode makes the same
+                # choice for the binary path, so make it here too rather than keeping an IO
+                # encoding this branch has already decided says nothing about the document.
+                string.force_encoding(Encoding::ISO_8859_1)
               end
             end
           end
@@ -353,6 +362,41 @@ module Nokogiri
         body.encode(Encoding::UTF_8)
       end
 
+      # The standard's charset labels are not Ruby encoding names. "Get an encoding" maps
+      # roughly 220 labels onto encodings, while String#force_encoding accepts only Ruby's own
+      # names and aliases. Normalize the labels Ruby rejects but that appear in real documents,
+      # so that the IO path and the binary path agree on the same declaration. The standard
+      # folds latin1 and iso-8859-1 into windows-1252; Ruby keeps them as ISO-8859-1, and so
+      # does #reencode's fallback, so latin1 follows its synonym here rather than the table.
+      LABEL_ALIASES = {
+        "utf8" => "UTF-8",
+        "unicode-1-1-utf-8" => "UTF-8",
+        "latin1" => "ISO-8859-1",
+        "shift-jis" => "Shift_JIS",
+        "x-sjis" => "Shift_JIS",
+        "ms932" => "Shift_JIS",
+        "csshiftjis" => "Shift_JIS",
+        "x-user-defined" => "Windows-1252",
+      }.freeze
+      private_constant :LABEL_ALIASES
+
+      # Returns nil for an empty label, so that `charset=""` counts as no declaration rather
+      # than as a declaration of an encoding named "".
+      def normalize_encoding_label(label)
+        # "Get an encoding" strips leading and trailing ASCII whitespace and matches
+        # case-insensitively, so normalize first and keep the normalized form even when the
+        # label is not in the table above -- Encoding.find is case-insensitive too.
+        label = label.strip.downcase
+        return if label.empty?
+
+        label = LABEL_ALIASES.fetch(label, label)
+        # The standard's prescan replaces two of the encodings it gets: "If charset is
+        # UTF-16BE/LE, then set charset to UTF-8" (a document that really were UTF-16 could not
+        # have carried an ASCII-readable meta), and "If charset is x-user-defined, then set
+        # charset to windows-1252". The latter is in LABEL_ALIASES above.
+        /\Autf-16/i.match?(label) ? "UTF-8" : label
+      end
+
       # Return the encoding the document declares for itself, or nil when it declares none. Only
       # the first 1024 bytes are examined, per the HTML5 standard's prescan.
       def detect_encoding(body, content_type = nil)
@@ -371,6 +415,7 @@ module Nokogiri
         # look for a charset in a content-encoding header
         if content_type
           encoding = content_type[/charset=["']?(.*?)($|["';\s])/i, 1]
+          encoding = normalize_encoding_label(encoding) if encoding
           return encoding if encoding
         end
 
@@ -378,6 +423,7 @@ module Nokogiri
         data = head.gsub(/<!--.*?(-->|\Z)/m, "")
         data.scan(/<meta.*?>/im).each do |meta|
           encoding = meta[/charset=["']?([^>]*?)($|["'\s>])/im, 1]
+          encoding = normalize_encoding_label(encoding) if encoding
           return encoding if encoding
         end
 
--- a/lib/nokogiri/html5.rb
+++ b/lib/nokogiri/html5.rb
@@ -285,10 +285,26 @@ module Nokogiri
       def read_and_encode(string, encoding)
         # Read the string with the given encoding.
         if string.respond_to?(:read)
-          string = if encoding.nil?
-            string.read
-          else
-            string.read(encoding: encoding)
+          external = string.external_encoding if string.respond_to?(:external_encoding)
+          internal = string.internal_encoding if string.respond_to?(:internal_encoding)
+          string = string.read
+
+          if encoding
+            string = string.dup
+            string.force_encoding(encoding)
+          elsif internal.nil? && external == Encoding.default_external
+            # The IO was opened without an explicit encoding, so its external encoding is the
+            # process default and says nothing about the document. Prefer an encoding the document
+            # declares itself, and keep the IO's encoding when the document declares none.
+            declared = detect_encoding(string)
+            if declared
+              string = string.dup
+              begin
+                string.force_encoding(declared)
+              rescue ArgumentError
+                # an unknown encoding name is not a reason to discard what the IO gave us
+              end
+            end
           end
         else
           # Otherwise the string has the given encoding.
@@ -322,33 +338,8 @@ module Nokogiri
       #
       def reencode(body, content_type = nil)
         if body.encoding == Encoding::ASCII_8BIT
-          encoding = nil
-
-          # look for a Byte Order Mark (BOM)
-          initial_bytes = body[0..2].bytes
-          if initial_bytes[0..2] == [0xEF, 0xBB, 0xBF]
-            encoding = Encoding::UTF_8
-          elsif initial_bytes[0..1] == [0xFE, 0xFF]
-            encoding = Encoding::UTF_16BE
-          elsif initial_bytes[0..1] == [0xFF, 0xFE]
-            encoding = Encoding::UTF_16LE
-          end
-
-          # look for a charset in a content-encoding header
-          if content_type
-            encoding ||= content_type[/charset=["']?(.*?)($|["';\s])/i, 1]
-          end
-
-          # look for a charset in a meta tag in the first 1024 bytes
-          unless encoding
-            data = body[0..1023].gsub(/<!--.*?(-->|\Z)/m, "")
-            data.scan(/<meta.*?>/im).each do |meta|
-              encoding ||= meta[/charset=["']?([^>]*?)($|["'\s>])/im, 1]
-            end
-          end
-
           # if all else fails, default to the official default encoding for HTML
-          encoding ||= Encoding::ISO_8859_1
+          encoding = detect_encoding(body, content_type) || Encoding::ISO_8859_1
 
           # change the encoding to match the detected or inferred encoding
           body = body.dup
@@ -360,6 +351,37 @@ module Nokogiri
         end
 
         body.encode(Encoding::UTF_8)
+      end
+
+      # Return the encoding the document declares for itself, or nil when it declares none. Only
+      # the first 1024 bytes are examined, per the HTML5 standard's prescan.
+      def detect_encoding(body, content_type = nil)
+        head = body.byteslice(0, 1024).b
+
+        # look for a Byte Order Mark (BOM)
+        initial_bytes = head[0..2].bytes
+        if initial_bytes[0..2] == [0xEF, 0xBB, 0xBF]
+          return Encoding::UTF_8
+        elsif initial_bytes[0..1] == [0xFE, 0xFF]
+          return Encoding::UTF_16BE
+        elsif initial_bytes[0..1] == [0xFF, 0xFE]
+          return Encoding::UTF_16LE
+        end
+
+        # look for a charset in a content-encoding header
+        if content_type
+          encoding = content_type[/charset=["']?(.*?)($|["';\s])/i, 1]
+          return encoding if encoding
+        end
+
+        # look for a charset in a meta tag in the first 1024 bytes
+        data = head.gsub(/<!--.*?(-->|\Z)/m, "")
+        data.scan(/<meta.*?>/im).each do |meta|
+          encoding = meta[/charset=["']?([^>]*?)($|["'\s>])/im, 1]
+          return encoding if encoding
+        end
+
+        nil
       end
     end
   end
--- a/test/html5/test_encoding.rb
+++ b/test/html5/test_encoding.rb
@@ -94,14 +94,6 @@ class TestHtml5Encoding < Nokogiri::TestCase
     assert_equal("こんにちは！", doc.at_css("title").content)
   end
 
-  def test_io_honors_external_encoding_chosen_by_caller
-    doc = File.open(SHIFT_JIS_NO_CHARSET, encoding: "Shift_JIS") do |io|
-      Nokogiri::HTML5::Document.parse(io)
-    end
-
-    assert_equal("こんにちは！", doc.at_css("title").content)
-  end
-
   def test_io_honors_encoding_argument
     doc = File.open(SHIFT_JIS_NO_CHARSET) do |io|
       Nokogiri::HTML5::Document.parse(io, encoding: "Shift_JIS")
@@ -121,16 +113,18 @@ class TestHtml5Encoding < Nokogiri::TestCase
       file.write("<title>Señor</title>")
       file.close
 
-      doc = File.open(file.path) { |io| Nokogiri::HTML5::Document.parse(io) }
+      doc = File.open(file.path, encoding: "UTF-8") { |io| Nokogiri::HTML5::Document.parse(io) }
 
       assert_equal("Señor", doc.at_css("title").content)
     end
   end
 
-  def test_io_sniffs_utf8_bom
-    Tempfile.create(["utf8-bom", ".html"]) do |file|
+  # A UTF-8 BOM would decode correctly under a UTF-8 locale with or without the prescan, so
+  # use a UTF-16 BOM: those bytes are not readable as anything else.
+  def test_io_sniffs_utf16_bom
+    Tempfile.create(["utf16-bom", ".html"]) do |file|
       file.binmode
-      file.write("\xEF\xBB\xBF<title>Se\xC3\xB1or</title>".b)
+      file.write("\xFF\xFE".b + "<title>Señor</title>".encode("UTF-16LE").b)
       file.close
 
       doc = File.open(file.path) { |io| Nokogiri::HTML5::Document.parse(io) }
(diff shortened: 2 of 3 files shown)


lib/nokogiri/html5.rb
```diff
@@ -285,10 +285,26 @@ def fragment(...)
...
-          else
-            string.read(encoding: encoding)
+          external = string.external_encoding if string.respond_to?(:external_encoding)
+          internal = string.internal_encoding if string.respond_to?(:internal_encoding)
+          string = string.read
+
+          if encoding
+            string = string.dup
+            string.force_encoding(encoding)
+          elsif internal.nil? && external == Encoding.default_external
```
flavorjones: This is the part I'd most like to talk through, because I don't think it can be fixed in place.

The PR says a caller-selected external encoding still takes precedence. It only does when that encoding *differs* from `Encoding.default_external`. `File.open(path, encoding: "UTF-8")` and `File.open(path)` produce IOs that are indistinguishable on a host whose default is UTF-8, which is nearly all of them — `IO#external_encoding` returns the effective encoding in both cases and Ruby records nothing about how it was chosen. So an explicit caller choice is silently discarded whenever it happens to match the locale:

```ruby
# file contains UTF-8 bytes with a stale declaration:
#   <meta charset="iso-8859-1"><title>Señor</title>

File.open(path)                    { |io| Nokogiri::HTML5(io) }  # => "SeÃ±or"
File.open(path, encoding: "UTF-8") { |io| Nokogiri::HTML5(io) }  # => "SeÃ±or"  <- caller ignored
```

`test_io_honors_external_encoding_chosen_by_caller` uses `Shift_JIS`, which differs from `default_external`, so it passes without covering this.

I looked for a way to ask Ruby whether an IO's external encoding was chosen or inherited and couldn't find one. If it exists I'd be glad to be corrected — but if it doesn't, this needs a different framing rather than a tighter condition. Two that seem workable:

1. Always prefer the document's declaration for IO input, and document that an IO's external encoding is not authoritative. Callers who want to override say so with the `encoding:` argument, which this same commit makes work for the first time.
2. Treat only an explicit opt-out as authoritative — `binmode`, a non-nil `internal_encoding`, or the `encoding:` argument — and sniff otherwise.

Both are defensible and both are honest about what they do. The current condition promises something it can't deliver.

lib/nokogiri/html5.rb
```diff
@@ -285,10 +285,26 @@ def fragment(...)
...
+            # The IO was opened without an explicit encoding, so its external encoding is the
+            # process default and says nothing about the document. Prefer an encoding the document
+            # declares itself, and keep the IO's encoding when the document declares none.
```
flavorjones: This is the other half of the gate, and I think it undercuts the PR's own argument.

The premise is that `Encoding.default_external` "reflects the locale rather than the document" — correct. But when the document declares nothing, this branch falls back to exactly that locale encoding, where it is even less informative than in the case you're fixing.

The standard's answer for "no declaration" is windows-1252, and `reencode` already defaults to `ISO_8859_1` for the binary path. So after this change `File.open(path)` and `File.binread(path)` of the same declaration-less non-ASCII document still disagree, and under a non-UTF-8 locale the `File.open` form raises instead of parsing:

```
$ LC_ALL=C ruby -Ilib -Itest test/html5/test_encoding.rb
Encoding::InvalidByteSequenceError: "\xC3" on US-ASCII
    lib/nokogiri/html5.rb:353:in 'String#encode'
```

There's a real compatibility argument for keeping the IO's encoding here — it's the conservative choice and it avoids changing behavior for documents that currently work. I'm not saying it's wrong. But right now it reads as a consequence of the condition's shape rather than a decision, and it deserves a comment saying which it is.

lib/nokogiri/html5.rb
```diff
@@ -285,10 +285,26 @@ def fragment(...)
...
+            declared = detect_encoding(string)
+            if declared
+              string = string.dup
+              begin
+                string.force_encoding(declared)
+              rescue ArgumentError
```
flavorjones, line 310: This `rescue` is load-bearing more often than it looks, because the standard's charset labels are not Ruby encoding names.

"Get an encoding" maps roughly 220 labels onto encodings. `String#force_encoding` accepts only Ruby's own names and aliases. I confirmed all of these entirely valid HTML5 labels raise `ArgumentError`:

```
utf8   latin1   UTF8   shift-jis   ms932   csshiftjis   x-sjis   unicode-1-1-utf-8   x-user-defined
```

So for a document declaring any of them, this branch swallows the label and keeps the IO's encoding, while `reencode` maps the same label to `ISO_8859_1`. The two paths disagree, and neither matches the standard:

```
<meta charset="utf8">    File.open => "Señor"   File.binread => "SeÃ±or"
<meta charset="latin1">  File.open => "Señor"   File.binread => "SeÃ±or"
```

`<meta charset="utf8">` is common in the wild, so this isn't hypothetical.

A full label table is probably more than this PR should take on, and I'd understand deferring it. But it's worth knowing that it's the bulk of what the prescan actually is, and at minimum the two paths shouldn't disagree about the same unmappable label. A small normalization step in `detect_encoding` — downcase, strip, and map the handful of common aliases Ruby rejects — would cover most real documents without importing the whole table.

lib/nokogiri/html5.rb
```diff
@@ -361,6 +352,37 @@ def reencode(body, content_type = nil)
...
+        # look for a charset in a content-encoding header
+        if content_type
+          encoding = content_type[/charset=["']?(.*?)($|["';\s])/i, 1]
+          return encoding if encoding
+        end
+
+        # look for a charset in a meta tag in the first 1024 bytes
+        data = head.gsub(/<!--.*?(-->|\Z)/m, "")
+        data.scan(/<meta.*?>/im).each do |meta|
+          encoding = meta[/charset=["']?([^>]*?)($|["'\s>])/im, 1]
```
flavorjones, line 425: The label returned here goes straight to `force_encoding` without the overrides the standard requires after "get an encoding":

- a `meta` declaring `UTF-16BE` or `UTF-16LE` must be treated as UTF-8 (a document that really were UTF-16 could not have contained an ASCII-readable `meta` in the first place)
- `x-user-defined` must be treated as windows-1252

Neither is applied, so a UTF-8 document with a UTF-16 declaration now yields nothing parseable:

```ruby
# <meta charset="UTF-16LE"><title>Señor</title>
File.open(path) { |io| Nokogiri::HTML5(io) }.at_css("title")   # => nil
```

This is pre-existing in the `reencode` path, so it isn't something you introduced. But it was previously unreachable from IO input — that document parsed fine before this change — so the PR makes it newly visible to a much more common input shape. Two lines in `detect_encoding` would close it:

```ruby
return Encoding::UTF_8 if /\Autf-16/i.match?(encoding)
```

The BOM branch above is right, by the way, and I checked that a BOM correctly outranks a contradicting `meta`.

test/html5/test_encoding.rb
```diff
@@ -87,6 +87,75 @@ def test_tag_after_utf8_bom
...
+  def test_io_keeps_its_encoding_when_the_document_declares_none
+    Tempfile.create(["utf8-no-charset", ".html"]) do |file|
+      file.write("<title>Señor</title>")
+      file.close
+
+      doc = File.open(file.path) { |io| Nokogiri::HTML5::Document.parse(io) }
+
+      assert_equal("Señor", doc.at_css("title").content)
+    end
+  end
```
flavorjones, line 120: This test only passes under a UTF-8 locale. Under `LC_ALL=C`, where `Encoding.default_external` is `US-ASCII`, it errors:

```
Encoding::InvalidByteSequenceError: "\xC3" on US-ASCII
    lib/nokogiri/html5.rb:353:in 'String#encode'
    lib/nokogiri/html5.rb:320:in 'Nokogiri::HTML5.read_and_encode'
    test/html5/test_encoding.rb:124
```

`main` is green under the same locale, so this is new. CI runs UTF-8, which is why it isn't showing up there.

It's a fair test of the intended behavior — it's the behavior that's locale-dependent, not the test's fault — but as written it makes the suite fail for anyone running under a POSIX locale. Either pin the locale for this case, or let it follow from whatever you decide about the no-declaration fallback.

Two nearby notes on the rest of the new tests, both minor:

`test_io_sniffs_utf8_bom` passes before this change and would keep passing with the feature removed — under a UTF-8 default a UTF-8 BOM document decodes correctly either way, so it doesn't pin the BOM path. A UTF-16 BOM would actually exercise it.

Nothing covers a `StringIO`, an unmappable label, a `meta` past the 1024-byte window, empty or truncated input, or a BOM contradicting a `meta`. I checked those by hand and the interesting ones are in the review body.

flavorjones (commented):
Written with the help of Claude Opus 5.

Thanks for taking this on — #2801 has sat open for a while precisely because encoding detection is unpleasant, and you did the hard part: you found the actual mechanism (`#reencode` only ever sniffs `ASCII-8BIT`, so an IO's external encoding silently outranks the document's own declaration) and you wrote tests that fail before the change and pass after. I reverted `lib/nokogiri/html5.rb` alone and confirmed four of the seven fail without it. Pulling the sniffing out of `reencode` into `detect_encoding` is a good move on its own, and I checked it carefully: `byteslice(0, 1024).b` matches the old `body[0..1023]` for an `ASCII-8BIT` body, the BOM/`content_type`/`meta` precedence is unchanged, and even the quirk where a bare `charset=` yields a truthy `""` and suppresses the `meta` scan survives intact. Full suite green, `rubocop` clean.

My comments are on the specific lines. The rest is here.

## `StringIO` and `String` now decode differently

`StringIO#external_encoding` reports the underlying string's encoding, so a `StringIO` wrapping already-decoded text takes the new sniffing path:

```ruby
s = %(<meta charset="iso-8859-1"><title>Señor</title>)   # a correct UTF-8 String

Nokogiri::HTML5(s)                  # => "Señor"
Nokogiri::HTML5(StringIO.new(s))    # => "SeÃ±or"
```

Wrapping a String in a `StringIO` shouldn't change how it decodes. Whatever you decide about the gate, this pair is worth a test.

## The changelog understates the change

The entry reads as "Shift_JIS files opened with `File.open` now work". What actually changed is that **every** IO-parsed document now obeys its own `meta` declaration in preference to the bytes it was read as. A UTF-8 page carrying a stale `<meta charset="iso-8859-1">` parsed correctly before this change and mojibakes after it.

I don't think that's wrong — it is what the standard says, and it's what `File.binread` of the same document already did. But `Nokogiri::HTML5(File.open(...))` and `Nokogiri::HTML5(URI.open(...))` are common, so it's the part most likely to surprise someone on upgrade, and the changelog should say it outright.

## Exception class in the PR description

The description says a document whose declaration disagrees with its bytes "now raises `Encoding::UndefinedConversionError`". In every case I constructed it raises `Encoding::InvalidByteSequenceError`:

```
<meta charset="Shift_JIS"> + invalid Shift_JIS bytes  =>  Encoding::InvalidByteSequenceError
<meta charset="EUC-JP">    + invalid EUC-JP bytes     =>  Encoding::InvalidByteSequenceError
```

I couldn't construct one that raises `UndefinedConversionError`, and I don't believe one exists on this path: the destination is always UTF-8, which can represent every codepoint, so `String#encode` here can only fail on invalid source bytes. This is description-only — nothing in the code or changelog depends on it.

Relatedly, `test_io_prefers_the_declaration_over_the_bytes` asserts equality against the binary-String path rather than a literal, so it pins consistency between the two paths rather than correctness, and would keep passing if both broke together. The bytes it uses happen to be valid half-width katakana in Shift_JIS, so it never reaches the raise the description advertises. An explicit `assert_raises` alongside it would cover the documented behavior.

## Two smaller things

`string = string.dup` after `IO#read` copies the whole document for nothing — `IO#read` already returns a fresh, unshared String. Three occurrences on the IO path.

The comment "an unknown encoding name is not a reason to discard what the IO gave us" states a policy that `reencode` contradicts ten lines further down by falling back to `ISO_8859_1`. Both are defensible; having both in the same file isn't.

## Where this leaves #2801

Item 2 is genuinely fixed, and the diagnosis in your commit message is exactly right: `IO#read` takes a length, not an `encoding:` keyword. `force_encoding` after the read is the right shape and matches what the String branch already does.

Item 3 is partway there — the prescan now reaches IO input, which it didn't before. The line comments below are the remaining distance.

Worth flagging for expectations: the issue's own reproduction still fails on every HTML5 assertion, because `doc.encoding` is untouched and still reports `"UTF-8"`. Leaving that alone is the right call given @stevecheckoway's open question about how it should interact with serialization, but it does mean this PR can't close #2801.

One more thing to decide rather than to fix: `Nokogiri::HTML4::EncodingReader.detect_encoding` at `lib/nokogiri/html4/encoding_reader.rb:59` already exists, and item 3 of the issue floated reusing it. This PR adds a third sniffer to the codebase instead. Sharing one may or may not be worth it, but it's a maintainer call worth making explicitly.

## What I'm not sure about

- Whether any real-world caller depends on the old "IO encoding outranks the declaration" behavior. I only reasoned about it; the downstream CI jobs were still running when I looked.
- Whether the maintainers want the standard's full label table in Nokogiri at all, or consider Ruby's encoding names close enough. That's scope, not a defect.
- **Windows.** I tested only on Linux, where `Encoding.default_external` is UTF-8. On Windows it's typically `Windows-31J` or `IBM437`. Both the gate and the no-declaration fallback are keyed on `default_external`, so the behavior there is likely to differ from what I measured, and I have no data on it. Given CI covers Windows, that's worth checking before this lands.

lib/nokogiri/html5.rb
kataokatsuki: You're right that this can't be fixed in place, and I couldn't find a way to ask Ruby
either. Every IO attribute I could inspect is identical for the two forms —
external_encoding, internal_encoding, binmode?, autoclose?, sync, lineno, pos, path.

I took the first framing. The condition is now `internal.nil? && !wraps_a_string`, so an
IO's external encoding is never treated as authoritative. That also means
`File.open(path, encoding: "Shift_JIS")` no longer wins over a declaration; the only
override left is the `encoding:` argument to `parse`, which this commit makes work. With
`<meta charset="iso-8859-1">` over UTF-8 bytes:

    File.open(path)                                   => "SeÃ±or"
    File.open(path, encoding: "UTF-8")                => "SeÃ±or"
    File.open(path, encoding: "Shift_JIS")            => "SeÃ±or"
    HTML5::Document.parse(io, encoding: "Shift_JIS")  => "Seﾃｱor"
    HTML5::Document.parse(io, encoding: "UTF-8")      => "Señor"

`test_io_honors_external_encoding_chosen_by_caller` asserted the behavior that no longer
holds, so I removed it. `test_io_honors_encoding_argument` already covers the override.

The `!wraps_a_string` term is a separate fix — see the body comment about StringIO.

lib/nokogiri/html5.rb
kataokatsuki: Kept the IO's encoding and rewrote the comment so it reads as the decision it is:

    # When the document declares nothing, keep the IO's encoding. The standard's answer
    # there is windows-1252 and #reencode falls back to ISO_8859_1, so this is a third
    # answer and a deliberate one: it is the conservative choice and it leaves documents
    # that parse correctly today parsing the same way.

The `LC_ALL=C` failure you saw was the test rather than this branch; fixed there.

lib/nokogiri/html5.rb
kataokatsuki: Added the normalization step you suggested: strip, downcase, and a table of the labels
Ruby rejects — utf8, unicode-1-1-utf-8, latin1, shift-jis, x-sjis, ms932, csshiftjis,
x-user-defined. `latin1` follows `iso-8859-1`, which Ruby already maps to ISO-8859-1,
rather than the standard's windows-1252 fold — mapping only one of the two synonyms would
make them decode 0x80-0x9F differently. The two paths now agree:

    <meta charset="utf8">        File.open => "Señor"   File.binread => "Señor"
    <meta charset="latin1">      File.open => "SeÃ±or"  File.binread => "SeÃ±or"
    <meta charset="UTF-8">       File.open => "Señor"   File.binread => "Señor"
    <meta charset="iso-8859-1">  File.open => "SeÃ±or"  File.binread => "SeÃ±or"

I also made this rescue match `#reencode` — see the body comment.

lib/nokogiri/html5.rb
kataokatsuki: Both added. `x-user-defined` is in the alias table; UTF-16 is the line you suggested, in
`normalize_encoding_label`:

    <meta charset="UTF-16LE">        => "Señor"   (was nil)
    <meta charset="UTF-16BE">        => "Señor"   (was nil)
    <meta charset="UTF-16">          => "Señor"   (was Encoding::InvalidByteSequenceError)
    <meta charset="x-user-defined">  => "SeÃ±or"

The BOM branch is untouched, and I confirmed a BOM still outranks a contradicting `meta`.

test/html5/test_encoding.rb
kataokatsuki: Pinned the locale by opening with `encoding: "UTF-8"`, which is also what the test is
about — the IO's encoding being kept when the document declares nothing:

    LC_ALL=en_US.UTF-8    225 runs, 746 assertions, 0 failures, 0 errors
    LC_ALL=C              225 runs, 746 assertions, 0 failures, 0 errors
    LC_ALL=ja_JP.eucJP    225 runs, 746 assertions, 0 failures, 0 errors

Replaced `test_io_sniffs_utf8_bom` with a UTF-16 BOM, which errors on main and passes
here. Added tests for StringIO and an `assert_raises` pair; details in the body comment.

kataokatsuki:
Thanks for going through this so carefully — most of what follows is a direct result of it.
Each heading below is one of yours.

## StringIO and String now decode differently

You're right, and it's a regression this PR introduced: on main `Nokogiri::HTML5(s)` and
`Nokogiri::HTML5(StringIO.new(s))` agree, and with the sniffing added they don't.
`Document.parse` promotes a String's own encoding into the `encoding:` argument at
html5/document.rb:112, but a StringIO does not respond to `encoding`, so it fell through to
the IO path. `DocumentFragment.parse` never reaches that line, so I fixed it in
`read_and_encode` instead, by remembering whether the input is a StringIO. Both now match
the String they hold, for a UTF-8 String and for a binary one:

    Document  String="Señor"  StringIO="Señor"
    Fragment  String="Señor"  StringIO="Señor"

## The changelog understates the change

Split in two, with the behavior change under Changed and the stale-declaration case spelled
out.

## Exception class in the PR description

The description naming only one was wrong either way; fixed. I did turn up a case that
raises `UndefinedConversionError`, though, so I may be reading your point differently than
you meant it — it comes from the source side rather than the destination. Ruby raises it
when the source byte sequence is valid in the declared encoding and that position has no
character assigned to it, so there is nothing to hand to UTF-8 in the first place. An EUC-JP
page declaring Shift_JIS reaches it.

    body encoded EUC-JP, <meta charset="Shift_JIS">  => UndefinedConversionError ("\xFC\xCB")
    body encoded UTF-8,  <meta charset="Shift_JIS">  => InvalidByteSequenceError (incomplete "\x9E")

`test_io_raises_when_the_declaration_cannot_decode_the_bytes` covers both, asserting the
error class rather than agreement between the two paths, and asserts the same for the
binary String.

## Two smaller things

`string.dup`: I removed the two in `read_and_encode`'s IO branch. The third one you counted
is `body.dup` in `#reencode`, which an IO does reach when it arrives in binmode — I left that
one because `#reencode` is shared with the String path, where dropping it would change the
caller's own String as a side effect of parsing. My earlier reply pointed at the `dup` in the
String branch instead; that was a miscount on my part.

The rescue comment: made the IO path fall back to `ISO_8859_1` like `#reencode`, rather than
keeping an encoding this branch has already decided says nothing about the document. For
labels Ruby rejects that aren't in the alias table, the two paths now agree. Doing that alone
would have dragged the truthy-`""` quirk onto the IO path — a bare `charset=` would have
become ISO-8859-1 there instead of the IO's encoding — so `normalize_encoding_label` returns
nil for an empty label and the prescan moves on to the next `meta`, which is what "get an
encoding" failing means. `<meta charset="">` alone counts as no declaration on both paths,
and `<meta charset=""><meta charset="utf-8">` now finds the utf-8; on main the binary path
stopped at the empty one.

## Where this leaves #2801

Agreed, including that this can't close it. `doc.encoding` is untouched.

`HTML4::EncodingReader.detect_encoding`: left alone. It reads a chunk stream for a different
parser and its result goes to libxml2 rather than to a Ruby String's encoding, so sharing
looked like more than this PR should carry. #2513 is also still open on whether that class
should exist at all, and I would rather not add a dependency on something you are
considering removing. Happy to do it if you'd rather.

## What I'm not sure about

Real-world callers relying on the IO's encoding outranking the declaration: I don't have a
good answer here — I don't know of any, and I'm not in a position to rule it out. The
Changed entry in the changelog names that case so it is at least visible on upgrade, and
`encoding:` is the way back.

The label table: `latin1` maps to ISO-8859-1 here, not to the standard's windows-1252,
because Ruby already maps `iso-8859-1` to ISO-8859-1 and mapping only one of the two
synonyms would make them decode 0x80-0x9F differently inside Nokogiri. Folding both into
windows-1252 is the standard-conformant answer, but it changes what a binary String with
`<meta charset="iso-8859-1">` has decoded as for years. Happy to go that way if you'd rather
take the whole label table. One related gap I did not close, pre-existing on main: an
unmappable first `meta` is still taken as the answer instead of letting the prescan keep
looking, so `<meta charset="bogus"><meta charset="utf-8">` decodes as ISO-8859-1 on this
branch and on main alike. Closing it means `detect_encoding` returning a usable encoding or
failure rather than the raw label, so I left it; say the word if you want it in this PR.
(The regex still captures an empty string for `<meta charset=" UTF-8 ">`, so that document
is treated as declaring nothing — the same as on main.)

Windows: still unmeasured; I only have macOS here. `Encoding.default_external` no longer
appears in the condition, and `test/html5/test_encoding.rb` passes identically under
en_US.UTF-8, C and ja_JP.eucJP, so the sniffing decision is locale-independent. The
no-declaration fallback still keeps the IO's encoding, which is where Windows will differ.

    bundle exec rake test    6704 runs, 113853 assertions, 0 failures, 0 errors, 24 skips
    bundle exec rubocop      no offenses

kataokatsuki:
Pushed a follow-up after a second pass over the label handling; it changes nothing you
asked for, only what that pass turned up:

- A label Ruby cannot resolve is now skipped and the prescan continues, instead of being
  taken as the declaration. `<meta charset="bogus"><meta charset="utf-8">` finds the utf-8
  on both paths; on this branch the IO path had been decoding it as ISO-8859-1.
- The charset value is read whole before trimming, so `charset=" UTF-8 "` is UTF-8 rather
  than an empty label.
- The UTF-16 -> UTF-8 replacement is limited to the three UTF-16 labels and to a `meta`;
  `/\Autf-16/` had also caught `utf-16bogus`.
- `ms932` maps to Windows-31J. Ruby's Shift_JIS lacks its NEC and IBM extensions, so
  `①` under `ms932` raised `UndefinedConversionError`.

Each has a test that fails on 354eafed. The CHANGELOG entry now names the exceptions
(StringIO, a transcoding IO, `encoding:`) and the label changes, and I edited the body
comment above where it said Windows was unmeasured and where it described the gap.

PR #3686, opened 2026-08-31
Close files opened for Pathname document parsing

### LINKED ISSUE #3682: XML and HTML4 Pathname parsing leaks opened file descriptors
## Summary

`XML::Document.parse` and `HTML4::Document.parse` open `Pathname` inputs but do not close the resulting IO. Repeated parsing retains descriptors until garbage collection.

## Reproduction

```ruby
path = Pathname('/tmp/input.xml')
path.write('<root/>')
32.times { Nokogiri::XML::Document.parse(path) }
```

On a system with `/dev/fd`, the open descriptor count grows by 32 before GC. HTML4 has the same behavior.

## Candidate and verification

Use block-form `Pathname#open` and parse inside the block. Source-only candidate `277beb48391b57e3faab2a3144fc2786e8985854` is based on current main. Current passes 2,841 runs /10,329 assertions with 24 skips. Focused XML/HTML4 models show zero descriptor growth and preserve URL/encoding/options behavior; release leaks one descriptor per parse.

This is issue-first because Nokogiri requires behavioral tests/changelog changes and the consumer audit does not modify upstream tests.

OskarEichler:
## Summary

- parse Pathname inputs through block-form Pathname#open
- preserve XML and HTML4 URL, encoding, parse options, and XML XInclude behavior
- add cross-platform regression coverage asserting the internally opened stream is closed
- document the fix in the unreleased changelog

Fixes #3682.

## Problem

XML::Document.parse and HTML4::Document.parse opened Pathname inputs without closing the internally owned streams. Repeated parsing could therefore retain file descriptors until garbage collection.

## Verification

- XML Pathname regression test: 1 run, 7 assertions, 0 failures
- HTML4 Pathname regression test: 1 run, 7 assertions, 0 failures
- Ruby syntax checks for both implementation and test files
- git diff --check

The focused tests loaded the patched current-main Ruby sources against the official Nokogiri 1.19.4 arm64-darwin native extension. The full native and cross-platform matrix is left to CI.

### DIFF
--- a/test/xml/test_document.rb
+++ b/test/xml/test_document.rb
@@ -8,6 +8,25 @@ module Nokogiri
   module XML
     class TestDocument < Nokogiri::TestCase
       describe Nokogiri::XML::Document do
+        class TrackingPathname < Pathname
+          attr_reader :opened_io
+
+          def expand_path
+            self
+          end
+
+          def open(*args, &block)
+            if block
+              super(*args) do |io|
+                @opened_io = io
+                block.call(io)
+              end
+            else
+              @opened_io = super(*args)
+            end
+          end
+        end
+
         URI = if URI.const_defined?(:DEFAULT_PARSER)
           ::URI::DEFAULT_PARSER
         else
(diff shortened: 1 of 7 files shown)


test/xml/test_document.rb
```diff
@@ -8,6 +8,25 @@ module Nokogiri
...
+              super(*args) do |io|
+                @opened_io = io
+                block.call(io)
+              end
+            else
+              @opened_io = super(*args)
+            end
+          end
+        end
+
```
stevecheckoway: Rather than duplicating this class, what about putting it in `test/helpers`?

OskarEichler: Done

flavorjones:
Thanks for the PR. I'll review when I have time.

